PR #13, opened 2021-03-18
Add namespace status on list and details view, hide system namespaces…

akucharska:
… when neccessary

**Description**

- Kyma Namespaces are shown even if the setting is disabled in the prefferences
- Terminating state of a namespace is missing in Namespace List
- Remove the Upload label from New Workload Popup
**Related issue(s)**

### DIFF
--- a/core-ui/src/components/Predefined/List/Namespaces.list.js
+++ b/core-ui/src/components/Predefined/List/Namespaces.list.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import { NamespaceStatus } from '../Details/Namespace/NamespaceStatus';
+import LuigiClient from '@luigi-project/client';
+import { useShowSystemNamespaces } from 'react-shared';
+
+const FilterNamespaces = namespace => {
+  const showSystemNamespaces = useShowSystemNamespaces();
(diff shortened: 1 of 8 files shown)


core-ui/src/components/Predefined/List/Namespaces.list.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { NamespaceStatus } from '../Details/Namespace/NamespaceStatus';
+import LuigiClient from '@luigi-project/client';
+import { useShowSystemNamespaces } from 'react-shared';
+
+const FilterNamespaces = namespace => {
+  const showSystemNamespaces = useShowSystemNamespaces();
+  const systemNamespaces = LuigiClient.getContext().systemNamespaces; //.includes('showSystemNamespaces');
```
dariadomagala, line 8: please remove the comment

dariadomagala (commented):
It's probably not introduced by your PR, but the 'show system namespaces' toggle does not work as intended. If I go to the 'preferences' view, the toggle is always disabled, even tho I previously enabled it. The state of a toggle doesn't always show what's really set on the cluster.

dariadomagala (approved):

PR #42, opened 2021-03-24
Add EventSubscriptions to service details

parostatkiem:
**Description**

- Add _Cluster IP_ and _Ports_ information to the header
- Move some of the `EventSubscriptions` code to the shared folder
- Use `EventSubscriptions` on service details view
- Improve styling of `InputWithPrefix` (decrease the gap between prefix and input)

**Related issue(s)**


Wawrzyn321:
Would you kindly add a tiny margin over here?
image

Wawrzyn321 (changes requested):
Aaaand I can see `Encountered two children with the same key, '53'.` I think we could use `port:targetPort` instead of just `port`, both as key and on the view.

Wawrzyn321 (approved):

PR #43, opened 2021-03-24
Various fixes & improvements

Wawrzyn321:
**Description**

- Add fallback value for deployment replicas so it doesn't show e.g. `UNDEFINED / 1`
- Add 'restarts' field in pod list.
- Add status for auth client details

**Related issue(s)**


core-ui/src/components/Predefined/List/Pod/PodRestarts.js
```diff
@@ -0,0 +1,24 @@
...
+  const tooltipContent = (() => {
+    if (!restartCount) return null;
+    return (
+      <ul style={{ textAlign: 'left' }}>
+        {statuses.map(s => (
+          <li key={s.name}>{`${s.name}: ${s.restartCount}`}</li>
+        ))}
+      </ul>
+    );
+  })();
```
parostatkiem, line 17: Why did you use a self executing function instead of a normal _Component_?

Wawrzyn321: I didn't want to write
```
let tooltipContent = null;
if (restartCount) {
  tooltipContent = 'yadadada'
}
```
nor
```
const getTooltipContent = () => {
	if (!restartCount) return null;
	tooltipContent = ...
}
...tooltipContent={getTooltipContent()}>
```
so I wanted something else.

Yeah, tiny component will be also okay.

Wawrzyn321: Or nope, it won't be okay. Let's go with function.

parostatkiem (approved):

PR #45, opened 2021-03-24
Make Brokers and ClusterBrokers readonly

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


Wawrzyn321 (changes requested):
Should we also block creation of resource if it's readonly?

valentinvieriu (approved):

PR #74, opened 2021-04-06
Handle token expire

Wawrzyn321:
**Description**

- Show Luigi error prompt when initial requests failed - before they were either shown in `alert` with no concrete message, or just pushed to dev console.
- About the token expiration:
	- This is kind of funny, but normally you should use `access_token` for auth, and `id_token` for displaying user data, but for authenticating our requests, we do send `id_token`. [ID Tokens should never be used to obtain direct access to APIs or to make authorization decisions.](https://auth0.com/docs/tokens/id-tokens) More [here](https://github.com/kubernetes/kubernetes/issues/67018).
	- I wanted Luigi to handle expiration stuff, but despite `automaticSilentRenew` set to _true_, it doesn't fire at all.
	- I thought about using refresh token to get new pair of access and id token, but it looks like we'd [need to store and send `clientSecret`](https://auth0.com/docs/tokens/refresh-tokens/get-refresh-tokens).
```
When a new access token is needed, the application can make a POST request back to the token endpoint using a grant type of refresh_token (web applications need to include a client secret).
```
- Faced with this dire situation, I decided to just reload the page when we get 401 AND our `id_token` is expired.


shared/hooks/BackendAPI/checkForTokenExpiration.js
```diff
@@ -0,0 +1,12 @@
...
+import LuigiClient from '@luigi-project/client';
+
+export function checkForTokenExpiration(idToken) {
+  try {
+    const expirationTimestamp = parseJWT(idToken).exp * 1000;
+    if (new Date(expirationTimestamp) < Date.now()) {
+      // we cannot run parent.location.reload() here
+      LuigiClient.sendCustomMessage({ id: 'busola.refreshMainFrame' });
+    }
+  } catch (_) {} // ignore errors from non-JWT tokens
```
Sawthis, line 11: Do we want swallow all exception here?

Wawrzyn321: Yeah, because idToken may be a non-jwt token, for example in case kubeconfig uploaded from file.

Sawthis (dismissed):

core/src/luigi-config/navigation/navigation-data-init.js
```diff
@@ -77,7 +77,10 @@ export function getNavigationData(authData) {
             clearAuthData();
             window.location.pathname = '/nopermissions.html';
           } else {
-            alert('Config init error, see console for more details');
+            Luigi.ux().showAlert({
+              text: `Could not load initial configuration: ${err.message}`,
+              type: 'error',
+            });
             console.warn(err);
```
Sawthis, line 84:
```suggestion
            console.warn(err.message);
```

Wawrzyn321: I don't think we need this in `console` calls:
image
When we show only message, we don't get the call stack.

Sawthis (approved):

dariadomagala (approved):

PR #82, opened 2021-04-08
Fix Function's Subscription creation

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core-ui/src/shared/components/EventSubscriptions/EventSubscriptions.js
```diff
@@ -32,7 +33,7 @@ export default function EventSubscriptions({
       namespace,
       ownerRef,
       sink,
-      eventType,
+      EVENT_TYPE_PREFIX + eventType,
```
dariadomagala, line 36: why won't we just add it in the `CreateEventSubscriptionForm` where we have the `EVENT_TYPE_PREFIX` already defined?

parostatkiem: Because there we use the pure HTML form's field value in the `onSubmit` callback. It is possible but it would complicate things. That's why I decided to put it here.

dariadomagala (approved):

PR #86, opened 2021-04-12
Make login prettier

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core/src/assets/login.js
+++ b/core/src/assets/login.js
@@ -36,11 +36,12 @@ function toggleError(isVisible) {
   document.querySelector('#error').style.display = isVisible ? 'block' : 'none';
 }
 
-async function onKubeconfigUploaded(e) {
+async function onKubeconfigUploaded(file) {
+  document.querySelector('#file-name').textContent = file.name;
   toggleError(false);
   toggleOIDCForm(false);
   try {
-    const kk = jsyaml.load(await readFile(e.target.files[0]));
+    const kk = jsyaml.load(await readFile(file));
     cluster = {
       server: kk.clusters[0].cluster.server,
       'certificate-authority-data':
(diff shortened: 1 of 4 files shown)


core/src/assets/login.js
```diff
@@ -61,13 +79,38 @@ async function onKubeconfigUploaded(e) {
...
   } catch (e) {
     toggleError(true);
     console.warn(e);
   }
 }
 
+async function onKubeconfigPasted(kubeconfig, type) {
+  displayInitialView();
+  try {
+    const kk = jsyaml.load(kubeconfig);
```
parostatkiem, line 94: Even in this context, I think this name is too mysterious. What about `kubeconfigParsed`?

core/src/assets/login.js
```diff
@@ -61,13 +79,38 @@ async function onKubeconfigUploaded(e) {
...
+    toggleError(true);
+    console.warn(e);
+  }
+}
+
+async function onKubeconfigUploaded(file) {
+  document.querySelector('#file-name').textContent = file.name;
+  displayInitialView();
+  try {
+    const kk = jsyaml.load(await readFile(file));
```
parostatkiem, line 106: The same as above 🆙

core/src/assets/login.js
```diff
@@ -32,21 +32,39 @@ function toggleOIDCForm(isVisible) {
...
 }
 
-async function onKubeconfigUploaded(e) {
+function displayInitialView() {
   toggleError(false);
   toggleOIDCForm(false);
+  toggleConfigForm(true);
+  toggleBackButton(false);
+}
+function handleKubeconfig(kubeconfig, type) {
```
parostatkiem, line 60: You cannot handle a file 😄  
What about `handleKubeconfigAdded` or `applyKubeconfig`?

parostatkiem (changes requested):
Looks good, I just noticed some mysterious names 👌

parostatkiem (approved):

PR #87, opened 2021-04-12
Avoid too many renders and requests sent on route change

parostatkiem:
**Description**

- Update Luigi to the latest version
- improve `useGet` hook to not overreact to `MicrofrontendContext` updates
- improve `getComponentFor` to be a normal React component
- remove the `preloadingStrategy` wrapper over Core-UI
- improve `MicrofrontendContext` to ignore duplicated context updates from Luigi
- wrap getting predefined renderer in `useMemo`

**Related issue(s)**


shared/components/PageHeader/PageHeader.js
```diff
@@ -46,7 +46,7 @@ export const PageHeader = ({
     <Panel.Header>
       <section className="header-wrapper">
         {breadcrumbItems.length ? (
-          <section className="fd-has-margin-bottom-s">
```
dariadomagala, line 49: why did we remove this class?

dariadomagala (approved):

shared/components/PageHeader/PageHeader.js
parostatkiem: I decided it looks better by making the high header a bit smaller. Also, it suggests that (which is true) the "path" to the resource is
```
<resource_type> /
<resource_name>
``` 
Previously it was more like
```
<resource_type> /


<resource_name>
```



I can change that back if you want 😃

PR #89, opened 2021-04-13
403  error message propagation

kwiatekus:
**Description**

- propagate full response object so that error message could be propagated to `nopermissions` page


Wawrzyn321 (changes requested):
If the request fails for some other reason (e.x. X-Api-Url is unreachable), the error displayed is 
image

Wawrzyn321 (approved):

PR #91, opened 2021-04-13
Fix FF bugs

Wawrzyn321:
**Description**

- Make ModalWithForm submit event cancellable.
- Add type="button" on buttons in AddonsConfiguration form, so that these buttons don't act as "submit".
- Prevent defaulted form submit in some ModalWithForm forms.

**Related issue(s)**


dariadomagala (changes requested):
This error still occurs when adding
- Service instance
- Git repository
- Event subscription on function details

dariadomagala (approved):

PR #96, opened 2021-04-14
Token refresh

Wawrzyn321:
**Description**

- ~Add navigation out guard for function code~
	- ~Unfortunately, Luigi's `dirtyState` only works for navigation inside MF app...~
	- ~... so in case of browser navigation, we need to fall back to built-in popup.~
	- I decided not to add navigation guards, as token refetch won't interfere with current application state.
- Refetch token when the time comes

We require _yet another_ additional entry in OIDC config - `domain + /silentCallback.html`.

**Related issue(s)**


dariadomagala (changes requested):
I checked it locally and it doesn't work. 
The `refreshAuth` function is called on each navigation change (token lives for 40s in my configuration so it's always triggered) but it doesn't change the token at all. Even after the token is expired, it still uses it and gets 401 as a result.

PR #99, opened 2021-04-20
Smoke tests written in Cypress

Sawthis:
**Description**

- Add smoke tests for Busola written in Cypress

**Related issue(s)**


tests/support/iframe-commands.js
```diff
@@ -1,83 +1,16 @@
...
-      () => {
-        $iframe.on('load', () => {
-          resolve($iframe.contents().find('body'));
-        });
-      },
-    );
-  });
+  return (
+    cy
+      .get('.iframeContainer iframe', { log: false })
```
parostatkiem, line 8: I think we should get the iframe which is currently visible. Luigi can store many iframes inside this container.
image

tests/tests/busola-login-failed.spec.js
```diff
@@ -0,0 +1,28 @@
...
+});
+
+context('Busola Login Failed', () => {
+  before(() => {
+    cy.clearSessionStorage().clearLocalStorage();
+  });
+
+  it('Use wrong kubeconfig', () => {
+    cy.visit(ADDRESS)
+      .get('#textarea-kubeconfig')
```
parostatkiem, line 21: Please use selectors which determine what the user can see. He/she cannot see the ID of the element :) 
https://docs.cypress.io/guides/references/best-practices#Selecting-Elements


This comment applies to almost evry selector in your code.

tests/tests/busola-smoke-tests.spec.js
```diff
@@ -0,0 +1,110 @@
...
+  ? `http://localhost:4200`
+  : `https://busola.${config.domain}`;
+
+const random = Math.floor(
+  Math.random() * 10 + Math.random() * 100 + Math.random() * 1000,
+);
+const NAMESPACE_NAME = `orders-service-${random}`;
+
+Cypress.on('uncaught:exception', (err, runnable) => {
+  // returning false here prevents Cypress from failing the test
```
parostatkiem, line 15: Why would the exception happen?

tests/tests/busola-smoke-tests.spec.js
```diff
@@ -0,0 +1,110 @@
...
+      .should('be.visible')
+      .getIframeBody()
+      .contains('Healthy Resources')
+      .should('be.visible')
+      .getIframeBody()
+      .contains('Warnings')
+      .should('be.visible')
+      .get('body')
+      .contains('Back to Namespaces')
+      .click();
```
parostatkiem, line 72: Please avoid function chains that refer to more than one action.

parostatkiem (changes requested):

PR #100, opened 2021-04-20
Fix system namespaces filtering

Wawrzyn321:
**Description**

To reproduce: create NS with name 'm'. It won't show up on namespace list.

- In case of link-created init params, we split the system namespaces in initParams.js, but in case of kubeconfig drag-n-drop created init params, they weren't split - and 'm' indeed belongs in long string of 'kyma-syste**m** kube-syste**m**...', causing the bug.

**Related issue(s)**


akucharska (commented):
Can you add an image bump?

akucharska (approved):

PR #101, opened 2021-04-20
Remove any external linked javascript dependency

Wawrzyn321:
**Description**

- Reorganize statically served external libraries into `libs/` folder
- Serve `js-yaml` from core
- About the `monaco-editor`... we can't link it during the build time (more [here](https://github.com/suren-atoyan/monaco-react/issues/12)), and it's located in `shared/`. Seeing that I went for symbolic link to the `vs/` folder (monaco core installed in `shared`) that would be resolved during runtime. I had to add a SHA for inline script added by Monaco's `loader`.

<details>
<summary></summary>
BTW I do know what "internalize" means

</details>

**Related issue(s)**


shared/package.json
```diff
@@ -37,6 +37,7 @@
     "caniuse-lite": "^1.0.30001081",
     "jsoneditor": "^8.6.1",
     "jwt-decode": "^3.1.2",
+    "monaco-editor": "^0.23.0",
```
dariadomagala, line 40: why do we use `monaco-editor` here and `monaco-editor/react` in main package.json? Those are 2 different libraries and all our dependencies in the code use `monaco-editor/react`.

dariadomagala: Additionally, we use `monaco-editor/react` not only in shared, but also in core-ui. It would make more sense to put it in the main package.json

shared/package.json
```diff
@@ -7,7 +7,7 @@
     "test:watch": "jest --watch",
     "build": "webpack --env.production",
     "start:kyma": "webpack --env.development --watch",
-    "build:npx": "echo no-op, done during bootstrap"
+    "build:npx": "ln -sf ../shared/node_modules/monaco-editor/min/vs ../backend/vs"
```
dariadomagala, line 10: do we need `../shared` part in `../shared/node_modules/monaco-editor/min/v`?

dariadomagala (changes requested):
running `npm run bootstrap` generated 10+ files that were not included in the gitignore

shared/package.json
Wawrzyn321: Yup, when in my infinite naivety I tried to do `ln -s node_modules/monaco-editor/min/vs ../backend/vs`, it created a symlink in valid target directory, but didn't take the path into account, so it tried to link `busola/backend/node_modules/monaco...`.

dariadomagala (approved):

PR #105, opened 2021-04-21
Add cluster name display

Wawrzyn321:
**Description**

- Unhide top left title
- Add cluster name (instead of Kyma) to title
- Ignore view-specific title for now - I didn't remove any of the `useWindowTitle` or `withTitle` calls, as it would be back after Luigi decouples tab title and top nav title.

**Related issue(s)**


core/src/luigi-config/communication.js
```diff
@@ -39,7 +40,7 @@ export const communication = {
     },
     'busola.setWindowTitle': ({ title }) => {
       const luigiConfig = Luigi.getConfig();
-      luigiConfig.settings.header.title = `Kyma - ${title}`;
+      luigiConfig.settings.header.title = `${config.domain} | ${title}`;
```
dariadomagala, line 43: Maybe it would make more sense to switch places here?
if the domain name would be long, then no one will see the title, and titles are usually short so it shouldn't be a problem the other way around.

valentinvieriu: I've looked around and the patterns are mixed. Let's go for the Suggestion from @dariadomagala ! 

image

image

dariadomagala (approved):

dariadomagala (changes requested):
1. It shows the domain of the busola cluster, not the one we use API from
2. This title should stay as Kyma only
Screen Shot 2021-04-22 at 10 35 13

valentinvieriu (approved):

valentinvieriu:
> 1. It shows the domain of the busola cluster, not the one we use API from
> 2. This title should stay as Kyma only
> 
> Screen Shot 2021-04-22 at 10 35 13

Ups! We need to show the cluster we are pointing to

core/src/luigi-config/communication.js
valentinvieriu: Can we separate the Window tile from the Luigi title? Having the Application name in front there looks strange.

dariadomagala (approved):

PR #110, opened 2021-04-23
(Config Map, Secrets) x (create, update)

Wawrzyn321:
**Description**

- Add `KeyValueForm` component
- Add create & edit for CMs and Secrets

Considerations for Secrets:
* In this iteration, there's no choosing the Secret type. Support for that (and custom validation for specific secret types!) is to be added in future. [More 'bout that](https://kubernetes.io/docs/concepts/configuration/secret/#secret-types).

Considerations for CMs:
* We don't care about [`immutable`](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-immutable)
* For this iteration, no support for `binaryData`

**Related issue(s)**


Wawrzyn321:
bruh

parostatkiem (changes requested):
I think the `value` inputs of the CreateSecretForm should be readonly if the data is encoded. There's no point to allow the user to edit encoded data as he/she can only break it this way.
Also, the "read from file" feature is unclear. I loaded a valid ConfigMap yaml file and got this result:
image
Now I guess I should load a yaml file with `key: value` entries on root level only, should I? Don't make user guess what type file he/she should upload ;)

Wawrzyn321:
`There's no point to allow the user to edit encoded data` - what if user wants to paste base64 value in here? I know they could just use YAML editor, too.
`Read from file` is for reading a file into a single data entry of config map, not for reading entire CM of secret. Should we rename it to `Read value from file`?

parostatkiem:
> `There's no point to allow the user to edit encoded data` - what if user wants to paste base64 value in here? I know they could just use YAML editor, too.
> `Read from file` is for reading a file into a single data entry of config map, not for reading entire CM of secret. Should we rename it to `Read value from file`?

All I care is that the user knows exactly what content of the file we expect. A tooltip with an explanation maybe? 
And about editing the encoded data - you've got a point. However, any other way of editing than pasting is pointless. On the other hand, whatever one pastes, one can also see in the decoded form (because one knows what one pastes). You can leave it as it is, I'm fine with it.

parostatkiem (approved):

PR #117, opened 2021-04-28
Bring back window title

Wawrzyn321:
**Description**

- I decided not to use "Kyma" prefix in title, as we have this Kyma favicon, haven't we?


dariadomagala (approved):
borat

PR #138, opened 2021-05-07
Test deployments

akucharska:
**Description**

- Fix problem with circles being above modal
- Test deployments

**Related issue(s)**


Sawthis (approved):

tests/tests/busola-create-a-deployment.spec.js
```diff
@@ -0,0 +1,162 @@
...
+  });
+
+  it('Check if deployment and service exist', () => {
+    cy.getIframeBody()
+      .contains('a', DEPLOYMENT_NAME)
+      .should('be.visible');
+
+    getLeftNav()
+      .contains('Discovery and Network')
+      .click({ force: true });
```
parostatkiem, line 151: Is there any reason you use `force` here and below?

tests/tests/busola-create-a-deployment.spec.js
```diff
@@ -0,0 +1,162 @@
...
+
+    getLeftNav()
+      .contains('Services')
+      .click({ force: true })
+      .wait(2000);
```
parostatkiem, line 156: You could use `get` or `find` below instead. Cypress will retry looking for the desired element.

tests/cypress.json
```diff
@@ -3,9 +3,11 @@
   "video": false,
   "defaultCommandTimeout": 5000,
   "responseTimeout": 4000,
-
   "fixturesFolder": "fixtures",
   "integrationFolder": "tests",
+  "pageLoadTimeout": 10000,
```
parostatkiem, line 8: I think we should adjust all the timeouts to our *expectations* to also test if the performance of Busola isn't below the expected threshold. A page loading for 9.5s isn't a success, don't you think?

parostatkiem (changes requested):

Sawthis (changes requested):
Please look on Jan's comments.

tests/tests/busola-create-a-deployment.spec.js
```diff
@@ -0,0 +1,162 @@
...
+const NAMESPACE_NAME = `b-busola-test-${random}`;
+
+const DOCKER_IMAGE = 'eu.gcr.io/kyma-project/pr/orders-service:PR-162';
+const DEPLOYMENT_NAME = 'orders-service';
+
+context('Busola - Create a Deployment', () => {
+  const getLeftNav = () => cy.get('nav[data-testid=semiCollapsibleLeftNav]');
+
+  before(() => {
+    cy.visit(ADDRESS)
```
Sawthis, line 19: You need to adjust login flow. Example is here: https://github.com/kyma-project/busola/blob/main/tests/tests/busola-smoke-tests.spec.js#L18

Sawthis (changes requested):

PR #144, opened 2021-05-12
Handle roles with restricted permission set

Wawrzyn321:
**Description**

Important assumption: if there is a functionality explicitly listed on user's permission list, we assume this functionality actually exists on a cluster.

- Rename `apiGroups` to `apiPaths` to get closer to the Truth
- Move "init values" to proper file, avoiding shared global state
- Move `selfSubjectReview` query before other calls to check the permissions.
- Stop polling in `useGet` when 403 is received.
- More about permissions in Kyma 2.0 - https://sap-btp.slack.com/archives/C01N87ZQ34J/p1618479703004900

To test with `kymaopensource` account, grab full login and password from [slack](https://sap-btp.slack.com/archives/G01KBPVEXB9/p1611832379008700). You can assign roles [here](https://apskyxzcl.accounts400.ondemand.com/admin/#/users/P000003/details).

Scenarios:

- Kubeconfig with embedded token:
	- It has worked before - we have access to anything, so '/' call will be successful.
	- Checked by `hasWildcardPermissions` function.
	- User has access to all resources in all apiGroups, so in `filterNodesByAvailablePaths` we just need to check if that particular resource exists
- `runtimeAdmin` (login with your personal account via OIDC)
	- It doesn't have wildcard '*' role, we need to filter the permissions.
- `view` user (`kc get clusterrole view`)
	- This user can't even query '/' to get the apiGroups!
	- We need to resolve their permissions to show/hide nodes in Busola. We'll use the same algorithm for `runtimeAdmin`.

Test scenarios:
- Test with gardener kubeconfig or login with `runtimeAdmin` and `clusterAdmin` roles:
	- Kyma cluster: you should be able to see all resources, business as usual, see `ApiRules` on `Service` details
	- `clean` (no-kyma) cluster: you should be able to see k8s-only resources, no lists on `Service` details
- Test with k3d cluster:
	- You should see only installed resources
- Test with non-existing role (e.g. `runtimeNamespaceAdmin` on `clean` cluster)
	- You should see be redirected to `preferences` view, because all resources are unavailable.
- Test with `view` role (`kmain`, `clean`), `function-but-no-apirule` (`kmain`, `clean`), `some-ns-only` (`kmain`)

You can test your own 

A problem:
- If role doesn't exists on a cluster, user will receive only 3 permissions: 2x ['use'] and one for `selfSubjectX` queries. It means they will be only able to enter `preferences` node. I wanted to detect that and navigate them to `no-permissions`, but I didn't feel like adding a flaky check. Yikes.

**Related issue(s)**


dariadomagala (changes requested):
When I'm logged in using a Kubeconfig without a token, I can only see the permissions view.
It doesn't matter what groups I have assigned - it doesn't work even if I have runtimeAdmin, runtimeNamespaceAdmin, and cluster-admin user groups. 

Additionally - If I get `500` error on the Self Subject Rules Review call (I was still connected to a cluster that doesn't exist anymore), I get a blank page and cannot do anything, I had to clear my storage to add a config.

dariadomagala (approved):

PR #174, opened 2021-05-27
Restructuring enkode

Wawrzyn321:
**Description**

- Restructure enkode (http://pr.enkode.surge.sh/)
- Rename _systemNamespaces_ to _hiddenNamespaces_
- Make sure `hiddenNamespaces` & `disabledNavNodes` are lists instead of space separated strings 🐍 
- Adjust busola to enkode

TODOs for next tasks:
- For now, we use only the first cluster from the given kubeconfig. We could introduce _context chooser_, just like in Lens.

Test scenarios:
- Upload/paste kubeconfig with token - smooth sailing
- Former, but remove the token - the UI for filling OIDC should pop up.
- Enkode - use unmodified config in Busola - there's no cluster, so this should navigate use to "Add cluster" with "config set, gimme more params"
- Enkode - upload KK with token, paste to URL - that's all
- Enkode - upload KK without token, setup OIDC section - OIDC flow should kick in
- Enkode - upload KK without token, remove auth section - we should see "gimme auth", with kubeconfig already filled.

UP TO DISCUSS:
- We can see things like `currentContext.user.user` or `currentContext.cluster.cluster` all over the code. I decided not to change it, as it resembles original kubeconfig structure of e.g.

```
users: [
	- name: '...'
	  user: {
         token: '...'
      }
    ...
]
```

Of course, we can flatten it and don't repeat `cluster` or `user`.
- I decided not to touch this `modules` section in this task.

**Related issue(s)**


core/src/luigi-config/init-params.js
```diff
@@ -18,60 +27,55 @@ function getResponseParams(usePKCE = true) {
...
-    );
     const params = {
       ...decoded,
       config: {
         ...decoded.config,
-        systemNamespaces,
         modules: { ...DEFAULT_MODULES, ...(decoded.config?.modules || {}) },
       },
+      currentContext: {
+        cluster: decoded?.kubeconfig?.clusters[0],
```
dariadomagala, line 56: If someone would accidentally add an incorrect kubeconfig object where for example the `clusters`/`users` field doesn't exist, then there is a blank page, no message, no redirection, no info for the user.
Probably, the `Luigi.ux().showAlert()` is called too early for Luigi.

dariadomagala (commented):
The ```Luigi.ux().showAlert() is only available inside your configuration, after the configuration was initialized with Luigi.setConfig().``` error is shown in the browser console if I add the initparams but without the kubeconfig.

dariadomagala (approved):

PR #182, opened 2021-06-01
Add step with search of API Rule

Sawthis:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/tests/tests/create-a-function.spec.js
+++ b/tests/tests/create-a-function.spec.js
@@ -35,8 +35,20 @@ context('Create a Function and access it', () => {
   let apiRuleUrl;
   it('Get Host value for the API Rule', () => {
     cy.getLeftNav()
+      .contains('Discovery and Network')
+      .click();
+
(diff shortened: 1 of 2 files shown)


Wawrzyn321:
Retesting, as UI test failed on some invisible elements.

tests/support/views/api-rules-view.js
```diff
@@ -36,3 +36,25 @@ Cypress.Commands.add('createApiRule', (ApiRuleName, ApiRuleHost) => {
     .contains('Discovery and Network')
     .click();
 });
+
+Cypress.Commands.add('checkApiRuleStatus', ApiRuleName => {
```
akucharska, line 40: Shouldn't we move it to support/commands file?

Sawthis: it's connected to api rules view so I think it can stay here

akucharska (approved):

PR #186, opened 2021-06-02
Add Kubeconfig download

dariadomagala:
**Description**

- Add kubeconfig download in the context switcher and clusters list
- Add oidc data to the kubeconfig if token is missing
- Add UI tests

**Related issue(s)**


akucharska (commented):
I tested it on cluster and it works well. I didn't have time to check code.

Although I encountered one random behavior. The whole app crushed when I had 2 clusters defined (and one of them was added from old kubeconfig). Unfortunately I wasn't able to replicate this so this might be unrelated.

valentinvieriu:
I'm getting a strange error when I use enkode params
image

tests/tests/download-a-kubeconfig.spec.js
```diff
@@ -0,0 +1,33 @@
+/// <reference types='cypress' />
+
+const path = require('path');
+const downloadsFolder = Cypress.config('downloadsFolder');
+const kubeconfigPath = path.join(downloadsFolder, 'kubeconfig.yaml');
```
akucharska, line 5: Rename the variables to uppercase

akucharska (commented):
Nice :)

One small comment below

dariadomagala:
@valentinvieriu your problem is caused by the old init params (with the old schema) still saved in the session storage. Please remove session storage and try it out again.

akucharska (approved):

PR #202, opened 2021-06-09
Improve useGetHook

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


shared/hooks/BackendAPI/useGet.js
```diff
@@ -5,37 +5,44 @@ import { useFetch } from './useFetch';
...
         if (typeof onDataReceived === 'function') onDataReceived(payload.items);
-        if (error) setError(null); // bring back the data and clear the error once the connection started working again
-        processDataFn(payload, currentData, setData, lastResourceVersion);
+        if (error) setImmediate(_ => setError(null)); // bring back the data and clear the error once the connection started working again
+        setImmediate(_ =>
+          processDataFn(payload, currentData, setData, lastResourceVersion),
+        );
       } catch (e) {
-        processError(e);
+        setImmediate(_ => processError(e));
```
dariadomagala, line 41: ```This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.```

dariadomagala: ```This method is not expected to become standard, and is only implemented by recent builds of Internet Explorer and Node.js 0.10+. It meets resistance both from Gecko (Firefox) and Webkit (Google/Apple).```

dariadomagala: Do we have to use this function?

parostatkiem: Well, I couldn't find any other solution...
You're right, most browsers [refuse to support it ](https://caniuse.com/setimmediate)(although it works well in Chrome and Firefox; somebody's lying here...)

dariadomagala (approved):

PR #205, opened 2021-06-09
Add padding to service ports

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core-ui/src/components/Predefined/Details/Service/Service.details.scss
+++ b/core-ui/src/components/Predefined/Details/Service/Service.details.scss
@@ -1,4 +1,4 @@
-.fd-token {
+.service-ports.fd-token {
   display: inline-grid;
   grid-auto-columns: auto;
   grid-auto-flow: column;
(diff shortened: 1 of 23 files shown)


core-ui/src/components/Predefined/Details/Service/Service.details.scss
```diff
@@ -1,10 +1,10 @@
-.service__port-container.fd-token {
+.fd-token {
```
akucharska, line 1: Please don't overwrite fundamental styles globally

akucharska (changes requested):

core-ui/src/components/Predefined/Details/Service/Service.details.scss
akucharska: You can leave it as `.service-ports.fd-token {` and add class `className="service-ports" ` to the `<Token>`

akucharska (approved):

PR #209, opened 2021-06-10
Show a prettier metrics error

dariadomagala:
**Description**

❗ I don't know what's the issue exactly, the error is shown properly locally, but not on a cluster (no `statusText` as you can see [here](https://sap-btp.slack.com/archives/G01KBPVEXB9/p1623391666003400?thread_ts=1623325616.015700&cid=G01KBPVEXB9)). It's the same call that Piotrek also had a problem with, it behaves differently on a cluster.

If an error message doesn't exist, we show a custom message. Additionally, I improved the display of the error a bit. 

You can test it by deleting the metrics service (it will be recreated after few seconds)
``` bash
kubectl delete -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml
```

- ...
- ...
- ...

**Related issue(s)**


core-ui/src/components/Clusters/views/ClusterOverview/ClusterNodes/ClusterNodes.js
```diff
@@ -36,8 +34,15 @@ export function ClusterNodes() {
 
   return (
     <>
-      {loading && <Message content="Loading..." />}
-      {error && <Message content={error.message} />}
+      {loading && <Spinner compact={true} />}
+      {error && (
+        <LayoutPanel className="fd-margin--md">
```
parostatkiem, line 39: Maybe we could create a dedicated component to display data loading errors?

parostatkiem (commented):
What about displaying the status code in an unknown error case?

parostatkiem (approved):

parostatkiem (approved):

PR #220, opened 2021-06-15
Modal validation

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


parostatkiem (changes requested):
There's no "invalid" outline on the _scopes_ input but there is one on the _name_ input just above. I think you can add it to the _scopes_.
image

parostatkiem (approved):

PR #221, opened 2021-06-16
Fix too wide deployment list

parostatkiem:
**Description**

- ...
- ...
- ...

**Related issue(s)**


akucharska (commented):
Great job.
Can we also present a similar solution in the Apirules list (the host can be quite long) and service instances (long names in name and service class columns)

akucharska (approved):

PR #229, opened 2021-06-16
Prefill oidc scope with openid

wojtas1413:
**Description**

- Add default oidc scope
- Add AuthForm test cases
- ...

**Related issue(s)**


Wawrzyn321 (changes requested):
It works really nice, but we can make it _even_ better - why don't we query the scopes field by it's placeholder and just check the current value? Something like this:

```
const { getByPlaceholderText } = render(authForm);
expect(getByPlaceholderText('Scopes').value).toEqual(DEFINED_SCOPE_VALUE);
```

Using `{ render }` from `@testing-library/react`.

Wawrzyn321 (approved):

PR #231, opened 2021-06-17
Crds lists

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**


shared/components/ResourcesList/ResourcesList.js
```diff
@@ -172,7 +177,9 @@ function Resources({
       <Link
         className="link"
         onClick={_ =>
-          fixedPath
```
Wawrzyn321, line 175: I think we could use a function with early returns here, double ternary statement makes it _complicated_ instead of _complex_.

Wawrzyn321 (changes requested):
Could we also change the side nav width to avoid that ellipsis? 
```
--luigi__left-sidenav--width: 17rem;
```

core-ui/src/components/Predefined/Details/CustomResourceDefinitions/CustomResources.details.js
```diff
@@ -0,0 +1,87 @@
...
+import {
+  ResourceDetails,
+  useGet,
+  EMPTY_TEXT_PLACEHOLDER,
+  GenericList,
+  Spinner,
+} from 'react-shared';
+
+export function CustomResource({ params }) {
+  const namespace = LuigiClient.getContext().namespaceId;
```
Wawrzyn321, line 14:
```suggestion
  const { namespaceId: namespace } = useMicrofrontendContext();
```

Wawrzyn321:
One more thing - it would be nice to inspect `/shoot--hasselhoff--prod/namespaces/default/customresourcedefinitions/details/verticalpodautoscalercheckpoints.autoscaling.k8s.io` - there is 404 error here.
image
The same goes with `verticalpodautoscalercheckpoints.autoscaling.k8s.io`, but I guess the issue is the same.

Wawrzyn321:
What's more, when I refresh the page (e.g. http://localhost:4200/cluster/shoot--hasselhoff--prod/namespaces/default/customresourcedefinitions/details/certificaterevocations.cert.gardener.cloud), it stops working).
image

Wawrzyn321 (approved):

PR #233, opened 2021-06-18
Adjust auth2Client responseTypes validation

Wawrzyn321:
**Description**

It turns out `responseTypes` can be either `null` or non-empty string array - uploading an empty array results in 422 error.

- Handle fallback for falsy `responseTypes`, so that `namespaces/kyma-system/oauth2clients/details/eventing-controller` on kmain doesn't crash.
- Adjust validation for create - we don't require `responseTypes` anymore.

**Related issue(s)**


core-ui/src/components/Predefined/Create/OAuthClients/createOAuthClient.js
```diff
@@ -9,7 +9,8 @@ export const createOAuthClient = (namespace, spec) => {
     },
     spec: {
       grantTypes,
-      responseTypes,
+      // we can either specify a non-empty array, or null - empty array is forbidden
+      responseTypes: responseTypes.length ? responseTypes : null,
```
parostatkiem, line 13: `responseTypes.length` will throw an error if `responseTypes` is `null`.

parostatkiem (changes requested):

parostatkiem (approved):

PR #241, opened 2021-06-23
Add confirmation before connecting to the new cluster with init params

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


valentinvieriu:
We choose to implement this different by pulling the kubeconfig from a trusted server

PR #249, opened 2021-06-28
Config versioning

Wawrzyn321:
**Description**

Semver vs "major-only" versioning - in the end, I went for semver.

Scenarios:

- User uses params with incompatible version - show a warning.
- User has updated their Busola, and some of the existing configs are now outdated - show a warning, but don't delete the cluster.
- User adds a kubeconfig with no params - just add the last version to the params.

"version" field is already on [enkode](http://enkode.surge.sh/).

**Related issue(s)**

### DIFF
--- a/core/src/luigi-config/init-params/params-version.js
+++ b/core/src/luigi-config/init-params/params-version.js
@@ -1,13 +1,17 @@
 import { major, satisfies, coerce } from 'semver';
 import { PARAMS_VERSION } from './constants';
 
+export function formatVersion(version) {
+  const fullVersion = coerce(version);
+  return `${fullVersion.major}.0`;
+}
+
 export function areParamsCompatible(paramsVersion) {
-  const majorVersion = major(coerce(PARAMS_VERSION));
-  return satisfies(coerce(paramsVersion), `^${coerce(majorVersion)}`);
+  return satisfies(coerce(paramsVersion), `^${formatVersion(PARAMS_VERSION)}`);
 }
 
 export function showIncompatibleParamsWarning(paramsVersion) {
-  const minimalVersion = coerce(major(coerce(PARAMS_VERSION)));
+  const minimalVersion = formatVersion(PARAMS_VERSION);
   const message = `Configuration incompatible: version: ${paramsVersion ||
     "'unknown'"}, supported versions: ^${minimalVersion}. Errors may occur.`;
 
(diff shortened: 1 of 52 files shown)


core/src/luigi-config/utils/params-version.js
```diff
@@ -0,0 +1,23 @@
+import { major, satisfies, coerce } from 'semver';
+
+const PARAMS_VERSION = '1.0'; // make sure to sync it in core-ui
```
dariadomagala, line 3: can we move init-params-related constants to one file? (stuff like DEFAULT_MODULES etc that we have in shared)

core/src/luigi-config/utils/params-version.js
```diff
@@ -0,0 +1,23 @@
...
+
+export function areParamsCompatible(paramsVersion) {
+  const majorVersion = major(coerce(PARAMS_VERSION));
+  return satisfies(coerce(paramsVersion), `^${coerce(majorVersion)}`);
+}
+
+export function showIncompatibleParamsWarning(paramsVersion) {
+  const message = `Configuration incompatible: version: ${paramsVersion ||
+    "'unknown'"}, supported version: ${PARAMS_VERSION}. Errors may occur.`;
```
dariadomagala, line 12: Maybe something like this since we support all minor versions?
```suggestion
    "'unknown'"}, supported versions: ^${PARAMS_VERSION}. Errors may occur.`;
```

core/src/luigi-config/init-params/params-version.js
```diff
@@ -1,15 +1,15 @@
...
 export function areParamsCompatible(paramsVersion) {
   const majorVersion = major(coerce(PARAMS_VERSION));
   return satisfies(coerce(paramsVersion), `^${coerce(majorVersion)}`);
 }
 
 export function showIncompatibleParamsWarning(paramsVersion) {
+  const minimalVersion = coerce(major(coerce(PARAMS_VERSION)));
   const message = `Configuration incompatible: version: ${paramsVersion ||
-    "'unknown'"}, supported version: ${PARAMS_VERSION}. Errors may occur.`;
+    "'unknown'"}, supported versions: ^${minimalVersion}. Errors may occur.`;
```
dariadomagala, line 12: now it shows `supported versions: ^1.0.0` yet we don't have the `1.0.0` version

dariadomagala (approved):

PR #259, opened 2021-07-01
Features from params

Wawrzyn321:
**Description**

Use with [new enkode](pr.enkode.surge.sh).

- Introduce features, that can be controlled based on various conditions - for now, I've added `apiGroup` one and `configMapJsonPath`. Others could be a simple single value, config map string (not object)
- Some of them the are lazy-evaluated - we wouldn't want to fire a lot of requests while starting the app, would we?

**Related issue(s)**


core/src/luigi-config/init-params/constants.js
```diff
@@ -14,7 +14,6 @@ export const DEFAULT_FEATURES = Object.fromEntries(
     {
       selectors: [
         {
-          lazy: false,
```
valentinvieriu: if we use isDefault, should we also use isLazy ?

core/src/luigi-config/features.js
```diff
@@ -0,0 +1,54 @@
...
+    data.crds.some(crd => crd.includes(selector.apiGroup)),
+  configMapJsonPath: async (selector, data) => {
+    const { namespace, name, entry, jsonPath, expectedValue } = selector;
+
+    const cmUrl = `${config.backendAddress}/api/v1/namespaces/${namespace}/configmaps/${name}`;
+    const response = await failFastFetch(cmUrl, data.authData);
+    const configMap = await response.json();
+
+    const value = jsonpath.query(
+      JSON.parse(configMap.data[entry]),
```
dariadomagala, line 16: config map doesn't have to be JSON and it fails here if it's not

dariadomagala: also, this `jsonpath.query` might fail and we don't show why exactly, we should wrap this whole thing in a try/catch

dariadomagala (changes requested):

core/src/luigi-config/features.js
```diff
@@ -0,0 +1,54 @@
...
+import { config } from './config';
+import jsonpath from 'jsonpath';
+
+const resolvers = {
+  apiGroup: (selector, data) =>
+    data.crds.some(crd => crd.includes(selector.apiGroup)),
+  configMapJsonPath: async (selector, data) => {
+    const { namespace, name, entry, jsonPath, expectedValue } = selector;
+
+    const cmUrl = `${config.backendAddress}/api/v1/namespaces/${namespace}/configmaps/${name}`;
```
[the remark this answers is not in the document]
Wawrzyn321: Well, that's why I've introduced the "lazy" option.

I did thought about caching the CMs somehow, or (as you pointed out) getting all of them at once, but all CMs in all namespaces could be large, and what about values from secrets or other resources that could be required here?

I guess we need to discuss it more.

core/src/luigi-config/features.js
Wawrzyn321: I've added a try/catch block with appropriate message, good point.

About the JSON... We assume here that the CM value will be a valid JSON, as we use JsonPath.

core/src/luigi-config/features.js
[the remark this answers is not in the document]
dariadomagala: secrets? afaik we only support CMs for now?

PR #264, opened 2021-07-06
Add Observability links

### LINKED ISSUE #255: Introduce the Observability Component
**Description**

As [Matthew](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#matthew---cloud-native-software-engineer) or [Sebastian](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#sebastian---tech-lead) I would like to have have the links to the observability stack inside Busola

**Reasons**

Observability stack is a common stack that is needed for a Dashboard, we need to have a flexible way to be able to identify if what observability applications are running on the cluster, and what are their urls.

**Acceptance criteria**

- [x] The selection path and the label of the tool are part of the Busola Config, and they have sensitive defaults, can be overwritten by enkode params and Busola Cluster Config 
- [x] We use the Kyma defaults, that are using VirtualServices(specs/host) to identify if the tools are available and what urls they have
- [x] If multiple hosts are present, we pick the first one in the list
- [x] If No tools is available, the whole observability section is hidden
- [x] We are going to add the Observability section above the External Nodes section
- [x] We target for beginning Grafana, Tracing  and Kiali

parostatkiem:
**Description**

- Add `Observability` section only if there's at leas one nav node to display
- Allow user to provide the Title and VirtualService link via Config Params
- Default to the Busola ConfigMap if the `Observability` is not configured via Config Params
- Ignore the navigation node if the virtualService it's linked to doesn't exist (`host` value cannot be read)

**Related issue(s)**
Resolves #255


core/src/luigi-config/navigation/navigation-data-init.js
```diff
@@ -221,6 +223,35 @@ async function fetchNavigationData(authData, permissionSet) {
   }
 }
 
+async function getObservabilityNodes(authData, enabledFeatures) {
```
dariadomagala, line 226: can we move the logic to the `features.js` file?

core/src/assets/config/config.json
```diff
@@ -1,62 +1,80 @@
...
+      },
+      "SERVERLESS": {
+        "selectors": [
+          {
+            "type": "apiGroup",
+            "apiGroup": "serverless.kyma-project.io"
+          }
+        ]
+      },
+      "OBSERVABILITY": {
```
dariadomagala, line 60: we should add the `isEnabled` flag here so if someone doesn't want the default links, he/she can disable it

core/src/luigi-config/navigation/queries.js
```diff
@@ -51,6 +51,13 @@ export async function failFastFetch(input, auth, init = {}) {
   }
 }
 
+export async function fetchObservabilityHost(auth, vsPath) {
+  const res = await failFastFetch(config.backendAddress + '/' + vsPath, auth, {
+    method: 'GET',
+  });
+  return (await res.json()).spec.hosts[0];
```
dariadomagala, line 58: do we know if it's always the first port?

core/src/luigi-config/navigation/navigation-data-init.js
```diff
@@ -221,6 +223,35 @@ async function fetchNavigationData(authData, permissionSet) {
...
+    enabledFeatures.OBSERVABILITY?.config.links || // take the custom config at first
+    (await getClusterParams()).config.features.OBSERVABILITY?.config.links; // use the Busola configMap as a fallback
+
+  const CATEGORY = {
+    label: 'Observability',
+    icon: 'stethoscope',
+    collapsible: true,
+  };
+  const navNodes = await Promise.all(
+    links.map(async ([label, path]) => {
```
dariadomagala, line 237: what if links are empty? the config map can be edited to not have this field, and then the whole app is unusable

core/src/assets/config/config.json
```diff
@@ -1,62 +1,80 @@
...
+        "config": {
+          "links": [
+            [
```
dariadomagala, line 63: why are those arrays? I think it would be more readable to have them as objects with `label` and `path`

core/src/assets/config/config.json
dariadomagala: the second question would be if we should even have the defaults? we want to create the product for all Kubernetes clusters, not only Kyma, so maybe we shouldn't have anything if those are not specifically defined?

dariadomagala (changes requested):

core/src/assets/config/config.json
dariadomagala: Also, please add those to our `configmap.yaml`

dariadomagala (changes requested):

core/src/luigi-config/navigation/navigation-data-init.js
parostatkiem: But it's mostly about adding nodes

core/src/assets/config/config.json
parostatkiem: @valentinvieriu

core/src/luigi-config/navigation/queries.js
parostatkiem: Please look at the description of this https://github.com/kyma-project/busola/issues/255

core/src/luigi-config/navigation/navigation-data-init.js
parostatkiem: I thought we agree to use some kind of schema for the config... If not, then what if the `config` is empty or is not an object type or the `links` are not strings or...

core/src/assets/config/config.json
valentinvieriu: yes, we should add `isEnabled` on all our features

core/src/luigi-config/navigation/navigation-data-init.js
valentinvieriu: That's a good point there, I think we have a JSON schema fo our config already

core/src/luigi-config/navigation/queries.js
valentinvieriu: It's an assumption. We will see if this is not correct

core/src/assets/config/config.json
valentinvieriu: I agree with @dariadomagala . Can we have and array of objects?

core/src/assets/config/config.json
valentinvieriu: I agree to @dariadomagala . Our defaults should work with any K8s cluster by default. We should use the overwrites to enable the features that are specific to Kyma.

parostatkiem: Sooo... I should remove it from `config.json` and leave it only in `configmap.yaml`? If yes, I'm afraid our developer experience will suffer as we'll have to manually add this section to our `config.json` and never include it in a commit..

core/src/luigi-config/navigation/navigation-data-init.js
parostatkiem: I don't think we do. If we decided to create one, I suggest to do this in a separate task.

core/src/assets/config/config.json
```diff
@@ -60,18 +60,18 @@
...
-            ],
-            [
-              "Kiali",
-              "apis/networking.istio.io/v1beta1/namespaces/kyma-system/virtualservices/kiali"
-            ],
-            [
-              "Tracing",
-              "apis/networking.istio.io/v1beta1/namespaces/kyma-system/virtualservices/tracing"
-            ]
+            {
```
dariadomagala, line 63: what about the `isEnabled` flag? 
I'm thinking if we should have this flag for the whole OBSERVABILITY feature only or should we also flag every single link 🤔

core/src/assets/config/config.json
dariadomagala: Maybe we could keep it in the config map just with the `isEnabled` flag set to false by default and change it in the config map if needed?

dariadomagala (changes requested):

core/src/luigi-config/navigation/navigation-data-init.js
```diff
@@ -224,10 +224,19 @@ async function fetchNavigationData(authData, permissionSet) {
 }
 
 async function getObservabilityNodes(authData, enabledFeatures) {
-  const links =
-    enabledFeatures.OBSERVABILITY?.config.links || // take the custom config at first
-    (await getClusterParams()).config.features.OBSERVABILITY?.config.links; //  use the Busola configMap as a fallback
+  console.log();
```
dariadomagala, line 227: please remove

dariadomagala (changes requested):
something is broken now. Calls go to ```https://busola.DOMAIN/backend/undefined```

dariadomagala (approved):

dariadomagala (approved):

dariadomagala (approved):

PR #273, opened 2021-07-07
Unify dropdowns

dariadomagala:
**Description**

- Bump fundamental to 0.14-rc100000000
- Use one shared component for all dropdowns
- Fix the issue with deleting Bindings from Instance Details
- Fix the issue of always disabled 'Add Service Binding' buttons (it expected bindable Instances to exist for the button to be enabled, but the call for the instances was triggered after clicking the button..)
- Fix the issue of not being able to choose a different auth method for the Git Repository

note: `Select` component doesn't currently pass the `ref` properly to the input (https://github.com/SAP/fundamental-react/issues/1336), it just passes it to the wrapper. I still used refs in most places as hopefully this will be fixed, and it would be enough to just remove few lines of workarounds.

**Related issue(s)**


core-ui/src/components/ApiRules/ApiRuleForm/AccessStrategyForm/AccessStrategyForm.js
```diff
@@ -60,34 +83,12 @@ export default function AccessStrategyForm({
...
-                  handleFormChanged();
-                }}
-              >
-                {Object.values(accessStrategyTypes).map(ac => (
-                  <option key={ac.value} value={ac.value}>
-                    {ac.displayName}
-                  </option>
-                ))}
-              </FormSelect>
+              <Dropdown
```
akucharska, line 86: This dropdown on apirules page is lower than others elements in the row (have margin-top:4px)

core-ui/src/components/ApiRules/ApiRuleForm/AccessStrategyForm/test/AccessStrategyForm.test.js
```diff
@@ -83,7 +83,7 @@ describe('AccessStrategyForm', () => {
     expect(queryByRole('alert')).not.toBeInTheDocument();
   });
 
-  it('renders OAuth2 strategy', () => {
```
akucharska, line 86: Will we adjust the tests?

akucharska (approved):

PR #285, opened 2021-07-15
remove title dependency from ResourceCircle

qbalukom:
**Description**

- clean up ResourceCircle not to depend on the human-readable "title"


akucharska (approved):
Very good job

PR #296, opened 2021-07-22
Sanitize and check backend headers and errors

parostatkiem:
**Description**
I had no idea which behaviour will be good in case of a vulnerability detected (`\` in headers or non-alphanumeric error). Please feel free to leave some suggestions!

- ...
- ...
- ...

**Related issue(s)**


backend/common.js
```diff
@@ -52,15 +52,23 @@ export const handleRequest = async (req, res) => {
...
         'Content-Type': k8sResponse.headers['Content-Type'] || 'text/json',
         'Content-Encoding': k8sResponse.headers['content-encoding'] || '',
       });
-
       k8sResponse.pipe(res);
     })
     .on('error', function(err) {
       console.error('Internal server error thrown', err);
+
+      if (!err.match(/^[0-9a-zA-Z]$/)) err = '';
```
akucharska, line 70: You can try to return the error after it is sanitized.
You may check:
https://www.npmjs.com/package/string-sanitizer
or
https://developer.mozilla.org/en-US/docs/Web/API/Sanitizer/sanitizeToString

parostatkiem: I think we can allow the following chars `:{},"[]`. This will allow JSON error to be thrown which we handle on the frontend. What else do we possibly expect other than plain text or JSON? I think every other format of the error could be potentially malicious and there's no point to sanitize it.

dariadomagala (approved):

PR #298, opened 2021-07-26
i18n PoC

qbalukom:
**Description**

- i18n proof of concept


NHingerl:
Two quick ideas that came up chatting with @Sawthis: 
- I'd suggest to support providing contextual information (something. like `description`, `context`, `comment`, `developer_comment`) for each string, as described for object-based/array-based YAML in the [localazy documentation](https://localazy.com/docs/cli/yaml-format#contextual-metadata). 
- Two stackoverflow discussions ([first](https://stackoverflow.com/questions/10282857/how-do-you-structure-i18n-yaml-files-in-rails) and [follow-up](https://stackoverflow.com/questions/11097572/refactoring-ruby-on-rails-i18n-yaml-files-using-dictionaries)) discuss the pro and cons of using "dictionaries" of reoccurring terms. We probably don't need it now but it's an interesting idea to keep in mind for the future.

qbalukom:
@NHingerl 1. I'm not sure about that, as I believe most gui editors will treat each such string as a separate translation item. So instead
2. second is a good idea and can be seen in action in en.yaml - for example `namespaces.overview.workloads.pods` actually uses the value of `pods.title`

dariadomagala (approved):

PR #299, opened 2021-07-26
Set Editor theme, adjust lambda details view to theming, fix YAML drawer buttons

dariadomagala:
**Description**

- Set Editor theme
- Adjust lambda details view to theming
- Fix YAML editor drawer buttons

**Related issue(s)**


shared/contexts/ThemeContext.js
```diff
@@ -24,18 +24,30 @@ function addLinkNode() {
...
     if (typeof env.PUBLIC_URL === 'undefined') return;
     applyThemeToLinkNode(theme, env.PUBLIC_URL);
     localStorage.setItem('busola.theme', theme);
     LuigiClient.sendCustomMessage({ id: 'busola.theme', name: theme });
+    setEditorTheme(getEditorTheme(theme));
   }, [theme]);
 
   return (
-    <ThemeContext.Provider value={{ theme, setTheme }}>
+    <ThemeContext.Provider value={{ theme, editorTheme, setTheme }}>
```
parostatkiem, line 50: I think we can save some renders here by not using a dedicated state for the `editorTheme`
```suggestion
    <ThemeContext.Provider value={{ theme, editorTheme: getEditorTheme(theme), setTheme }}>
```

parostatkiem (changes requested):

parostatkiem (approved):

PR #315, opened 2021-08-02
Move menu texts to i18n yaml file

Sawthis:
**Description**

- Move menu texts to i18n yaml file
- Improve tooltip text

**Related issue(s)**

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -50,6 +50,17 @@ instances:
   title: Instances
 integration:
   title: Integration
+jobs:
+  completion-time: Completion Time
+  completions: Completions
+  conditions:
+    last-probe: Last Probe
+    last-transition: Last Transition
+    status: Status
+    title: Conditions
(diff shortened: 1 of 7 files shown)


core/src/i18n/en.yaml
```diff
@@ -1,40 +1,89 @@
...
-service-catalog:
+replica-sets:
+  title: Replica Sets
+role-bindings:
+  title: Role Bindings
+roles:
+  title: Roles
+secrets:
+  title: Secrets
+service-management:
```
dariadomagala, line 82: Maybe we could split those into categories as well?

something like:
```
service-management:
  title: Service Management
  children:
  - catalog:
    title: Service Catalog
  - ...
```
or 
```
categories:
 - service-management: ...
subcategories:
- catalog: ...
```
Just ideas to discuss, not suggestions. It just confuses me that 2 different things are on the same level.

dariadomagala (approved):

PR #316, opened 2021-08-03
Delete confirmation

qbalukom:
**Description**
- added delete confirmation dialog to all ResourceLists
- added preferences option for confirming item deletion
- added more universal luigi feature tggles handling and converted namespaces to it


core/src/luigi-config/luigi-config.js
```diff
@@ -47,7 +47,7 @@ async function luigiAfterInit() {
     return;
   }
 
-  loadHiddenNamespacesToggle();
+  readFeatureToggles(['dontConfirmDelete', 'showHiddenNamespaces']);
```
Wawrzyn321, line 50: I think we should abstract away to `readAllFeatureToggles` - in this place we don't care about any particular toggle, we just want them all.

Wawrzyn321:
Should we show the message also on other lists (e.g. API Rules on Service details)? If so, it could be nice to introduce this GenericList paired up with Dialog as a shared component.

Wawrzyn321:
If I check the "Don't show delete confirmation" option while deleting something, it works fine (I won't see the confirmation again, nice), but it isn't checked on preferences view... for the first time. If I get back (without refreshing) to the resources list and then again, on preferences - it's checked.

core/src/luigi-config/luigi-config.js
qbalukom: the problem is we don't have a list of all toggles stored anywhere - so we need to provide a list of what to read from localStorage and assign to Luigi

qbalukom:
> Should we show the message also on other lists (e.g. API Rules on Service details)? If so, it could be nice to introduce this GenericList paired up with Dialog as a shared component.
AFAIK (but I might be mistaken), service catalog is in more or less maintenance mode. And since it's going to be reimplemented eventually, hopefully we'll implement it using ResourceList instead

Wawrzyn321:
The Service on API Rules does doesn't come from Service Catalog (_how shocking!_), it's the native k8s resource.

core/src/i18n/en.yaml
```diff
@@ -50,3 +50,14 @@ cron-jobs:
   not-scheduled-yet: None scheduled yet
 settings:
   language: Language
+  confirmations: Confirmations
+  dontConfirmDelete: Delete without confirmation
+test: Test!!
```
Wawrzyn321, line 55:
```suggestion

```

Wawrzyn321 (dismissed):

Wawrzyn321 (approved):

PR #318, opened 2021-08-03
Polish dark theming

parostatkiem:
**Description**

- Use theming CSS variables here and there instead of fixed colors
- remove the `Notification` component
- use Toast or Luigi.confirmationModal in `NotificationContext`

**Related issue(s)**


Wawrzyn321 (changes requested):
Lookin' nice, but 2 little things here:
1) Please make sure there's no sudden flash of white when user collapses left nav. Adding a color to `body. lui-semiCollapsible` should be enough.

2) I've tried out the new error message:
image

It looks like even falsy values are stringified here - `undefined` and `null` are also displayed.

Wawrzyn321 (dismissed):

Wawrzyn321 (dismissed):

Wawrzyn321 (approved):

PR #319, opened 2021-08-04
Job workloads

qbalukom:
**Description**

- add job list and details
- create a common `ResourcePods` component for pod lists
- **important** use the `ResourcePods` for replica sets - not 100% sure if the results are correct, as the original fetched everything and filtered manually (but since replica sets have all the necessary references, and I didn't notice any differences, I assume it's correct)


Wawrzyn321:
Awesome work dude!

Wawrzyn321 (approved):

Wawrzyn321:
I've approved this and added the _don't-merge_ label - what do you think about adding a reference to parent on Job's details view? We could take it's `ownerReference` and point to Function (building jobs), to CronJob or don't show this column at all (for example Job `hb-gcp-se-defaul-39fd4ba1-46e4-4415-b2ae-22290e71d533-gcp-servi` in default namespace on `kmain` cluster doesn't have an owner ref).

If you don't think it's worth it, just remove my label.

PR #332, opened 2021-08-11
enable luigi backdrop and add info

qbalukom:
**Description**
- add luigi overlay when dialog is open
- add an information box when disabling confirmations

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -61,9 +61,7 @@ cron-jobs:
       replace: No concurrent executions, replace previous run with the new one.
       skip: No concurrent executions, skip new run if the previous one hasn't finished yet.
     title: Concurrency Policy
-  last-job-execution: Last J>>>>>>> main
-45
-ob execution
+  last-job-execution: Last Job execution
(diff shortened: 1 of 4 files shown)


Wawrzyn321 (approved):

Wawrzyn321 (approved):

Sawthis (changes requested):
Please add Polish translation as well.

core/src/i18n/en.yaml
```diff
@@ -37,6 +37,7 @@ common:
     delete-confirm: Don't ask this again
     message: Are you sure you want to remove {{type}} {{name}}?
     title: Remove {{name}}
+    information: Delete confirmations will be disabled for all resources. It can be turned back on in Preferences.
```
NHingerl, line 40:
```suggestion
    information: You've switched off delete confirmations for all resources. You can switch them back on in **Preferences**.
```
I suggest active voice and no future tense. Not sure whether we can/want to use bold type here, thus only a suggestion, not editing myself.

Sawthis (approved):

NHingerl (approved):

PR #333, opened 2021-08-11
Move preferences view to modal

Wawrzyn321:
**Description**

- Move Preferences to microfrontend-based modal.
- Allow to open Preferences without a cluster chosen.

**Related issue(s)**

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -109,12 +109,19 @@ namespaces:
     create: Select this option to disable istio to mediate all communication between the pods in your namespace.
     memory-examples: Use plain value in bytes (128974848) or suffix equivalents (like 129e6, 129M, 123Mi).
     total-memory-limit: Define constraints that limit total memory consumption in your namespace.
+navigation:
+  profile:
+    preferences: $t(preferences.title)
+    remove-current-cluster-config: Remove current Cluster Config
+    download-current-cluster-config: Download current Cluster Config
 oauth2-clients:
   title: OAuth2 Clients
 observability:
   title: Observability
 pods:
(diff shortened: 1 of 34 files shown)


core/src/luigi-config/navigation/navigation-data-init.js
```diff
@@ -79,10 +79,16 @@ async function createClusterManagementNodes() {
           return false;
         },
       },
+      {
+        pathSegment: 'preferences',
+        viewUrl: config.coreUIModuleUrl + '/preferences',
+        openNodeInModal: { title: 'Preferences', size: 'm' },
```
dariadomagala, line 85: I believe we should start moving such texts into the i18n files

core/src/i18n/en.yaml
```diff
@@ -125,6 +125,11 @@ namespaces:
     create: Select this option to disable Istio to mediate all communication between the Pods in your Namespace.
     memory-examples: Use plain value in bytes (128974848) or suffix equivalents (like 129e6, 129M, 123Mi).
     total-memory-limit: Define constraints that limit total memory consumption in your Namespace.
+navigation:
```
dariadomagala, line 128: maybe a `profile`/`profile-dropdown` instead of `navigation`? that's how the dropdown is called in the sap/luigi world :)

dariadomagala (approved):

dariadomagala (changes requested):
please add also polish translations

dariadomagala (dismissed):

Sawthis (approved):

NHingerl (approved):

akucharska (approved):

PR #338, opened 2021-08-12
use MessageBox instead of Dialog

qbalukom:
**Description**
- use the new MessageBox element for delete confirmation dialog

**Notes**
This will require merging of fundamenta-react and bumping it's version


dariadomagala (changes requested):
It's not a problem with MessageBoxe itself, but probably with the new fundamental. 

Every modal doesn't have its background, we only have the Luigi one. MessageBox has the background.
Screen Shot 2021-08-18 at 15 23 02

dariadomagala (approved):

PR #340, opened 2021-08-12
Extra services info

qbalukom:
**Description**
addded "Service Type" and "External IPs" to service list and details


core-ui/src/components/Predefined/List/Services.list.js
```diff
@@ -1,29 +1,41 @@
...
-      return '';
+      return `${port.port}:${port.targetPort}/${port.protocol}`;
     }
   };
 
+  const getPorts = service =>
+    intersperse(service.spec.ports?.map(getPortString) || [], ', ');
+
+  const getExternalIPs = service =>
+    intersperse(service.spec.externalIPs || [], ', ');
```
Wawrzyn321, line 21: Wouldn't using the `Array<string>::join(', ')` be just enough? I know one evaluates to an array (of strings in this case), the `join` returns a single string, but for rendering on UI the result will be the same.

core-ui/src/components/Predefined/Details/Service/Service.details.js
```diff
@@ -90,6 +96,16 @@ export const ServicesDetails = ({ DefaultRenderer, ...otherParams }) => {
...
       ),
     },
+    {
+      header: t('services.external-ips'),
+      value: service => (
+        <ul>
+          {service.spec.externalIPs?.map(ip => (
+            <li key={ip}>{ip}</li>
+          ))}
+        </ul>
```
Wawrzyn321, line 106: Nitpicking time! When the value is empty, we use '-' as a placeholder, so please use it in `externalIPs` field on both details and list.

Wawrzyn321: Example - when there are no labels.

core-ui/src/components/Predefined/List/Services.list.js
qbalukom: technically it came from a few ideas along the way... didn't feel like getting rid of intersperse - could be useful in the future :)

Wawrzyn321 (approved):

Wawrzyn321 (approved):

NHingerl (approved):

PR #342, opened 2021-08-13
Storage configuration

Wawrzyn321:
**Description**

- Handle 3 cluster storage types (localStorage, sessionStorage, in memory)
- Small adjustments on Preferences visual side

**Related issue(s)**


core-ui/src/components/Clusters/views/ClusterStorageType.js
```diff
@@ -0,0 +1,38 @@
...
+import { StatusBadge } from 'react-shared';
+
+export function ClusterStorageType({ clusterConfig }) {
+  const { t } = useTranslation();
+
+  const storage = clusterConfig?.storage;
+
+  const descriptions = {
+    localStorage: {
+      name: 'localStorage',
```
dariadomagala, line 12: same as before, should we translate it?

core-ui/src/components/Clusters/views/ClusterStorageType.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    inMemory: {
+      name: t('clusters.storage.labels.inMemory'),
+      tooltip: t('clusters.storage.descriptions.inMemory'),
+    },
+  };
+
+  const description = descriptions[storage] || {
+    name: t('clusters.storage.labels.unknown'),
+    description: t('clusters.storage.descriptions.unknown'),
```
dariadomagala, line 27:
```suggestion
    tooltip: t('clusters.storage.descriptions.unknown'),
```

core/src/i18n/en.yaml
```diff
@@ -46,12 +46,34 @@ cluster-role-bindings:
...
   list:
     outdated:
       download: Download Kubeconfig
       tooltip: Outdated parameter version may cause errors. Delete and re-add your cluster.
+    no-clusters-found: No clusters found
   overview:
     title: Cluster Overview
+    version: Version
+  storage:
+    changed-message: Cluster storage changed from {{originalStorage}} to {{targetStorage}}.
```
dariadomagala, line 65: I think we add more info on why it happened. Something similar to `The chosen storage type has been overwritten by the Cluster configuration from {{originalStorage}} to {{targetStorage}}.`

core/src/i18n/en.yaml
```diff
@@ -46,12 +46,34 @@ cluster-role-bindings:
...
+    choose-storage:
+      label: Configuration storage type
```
dariadomagala, line 67: `Storage type configuration`?

core/src/i18n/en.yaml
```diff
@@ -46,12 +46,34 @@ cluster-role-bindings:
...
+    descriptions:
+      localStorage: Cluster data is persisted between browser reloads.
+      sessionStorage: Cluster data is cleared when the page session ends.
+      inMemory: Cluster data is cleared when the page is refreshed or navigated out of.
```
dariadomagala, line 71: please sort the yaml files. (you can try https://marketplace.visualstudio.com/items?itemName=PascalReitermann93.vscode-yaml-sort )

core/src/luigi-config/cluster-management/clusters-storage.js
```diff
@@ -0,0 +1,72 @@
...
+        break;
+      case 'inMemory':
+        inMemoryClusters[clusterName] = cluster;
+        break;
+      default:
+        console.warn(
+          clusterName,
+          ': Unknown storage type ',
+          cluster?.config?.storage,
+          'saving in localStorage',
```
dariadomagala, line 37: please add translations to this message

core/src/luigi-config/cluster-management/clusters-storage.js
```diff
@@ -0,0 +1,72 @@
...
+  const targetStorage = (await getActiveCluster()).config.storage;
+  if (targetStorage !== originalStorage) {
+    // move the cluster to the valid storage
+    const clusters = load();
+    clusters[getActiveClusterName()].config.storage = targetStorage;
+    save(clusters, null);
+
+    if (originalStorage) {
+      Luigi.ux().showAlert({
+        text: i18next.t('clusters.storage.changed-message', {
```
dariadomagala, line 58: I think we should use the `closeAfter` option to close it after a few seconds since it's only information.

dariadomagala (changes requested):

core-ui/src/components/Clusters/views/AddCluster/ChooseStorage.js
```diff
@@ -0,0 +1,26 @@
...
+
+import { FormLabel, Select } from 'fundamental-react';
+import { useTranslation } from 'react-i18next';
+import './ChooseStorage.scss';
+
+export function ChooseStorage({ storage, setStorage }) {
+  const { t } = useTranslation();
+
+  const storages = [
+    { key: 'localStorage', text: 'localStorage' },
```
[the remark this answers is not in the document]
Wawrzyn321: Ha! I didn't want to translate it, as it's more like technical idiom. Similarly we don't translate Deployments, and nobody translates keywoards like _while_ (unless you're [Logomocja](https://www.sp4.warszawa.pl/informatyka1/logomocjawprowadzenie.html)).

core/src/luigi-config/cluster-management/clusters-storage.js
Wawrzyn321: Yeah, I was thinking about it... do we translate `console` functions? Users normally doesn't look here.

core/src/luigi-config/cluster-management/clusters-storage.js
Wawrzyn321: Good one.

core/src/i18n/en.yaml
Wawrzyn321: Damnn

core-ui/src/components/Clusters/views/AddCluster/ChooseStorage.js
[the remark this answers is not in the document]
dariadomagala: Yet we translate things like `ReplicaSet` or `DaemonSets` which are also technical terms and seem to be 'less translatable'. We should probably discuss it and stick to one approach - either translate everything if it's possible or don't translate any technical term.

core/src/luigi-config/cluster-management/clusters-storage.js
dariadomagala: Ah, you're right. I didn't notice it's a log...

core-ui/src/components/Clusters/views/AddCluster/ChooseStorage.js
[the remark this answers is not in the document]
Wawrzyn321: So, given that TWs are to review it again anyway, I'd leave it as it is.

dariadomagala (dismissed):

core-ui/src/components/Clusters/views/AddCluster/ChooseStorage.js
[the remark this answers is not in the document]
NHingerl: In my other projects, even if, for example, DE uses the same English term as EN (for example, "Change Management"), it's still usually put into the translation file and not hardcoded. Translators can enter an English loan word into the German file if that's what's agreed upon.

We can't know yet whether it's going to be translated in any other language (maybe DE uses the English loan word, but Chinese won't), so it's good to have everything in the translation database without prior selection.

core-ui/src/components/Clusters/views/ClusterStorageType.js
NHingerl: +1

core-ui/src/components/Clusters/views/AddCluster/ChooseStorage.js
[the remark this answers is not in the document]
Wawrzyn321: I'll add rest of the translations then.

NHingerl (approved):

dariadomagala (approved):

PR #354, opened 2021-08-24
Add advances secret creation modal

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**


Wawrzyn321 (changes requested):
As we're working on `CreateModal`, would you kindly add a form revalidation when a preset is chosen? It won't do much on Secrets, but if we choose a Deployment present and it's valid already (like `Echo server`), then user may get confused why they can't submit form.

Wawrzyn321 (approved):

alexandra-simeonova (approved):

PR #370, opened 2021-08-31
Function variables: handle "all keys" in CM/Secret

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core-ui/src/components/Lambdas/LambdaDetails/Tabs/Code/LambdaVariables/VariableInputs/ResourceVariableInput.js
```diff
@@ -240,6 +260,11 @@ export default function ResourceVariableInput({
               }
             />
           </FormItem>
+          {takeAll && (
+            <MessageStrip className="fd-margin-top--sm" type="information">
+              {t('functions.variable.form.name-will-become-prefix')}
```
akucharska, line 265: Maybe we can add some information that if ALL KEYS is selected multiple variables will be created (one for every existing secret key)?

akucharska (commented):
I like it a lot. The only thing I would suggest to improve is informing user that not one env with multiple values will be created (when 'all keys' is selected) but many envs with one value

core-ui/src/components/Lambdas/LambdaDetails/Tabs/Code/LambdaVariables/VariableInputs/ResourceVariableInput.js
Wawrzyn321: Not a bad idea! I'll add it to this string.

akucharska (approved):

klaudiagrz (approved):

akucharska (approved):

PR #379, opened 2021-09-02
Create gateway

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core/src/i18n/en.yaml
```diff
@@ -556,6 +556,41 @@ functions:
...
 gateways:
+  create-modal:
+    title: Create Gateway
+    advanced:
+      add-server: Add server
+      delete-server: Delete server
+      enable-tls: Enable TLS
+      hosts: Hosts
+      placeholders:
+        hosts: 'Enter the hosts #e.g. *.api.mydomain.com'
```
Wawrzyn321, line 567: Should the '#' be there?

Wawrzyn321 (commented):
When I select `TLS Mode` to "simple" and change the `Credential Name` input, `TLS Mode` value switches back to "Select TLS Mode". And vice versa, when I choose mode credentials disappear.

Wawrzyn321:
Would it be possible to force the user to fill TLS form when HTTPS port is used? Now when I choose HTTPS and ignore TLS the form will be valid, but the request will fail.

Wawrzyn321:
Changes from `Hosts` field (simple form) or from YAML aren't propagated to `Hosts` in advanced form.

alexandra-simeonova (dismissed):

Sawthis (approved):

alexandra-simeonova (approved):

PR #381, opened 2021-09-02
API Rules meet gateways

Wawrzyn321:
**Description**

- One thing about the layout - I've changed it by the Valentin's suggestions, we're going to rewrite it to new modal anyway.
- Second thing about the "no modal edit" - firstly, it will be rewritten, so I didn't bother with removing it from code (should I?). And why was it removed? If there is a gateway with host 1: `*.example.com` and host 2: `test.example.com`, then we wouldn't be able to differentiate between apirule to `test.example.com` - was it plainly host 2, or host 1 with `test` subdomain?

**Related issue(s)**


valentinvieriu (approved):

qbalukom (approved):

tests/support/views/api-rules-view.js
```diff
@@ -14,7 +14,7 @@ Cypress.Commands.add('createApiRule', (ApiRuleName, ApiRuleHost) => {
 
   cy.getModalBody().within($modal => {
     cy.get('[placeholder="API Rule name"]').type(ApiRuleName);
-    cy.get('[placeholder="Enter the hostname"]').type(ApiRuleHost); //the host is ocupied by another virtualservice
+    cy.get('[placeholder^="Subdomain will replace the"]').type(ApiRuleHost); //the host is ocupied by another virtualservice
```
valentinvieriu, line 17: Any reason we don't use the translations here?

valentinvieriu (changes requested):

valentinvieriu (dismissed):

klaudiagrz (approved):

Sawthis (approved):

PR #395, opened 2021-09-08
Refactor CreateModal

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core-ui/src/components/Predefined/Create/ResourceForm/ResourceForm.js
```diff
@@ -0,0 +1,126 @@
+import React from 'react';
```
qbalukom, line 1: As I've mentioned before - I don't think this file should be in `Predefined/Create` - even though it's probably going to be used only by components in here.
I think it would make more sense to leave this directory for actual Resource definitions, and move the common library to `shared`

qbalukom:
As I've mentioned before I don't think `ResourceForm` belongs in `Predefined/Create` - that directory should be reserved for actual form definitions, not for shared components used by them. That's what the `shared` directory is for.

core-ui/src/shared/components/ResourceRef/ExternalResourceRef.js
```diff
@@ -1,53 +1,39 @@
...
+    onChange(e, {
+      name: selected.resource.metadata.name,
+      namespace: selected.resource.metadata.namespace,
+    });
   };
 
   return (
     <div className="external-resource-ref">
-      <ComboboxInput
+      <Dropdown
```
qbalukom, line 28: has this change been consulted from the UX perspective?
as right now we end up with a VERY long unsearchable list of secrets. and no way to enter one manually.

Wawrzyn321: I'll move the `ResourceForm` to `shared` 👌 

About the `ExternalResourceRef`, this change came from functional review with Valentin, and the problem was there was no label for the namespace, but you're probably right, I'll roll it back.

core-ui/src/shared/ResourceForm/ResourceForm.js
```diff
@@ -0,0 +1,126 @@
...
+      });
+    });
+
+  let editor = <Editor resource={resource} setResource={setResource} />;
+  editor = renderEditor
+    ? renderEditor({ defaultEditor: editor, Editor })
+    : editor;
+
+  return (
+    <article className="resource-form">
```
qbalukom, line 105: a bit of a nitpick, but i'm not sure if `article` is an appropriate tag here, as it's semantically meant to be used for actual text articles. I think `section` would suffice.

Wawrzyn321: Well, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article gives `interactive widget` as an example, so _not only_ text articles, but well, this form is still too far from a "widget". `section` it is!

core-ui/src/shared/ResourceForm/ResourceForm.js
```diff
@@ -0,0 +1,126 @@
...
+  createFn,
+  presets,
+  onPresetSelected,
+}) {
+  const notification = useNotification();
+  const { t } = useTranslation();
+  const [mode, setMode] = React.useState(ModeSelector.MODE_SIMPLE);
+
+  if (!onCreate) {
+    onCreate = async () => {
```
qbalukom, line 39: not critical, but I'm not sure about including this function in here - I think it might be a better idea to separate it to a completely standalone `resourceCreate` so it's more reusable outside of this form (for example in a wizard).

And perhaps even generalize it a bit more (for example add an option whether to redirect or not) so it can be used in Deployments as well - as two subsequent calls to it (one for deployments, one for services).

Might not be worth it, though. So it's not a requirement :)

core-ui/src/components/Predefined/Create/Deployments/Containers.js
```diff
@@ -0,0 +1,215 @@
...
+
+  if (!containers.length) {
+    return (
+      <MessageStrip type="warning">
+        {t('deployments.create-modal.advanced.one-container-required')}
+      </MessageStrip>
+    );
+  }
+
+  if (containers.length === 1) {
```
qbalukom, line 183: I thought the idea to not show collapsible containers was when there's no ability to add extra containers (and only the first one is displayed), not when there is only one on the list?

qbalukom (changes requested):

core-ui/src/shared/ResourceForm/ResourceForm.js
Wawrzyn321: That's a good point - I've created a `useResourceCreate` hook with option to override the behaviour after successful request.

And used it for Deployments, too.

It was worth it.

core-ui/src/components/Predefined/Create/Deployments/Deployments.create.js
```diff
@@ -1,6 +1,251 @@
...
+      checked={createService}
+      onChange={(_, checked) => setCreateService(checked)}
+      dir="rtl"
+    >
+      {t('deployments.create-modal.advanced.expose-service')}
+    </Checkbox>
+  );
+
+  const renderEditor = ({ defaultEditor, Editor }) => (
+    <div className="double-editor">
```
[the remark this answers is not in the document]
Wawrzyn321: I've tried to do this before, but if I want the Monaco Editor to fill the vertical space (with `height: 100%`), it grows infinitely, that's why I've set the height to a constant value in double editor.

If you have a better solution, I'm dying to see it.

core-ui/src/components/Predefined/Create/DNSProviders/DNSProviders.create.js
```diff
@@ -1,7 +1,107 @@
...
+        advanced
+        propertyPath="$.metadata.labels"
+        label={t('common.headers.labels')}
+      />
+      <ResourceForm.KeyValueField
+        advanced
+        propertyPath="$.metadata.annotations"
+        label={t('common.headers.annotations')}
+      />
+      <ResourceForm.FormField
```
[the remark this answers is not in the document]
Wawrzyn321: I've moved Labels & Annotations up, so that free fields are now next to each other, and the collapsibles follow.

qbalukom (approved):

PR #404, opened 2021-09-14
Add filter for local ip addresses

Wawrzyn321:
**Description**

- Originally the list of local/private addresses was [here](https://datatracker.ietf.org/doc/html/rfc1918#section-3), but I decided to use regex from [this answer](https://stackoverflow.com/questions/2814002/private-ip-address-identifier-in-regular-expression/62925185#62925185), removing 127.0.0.1 range to keep k3d working. That SO thread is nice, 'cause it takes also IPv6 into consideration.
- I'm not sure if we should keep or remove those `link-local addresses since 2005` - yup, we keep them in regex.
- Also handle addresses ending with `.cluster.local`.

**Related issue(s)**


backend/common.js
```diff
@@ -52,6 +54,14 @@ export const handleRequest = async (req, res) => {
     return;
   }
 
+  try {
+    filters.forEach(filter => filter(req, headersData));
+  } catch (e) {
+    req.log.error('Filters rejected the request:', e.message);
```
dariadomagala, line 60: It displays only `Filters rejected the request:`. Please use the 
```
`Filters rejected the request: ${e.message}`
```
format

dariadomagala (approved):

PR #405, opened 2021-09-14
When api rule has simple host get data from gateway

akucharska:
**Description**

- When api rule has simple host get data from gateway

**Related issue(s)**

### DIFF
--- a/core-ui/src/components/ApiRules/ApiRulesList/components.js
+++ b/core-ui/src/components/ApiRules/ApiRulesList/components.js
@@ -24,8 +24,7 @@ function getGatewayHost(gateway) {
     server => server.port.protocol === 'HTTPS',
   );
   if (!properServer?.length || !properServer[0].hosts?.length) return null;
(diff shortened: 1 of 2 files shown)


core-ui/src/components/ApiRules/ApiRulesList/components.js
```diff
@@ -19,6 +19,15 @@ function navigateToService(apiRule) {
     .navigate(`services/details/${apiRule.spec.service.name}`);
 }
 
+function getGatewayHost(gateway) {
+  const properServer = gateway.spec.servers.filter(
+    server => server.port.protocol === 'HTTPS',
+  );
+  if (!properServer?.length || !properServer[0].hosts?.length) return null;
+  const properHost = properServer[0].hosts[0].replace('*.', '');
+  return properHost;
```
dariadomagala, line 28: you can move return one line above
`return properServer[0].hosts[0].replace('*.', '')`

dariadomagala (approved):

PR #407, opened 2021-09-15
Params a as Feature

Wawrzyn321:


dariadomagala (commented):
~Something is not right, it ignores all init params, even if INIT_PARAMS is set to true~
My bad, it's too early for reviews :)

dariadomagala (approved):

PR #412, opened 2021-09-20
Gateway Create 2, but also deployment fixes

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**


dariadomagala (commented):
why 'server' part is not expandable in the basic form?

Wawrzyn321:
AFAIRemember that was one of the UX agreements. To be honest, I had no strong opinion here, we can also make the "Server" section expandable and then make the subsections inside non-expandable.

dariadomagala (dismissed):

core/src/i18n/en.yaml
```diff
@@ -716,12 +716,25 @@ functions:
       text: Warning
       variable-can-override-sbu: This variable can override or be overridden by a variable injected by one of the created ServiceBindings.
 gateways:
+  name_singular: Gateway
+  description: Gateway describes a load balancer operating at the edge of the mesh receiving incoming or outgoing HTTP/TCP connections.
   create-modal:
+    at-least-one-server-required: Gateway must have at least one server.
+    tooltips:
+      ca-certificates: The path to a file containing certificate authority certificates to use in verifying a presented client side certificate.
+      selector: One or more labels that indicate a set of pods on which the gateway configuration should be applied. By default workloads are searched across all namespaces based on label selectors.
```
NHingerl, line 725:
```suggestion
      selector: One or more labels that indicate a set of pods on which the gateway configuration is applied. By default, workloads are searched across all namespaces based on label selectors.
```

core/src/i18n/en.yaml
```diff
@@ -716,12 +716,25 @@ functions:
       text: Warning
       variable-can-override-sbu: This variable can override or be overridden by a variable injected by one of the created ServiceBindings.
 gateways:
+  name_singular: Gateway
+  description: Gateway describes a load balancer operating at the edge of the mesh receiving incoming or outgoing HTTP/TCP connections.
   create-modal:
+    at-least-one-server-required: Gateway must have at least one server.
+    tooltips:
+      ca-certificates: The path to a file containing certificate authority certificates to use in verifying a presented client side certificate.
```
NHingerl, line 724:
```suggestion
      ca-certificates: The path to a file containing certificate authority (CA) certificates, which verify a presented client-side certificate.
```

I'm adding the `(CA)` just to break up the 3 words, I had to read this 3 times till I was sure it's not just a duplication of words :D

core/src/i18n/en.yaml
```diff
@@ -716,12 +716,25 @@ functions:
...
+      selector: One or more labels that indicate a set of pods on which the gateway configuration should be applied. By default workloads are searched across all namespaces based on label selectors.
+      protocol: The protocol exposed on the port. TLS implies the connection will be routed based on the SNI header to the destination without terminating the TLS connection.
```
NHingerl, line 726:
```suggestion
      protocol: The protocol exposed on the port. **TLS** implies that the connection is routed based on the SNI header to the destination, without terminating the TLS connection.
```

I assume TLS is an option in a dropdown list/radio box or something similar?

...Can we do bold formatting in yaml?

core/src/i18n/en.yaml
```diff
@@ -716,12 +716,25 @@ functions:
...
+      target-port: The port number on the endpoint where the traffic will be received.
```
NHingerl, line 727:
```suggestion
      target-port: The port number on the endpoint where the traffic is received.
```

core/src/i18n/en.yaml
```diff
@@ -716,12 +716,25 @@ functions:
...
+      private-key: The path to the file holding the server’s private key.
+      server-certificate: The path to the file holding the server-side TLS certificate to use.
```
NHingerl, line 729:
```suggestion
      server-certificate: The path to the file holding the server-side TLS certificate you want to use.
```

core/src/i18n/en.yaml
```diff
@@ -716,12 +716,25 @@ functions:
...
+      credential-name: 'The name of the Secret that holds the TLS certs including the CA certificates. The secret (of type generic) should contain the following keys and values: key: <privateKey> and cert: <serverCert>. For mutual TLS, cacert: <CACertificate> can be provided in the same secret or a separate secret named <secret>-cacert. Secret of type tls for server certificates along with ca.crt key for CA certificates is also supported.'
```
NHingerl, line 730:
```suggestion
      credential-name: 'The name of the Secret that holds the TLS certs including the CA certificates. The secret (of type generic) must contain the following keys and values: key: <privateKey> and cert: <serverCert>. For mutual TLS (mTLS), a CA certificate (cacert): <CACertificate> can be provided in the same secret or in a separate secret named <secret>-cacert. Secret of type tls for server certificates along with ca.crt key for CA certificates is also supported.'
```
Why does this string start and end with ` ` `, but none of the others?

I assume it's a hard requirement (=> must...), not just a recommendation (should...)? If it's a recommendation, I'd suggest to write "We recommend that the secret (...) contains (...)" instead of "should".

Still trying to understand which kinds of credential name / certificate types are available.

core/src/i18n/en.yaml
```diff
@@ -716,12 +716,25 @@ functions:
...
+      hosts: At least one host exposed by the gateway, in FQDN format with optional wildcard character in the left-most component.
     advanced:
       add-server: Add server
       delete-server: Delete server
       enable-tls: Enable TLS
       hosts: Hosts
+      mode-tooltip: Indicates whether connections to the port should be secured using TLS.
```
NHingerl, line 737:
```suggestion
      mode-tooltip: Indicates whether connections to the port are secured using TLS.
```
as usual, "should" is problematic :D

core/src/i18n/en.yaml
```diff
@@ -742,8 +757,16 @@ gateways:
...
+        ca-certificates: CA Certificate
+        http-redirect: HTTP Redirect
+        disabled-for-non-https: TLS settings are not available for a non-HTTPS protocol.
+        http-redirect-description: Enable 301 redirect for all HTTP connections.
+        private-key: Private Key
+        server-certificate: Server Certificate
+        max-protocol-version: Max Protocol Version
+        min-protocol-version: Min Protocol Version
+        messages:
+          invalid-tls-warning: TLS Server of mode SIMPLE or MUTUAL require either credential name or private key and server certificate pair.
```
NHingerl, line 769:
```suggestion
          invalid-tls-warning: TLS Server of mode SIMPLE or MUTUAL need either credential name or private key and server certificate pair.
```
Ambiguous, this could mean the following
- either a "credential name" or the "private key + server certificate" pair
- either the "credential name + server certificate pair" or the "private key + server certificate" pair
- several other combinations of the words :D 
 
 We need to add commas etc to clarify it.

NHingerl (changes requested):

core/src/i18n/en.yaml
Wawrzyn321: Yup, it'a a list option... and unfortunately, we don't support formatting in translations - this is in the tooltip anyway.

core/src/i18n/en.yaml
Wawrzyn321: This is because it contains ':' character - YAML parser gets confused, so we wrap entire string with `'` to escape the ':'.

Wawrzyn321: This is a requirement.
About the change
```
For mutual TLS, cacert: <CACertificate> 
->
For mutual TLS (mTLS), a CA certificate (cacert): <CACertificate> 
```
We cannot write it like this, because it continues the pattern from previous sentence - they enumerated key and values, so keys: (key, cert, cacert) and values: (<privateKey>, <serverCert> and <CACertificate>).

Damn, they did made their docs confusing...
image

core/src/i18n/en.yaml
Wawrzyn321: Done in `Update translations`.

Wawrzyn321: (Hopefully)

core/src/i18n/en.yaml
NHingerl: :D It surely made total sense to everyone who is already familiar with that stuff!

NHingerl: > This is a requirement.
> About the change
> 
> ```
> For mutual TLS, cacert: <CACertificate> 
> ->
> For mutual TLS (mTLS), a CA certificate (cacert): <CACertificate> 
> ```
> 
> We cannot write it like this, because it continues the pattern from previous sentence - they enumerated key and values, so keys: (key, cert, cacert) and values: (, and ).
> 
> Damn, they did made their docs confusing...
> image

arrgh I'm too confused to improve it though :(

core/src/i18n/en.yaml
NHingerl: I think so! :)

core/src/i18n/en.yaml
Wawrzyn321: Oh, great, only now I noticed that I can't put words between '<' and '>', because they disappear. Oh well...

Wawrzyn321: Maybe just reword it to make it simpler?
```
The name of the Secret. Required properties are key and cert, in case of mutual TLS (mTLS) caceert property is also required. Secret of type tls for server certificates along with ca.crt key for CA certificates is also supported.'
```
I'd go for less information - just give clues, and if users need more info they can check out the other sources, not only istio docs.

NHingerl: I agree. They just need a starting point to know what to look for.

NHingerl (dismissed):

core-ui/src/shared/ResourceForm/useCreateResource.js
```diff
@@ -28,6 +28,7 @@ export function useCreateResource(
   };
 
   return async () => {
+    console.log('yeah');
```
dariadomagala, line 31: nope :)

dariadomagala (changes requested):

dariadomagala (dismissed):

NHingerl (dismissed):

dariadomagala (approved):

NHingerl (approved):

PR #428, opened 2021-09-28
Issuers

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core/src/i18n/en.yaml
```diff
@@ -309,10 +309,13 @@ common:
...
     loading-name: Loading name...
-    no-instances-found: No {{value}} found.
-    no-instances-found-namespace: No {{value}} found in this Namespace.
+    no-instances-found: No matching {{value}} found.
+    no-instances-found-namespace: No matching {{value}} found in this Namespace.
     parse-error: Parse error, previous valid input will be used.
     type-to-select: Start typing to select {{value}} from the list.
+  placeholders:
+    secret-ref-name: Select name
+    secret-ref-namespace: Select namespace
```
NHingerl, line 318:
```suggestion
    secret-ref-namespace: Select Namespace
```

core/src/i18n/en.yaml
```diff
@@ -324,6 +327,8 @@ common:
     read-file: Read file content as a single value, with file name as a key.
     save: Save
     search: Search
+    secret-ref-name: Name is unique within a namespace to reference a secret resource.
```
NHingerl, line 330:
```suggestion
    secret-ref-name: Unique name within a Namespace to refer to a Secret resource.
```
Is it Secret as in Kubernet Secret (then we need to capitalize), or another kind of secret?

core/src/i18n/en.yaml
```diff
@@ -469,6 +474,15 @@ deployments:
   title: Deployments
 discovery-and-network:
   title: Discovery and Network
+domains:
+  exclude:
+    label: Exclude Domains
+    placeholder: Domain that should be ignored
+    tooltip: Exclude are domain names for which certificate requests are forbidden (including any subdomains) for example, *.api.mydomain.com.
```
NHingerl, line 481:
```suggestion
    tooltip: The domains and subdomains for which certificate requests are forbidden, for example, *.api.mydomain.com.
```

core/src/i18n/en.yaml
```diff
@@ -469,6 +474,15 @@ deployments:
   title: Deployments
 discovery-and-network:
   title: Discovery and Network
+domains:
+  exclude:
+    label: Exclude Domains
+    placeholder: Domain that should be ignored
```
NHingerl, line 480: "should" - is this a recommendation? Or is it certainly going to be ignored if entered here?
Based on the following line, how about

 ```suggestion
    placeholder: Domain that is skipped for certificate requests
```

File: core/src/i18n/en.yaml [hunk 5]
```diff
@@ -469,6 +474,15 @@ deployments:
...
+    tooltip: Exclude are domain names for which certificate requests are forbidden (including any subdomains) for example, *.api.mydomain.com.
+  include:
+    label: Include Domains
+    placeholder: Domain that should be observed
```
NHingerl, line 484: again not sure about the "should"; I assume an entry here is not just a recommendation?

```suggestion
    placeholder: Domain that is observed
```

core/src/i18n/en.yaml
```diff
@@ -469,6 +474,15 @@ deployments:
...
+    tooltip: Include are domain names for which certificate requests are allowed (including any subdomains) for example, *.api.mydomain.com.
```
NHingerl, line 485:
```suggestion
    tooltip: The domains (including subdomains) for which certificate requests are allowed for example, *.api.mydomain.com.
```

core/src/i18n/en.yaml
```diff
@@ -806,6 +837,16 @@ issuers:
   skip-dns: Skip DNS challenge validation
   state: Status
   title: Issuers
+  tooltips:
+    email: Email is the email address to use for user registration.
```
NHingerl, line 841:
```suggestion
    email: The email address to use for user registration.
```

core/src/i18n/en.yaml
```diff
@@ -806,6 +837,16 @@ issuers:
...
+    external-account-binding: ACMEExternalAccountBinding is a reference to a CA external account of the ACME server.
+    issuer-type: Certification authority used
+    key-secret-ref: keySecretRef is the secret ref to the Secret which holds the symmetric MAC key of the External Account Binding with data key 'hmacKey'. The secret key stored in the Secret **must** be un-padded, base64 URL encoded data.
```
NHingerl, line 844:
```suggestion
    key-secret-ref: keySecretRef is the secret ref to the Secret that holds the symmetric MAC key of the External Account Binding with data key 'hmacKey'. The secret key stored in the Secret **must** be un-padded, base64 URL encoded data.
```

core/src/i18n/en.yaml
akucharska: Yes it is

core/src/i18n/en.yaml
akucharska: Not sure if `is skipped` is a good explanation here. This placeholder should be a short version of this:
```The domains (including subdomains) for which certificate requests are forbidden, for example, *.api.mydomain.com.```

Comment (NHingerl) [reply core/src/i18n/en.yaml:484]: That's what I was confused about - at one place we write "ignored", somewhere else "excluded", and then also "forbidden".  They're all related but don't mean the same thing. 
My assumption was that it's basically an allowlist of domains that are always "good" and certificate never must be checked. As opposed to the Include domains that always must be checked. Is that correct?

File: core/src/i18n/en.yaml [hunk 9]
```diff
@@ -843,7 +845,9 @@ issuers:
   title: Issuers
   tooltips:
     email: The email address to use for user registration.
+    excluded-domains: The domains (including subdomains) for which certificate requests are forbidden, $t(domains.example).
```
NHingerl, line 848:
```suggestion
    excluded-domains: The domains (including subdomains) for which certificate requests are forbidden, for example, $t(domains.example).
```

NHingerl (dismissed):

core-ui/src/components/Predefined/Create/Issuers/Issuers.create.js
```diff
@@ -1,56 +1,271 @@
...
+        customSetValue={name => {
+          jp.value(issuer, '$.metadata.name', name);
+          jp.value(issuer, "$.metadata.labels['app.kubernetes.io/name']", name);
+          setIssuer({ ...issuer });
+        }}
+      />
+      <ResourceForm.FormField
+        label={t('issuers.type')}
+        key={t('issuers.type')}
+        required
```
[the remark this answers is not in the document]
dariadomagala: Also, check this scenario:
* I choose ACME from this dropdown
* other required fields show up
* the submit button is still active, even tho required fields are empty
* I submit it with empty fields
* Modal is closed, no error, no message, nothing happened

core-ui/src/components/Predefined/Create/Issuers/templates.js
```diff
@@ -0,0 +1,36 @@
...
+  };
+}
+export function createACMETypeTemplate() {
+  return {
+    acme: {
+      server: '',
+      email: '',
+      autoRegistration: true,
+    },
+    // privateKeySecretRef:{
```
dariadomagala, line 31: why is it commented out?

core-ui/src/components/Predefined/Create/Issuers/Issuers.create.js
```diff
@@ -1,56 +1,271 @@
...
+          title={t('issuers.private-key')}
+          tooltipContent={t('issuers.tooltips.secret-ref-acme')}
+          onChange={value => {
+            jp.value(issuer, '$.spec.acme.privateKeySecretRef', value);
+            setIssuer({ ...issuer });
+          }}
+          actions={
+            <Checkbox
+              compact
+              checked={issuer.spec.acme?.autoRegistration}
```
dariadomagala, line 131: first click on the checkbox doesn't change it

dariadomagala (changes requested):
the `components.dropdown.empty-list` shows in the Secret name dropdown at the beginning.

Wawrzyn321 (approved):

PR #444, opened 2021-10-01
Docker secrete

mrCherry97:
**Description**
If the docker hub repository is private user need credentials from secrets.

- Added image pull secret input
- Added ImagePullSecret field to YAML generate
- Made secret fields for simple and advance

### DIFF
--- a/core-ui/src/components/Predefined/Create/Deployments/templates.js
+++ b/core-ui/src/components/Predefined/Create/Deployments/templates.js
@@ -44,6 +44,7 @@ export function createDeploymentTemplate(namespaceId) {
           },
         },
         spec: {
+          imagePullSecrets: [],
           containers: [createContainerTemplate()],
         },
       },
--- a/core-ui/src/shared/components/K8sResourceSelect.js
+++ b/core-ui/src/shared/components/K8sResourceSelect.js
@@ -50,6 +50,7 @@ export function K8sResourceSelect({
   required,
   isNamespaced = true,
   allowTyping = true,
+  ...props
 }) {
   const { t } = useTranslation();
 
@@ -106,6 +107,9 @@ export function K8sResourceSelect({
     </div>
   ) : (
     <Select
+      placeholder={t('common.messages.click-to-select', {
+        value: resourceType,
+      })}
       value={value}
       setValue={onSelect}
       defaultKey={value}
--- a/core-ui/src/shared/components/K8sResourceSelect.js
+++ b/core-ui/src/shared/components/K8sResourceSelect.js
@@ -5,6 +5,7 @@ import { useGetList, k8sNamePattern } from 'react-shared';
 import { useTranslation } from 'react-i18next';
 import pluralize from 'pluralize';
 import './K8sResourceSelect.scss';
+import { Select } from '../ResourceForm/components/FormComponents';
 
 const commonPropTypes = {
   onSelect: PropTypes.func.isRequired,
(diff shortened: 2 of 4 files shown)


mrCherry97:
If someone want test how credential for docker works you have to made that command:
```
kubectl create secret docker-registry --dry-run=true docker-regcred \
--docker-server=https://index.docker.io/v1/ \
--docker-username=xxx \
--docker-password=xxx \
--docker-email=yourmail@yourdomain.com \
--namespace=xxx \
-o yaml > docker-secret.yaml
```
ALERT: Credentials can be encoded in busola :P

core-ui/src/components/Predefined/Create/Deployments/templates.js
```diff
@@ -44,6 +44,7 @@ export function createDeploymentTemplate(namespaceId) {
           },
         },
         spec: {
+          imagePullSecrets: [],
```
dariadomagala, line 47: If it's an array then maybe we should add an option (at least in an advanced form) to add multiple secrets?

dariadomagala (commented):
Also, please change the PR name to something more descriptive, 'docker secrete' tells me nothing 😉

core-ui/src/components/Predefined/Create/Deployments/templates.js
Lyczeq: We talked about it during the daily and Valentin said that's enough for now.

core-ui/src/shared/components/K8sResourceSelect.js
```diff
@@ -88,7 +85,7 @@ export function K8sResourceSelect({
     } else return undefined;
   };
 
-  return allowTyping ? (
+  return (
     <div className="combobox--full-width">
       <ComboboxInput
```
dariadomagala: Please use the combobox with the [showAllEntries](https://github.com/SAP/fundamental-react/blob/main/src/ComboboxInput/ComboboxInput.js#L654) and [searchFullString](https://github.com/SAP/fundamental-react/blob/main/src/ComboboxInput/ComboboxInput.js#L643) flags enabled. (The documentation is not yet on the official site, since we don't have an official release yet, so please check the comments in the code, linked above)

I'd also suggest using the `selectionType='auto-inline'`, but that's up to you to decide.

mrCherry97: With `showAllEntries` flag it works not so well

image

without this flag searching matching entries works better

image

mrCherry97: We want to see all entries and matching entries on top or only matching entries?

dariadomagala: we should see all entries with the matching ones on top. the matching string should also be **bold**, I don't know why it's not, we could create a bug for that.

dariadomagala (dismissed):

core/src/i18n/en.yaml
```diff
@@ -473,6 +474,8 @@ deployments:
     simple:
       docker-image: Docker Image
       docker-image-placeholder: Enter the Docker image tag, for example, bitnami/nginx.
+      image-pull-secret: Image Pull Secret
+      image-pull-secret-tooltip: The secret for authorization to the private docker repository.
```
NHingerl, line 478:
```suggestion
      image-pull-secret-tooltip: The secret for authorization to the private Docker repository.
```

dariadomagala (approved):

PR #455, opened 2021-10-08
Dns entry create

### LINKED ISSUE #447: Refactor DNS Entry to match the general Create Resource Views
**Description**

We need to be consistent with the rest of the Create Resource Views and to use the same fields and patterns so that our users don't feel confused.

**Reasons**

We want our UI to guide the user and make him feel confident when he creates a new DNS Entry

**Acceptance Criteria**

- [ ] Arrange all the forms as we have in the Create Gateway View
- [ ] Introduce Tooltips to all elements explaining what are those fields about. The initial explanations come from the CRD Schema or Official Documentation
- [ ] Refactor the Secret Selection so that is a 2 step process. Select Namespace ( with option to search ) and Select Secret ( with option to search ).Make sure we don't show system namespaces if `Show hidden Namespaces` is toggles off
- [ ] All the Domains fields that allows more than one domain should use the new ui  ( this is targets )
- [ ] simplify the basic view  (
- [ ] Check if we are on gardener (namespaces/kube-system/configmaps/details/shoot-info) and add annotation `dns.gardener.cloud/class: garden`
- [ ] Text Name should use the same UI Targets ( multi items )

dariadomagala-sap:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Resolves #447

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -151,26 +151,19 @@ catalog:
   subtitle: Enrich your experience with additional services that are installed outside the cluster.
   title: Service Catalog
 certificates:
-  name_singular: Certificate
   common-name: Common Name
-  dns-names: DNS Names
-  expiration-date: Expiration Date
-  issuer: Issuer
-  refs: References
-  state: Status
-  title: Certificates
-  csr: CSR
-  with-csr: Use a CSR
-  renew: Trigger a renewal
-  existing-secret: Use the existing Secret
-  secret-name: Secret name
-  secret-ref: Secret to use
   create:
     messages:
       failure: Failed to create the certificate.
       success: Certificate created.
+  csr: CSR
+  dns-names: DNS Names
+  existing-secret: Use the existing Secret
+  expiration-date: Expiration Date
+  issuer: Issuer
   messages:
     decode-error: 'Failed to decode the CSR: {{message}}.'
+  name_singular: Certificate
   placeholders:
     common-name: Certificate CN (max 64 characters)
     csr: Plain text Certificate Signing Request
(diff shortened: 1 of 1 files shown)


core/src/i18n/en.yaml
```diff
@@ -514,12 +514,23 @@ dnsentries:
...
     ttl: Enter the time to live.
-    dns-name: Enter DNS Name
+    dns-name: Enter the DNS Name.
+    target-a: Enter the A record Target.
+    target-cname: Enter the CNAME record Target.
+  tooltips:
+    use-cname: Switch to CNAME record.
+    use-a: Switch to A record.
+    target: Target must be an A record, or a CNAME record.
+    targets: Targets records, either target ot text are required.
```
akucharska, line 530: Fix a typo: `either target or text`

akucharska (commented):
Good job. Just one small suggestion to fix:

On a smaller screen in the target section, the toggle is overlapping with the 'Use CNAME' label.

core/src/i18n/en.yaml
```diff
@@ -514,12 +514,23 @@ dnsentries:
...
     targets: Targets
+    use-cname: Use CNAME
+  name_singular: DNS Entry
   placeholders:
-    targets: Target records (CNAME or A records), one per line
-    text: Text records, one per line
     ttl: Enter the time to live.
-    dns-name: Enter DNS Name
+    dns-name: Enter the DNS Name.
+    target-a: Enter the A record Target.
```
NHingerl, line 524: why is target capitalized?

```suggestion
    target-a: Enter the A record target.
```

core/src/i18n/en.yaml
```diff
@@ -514,12 +514,23 @@ dnsentries:
...
+    target-cname: Enter the CNAME record Target.
```
NHingerl, line 525:
```suggestion
    target-cname: Enter the CNAME record target.
```

core/src/i18n/en.yaml
```diff
@@ -514,12 +514,23 @@ dnsentries:
...
+  tooltips:
+    use-cname: Switch to CNAME record.
+    use-a: Switch to A record.
+    target: Target must be an A record, or a CNAME record.
+    targets: Targets records, either target or text are required.
```
NHingerl, line 530:
```suggestion
    targets: Targets records, either target or text, are required.
```

Is it really "target**s** records", or rather "target records" ?

core/src/i18n/en.yaml
```diff
@@ -514,12 +514,23 @@ dnsentries:
...
+    text: Text record must be a string.
+    texts: Text records, either target or text are required.
```
NHingerl, line 532:
```suggestion
    texts: Text records, either target or text, are required.
```

dariadomagala: I think the comma changes the meaning of this sentence a little, suggesting that text records are required, and they can be either target or text type (at least that's how I would understand it). 
What I meant here is that this is a text record input field, but in general, there are 2 input fields, one with target records, one with text records, and at least one of those fields is required. See the screenshot below.
Screen Shot 2021-10-14 at 09 46 05

NHingerl: How about 

```suggestion
    texts: Either target or text are required.
```

or 

```suggestion
    texts: You must provide either target or text.
```

NHingerl (approved):

PR #458, opened 2021-10-11
Role details

Wawrzyn321:

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core-ui/src/components/Predefined/Details/Roles/Rules.js
+++ b/core-ui/src/components/Predefined/Details/Roles/Rules.js
@@ -6,7 +6,7 @@ import { Icon } from 'fundamental-react';
 function VerbStatus({ rule, verb }) {
   const hasVerb = rule.verbs?.includes(verb) || rule.verbs?.includes('*');
 
-  return hasVerb ? <Icon glyph="accept" ariaHidden /> : null;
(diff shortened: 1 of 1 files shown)


core-ui/src/components/Predefined/Details/Roles/Rules.js
```diff
@@ -1,45 +1,76 @@
...
-import { LayoutPanel } from 'fundamental-react';
+import { GenericList, EMPTY_TEXT_PLACEHOLDER } from 'react-shared';
 import { useTranslation } from 'react-i18next';
+import { Icon } from 'fundamental-react';
 
-import './Rules.scss';
+function VerbStatus({ rule, verb }) {
+  const hasVerb = rule.verbs?.includes(verb) || rule.verbs?.includes('*');
+
+  return hasVerb ? <Icon glyph="accept" ariaHidden /> : null;
```
dariadomagala, line 9: Maybe we should add an empty placeholder as well instead of null?

Wawrzyn321: I was thinking about it, but I was a little bit concerned that it would be too cluttered. When I try it out now, it actually... looks fine.

PR #462, opened 2021-10-13
Show no permissions

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**


qbalukom (dismissed):

core/src/i18n/en.yaml
```diff
@@ -297,6 +297,7 @@ common:
     no-permissions: You don't have enough permissions to view this content.
     no-permissions-header: Not enough permissions.
     no-permissions-message: Contact your administrator to grant you access.
+    no-permissions-no-role: It looks like you have no role assigned.
```
klaudiagrz, line 300:
```suggestion
    no-permissions-no-role: You have no role assigned.
```

klaudiagrz: Your version is okay, too, but the shorter the better 🙂

PR #472, opened 2021-10-18
Role create

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**


dariadomagala (changes requested):
Switching between tabs or collapsing/expanding sections doesn't always keep the values in the inputs. 
If I'd type in an API group in simple mode, then it's always empty after switching tabs. If I'd add it in advanced mode and switch tabs, then it's empty in simple mode but filled in in advanced mode. 🤷

PR #500, opened 2021-11-02
Function resources

qbalukom:


core-ui/src/shared/components/MemoryInput.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+// import * as jp from 'jsonpath';
+// import { FormInput, FormItem } from 'fundamental-react';
```
Wawrzyn321, line 3:
```suggestion
```

core-ui/src/shared/components/MemoryInput.js
```diff
@@ -0,0 +1,45 @@
...
+
+  const numericValue = value.match(/^\d*(\.\d*)?/)[0];
+  const unit = value.replace(numericValue, '');
+  const selectedUnit = units.includes(unit) ? unit : '';
+
+  // const setValue = val => {
+  // jp.value(container, propertyPath, val);
+  // setContainer(container);
+  // };
+
```
Wawrzyn321, line 26:
```suggestion
```

Wawrzyn321 (commented):
Do you think moving `RuntimeResources` presets dropdown to `CollapsiblePanel` header is worth it? We do this on `CronJobs`
image
I know we'd need to restructure the current code a bit, so... it's rather a sugestion.

Wawrzyn321 (changes requested):
In Deployments create modal the `Service` section content is gone, bring it back please.

Wawrzyn321 (approved):

mmitoraj (approved):

PR #504, opened 2021-11-02
add workload components definitions

grego952:
**Description**

- add definitions for Deployment, StatefulSet, Job, DaemonSet, ReplicaSet, Pod
- ...
- ...

**Related issue(s)**


core/src/i18n/en.yaml
```diff
@@ -960,6 +962,7 @@ jobs:
     status: Status
     title: Conditions
     type: Type
+  description: ' a finite or batch task that runs to completion.'
```
Wawrzyn321, line 965: [](url)
```suggestion
  description: ' is a finite or batch task that runs to completion.'
```

qbalukom:
I think it might be a good idea to use the [Trans component](https://react.i18next.com/latest/trans-component) instead of `<Link>...</Link> {t('...')`

Wawrzyn321 (changes requested):
Would you kindly replace all the occurrences of `className="fd-link fd-link"` with single fd-link? I know it's not the part of the task, but it's a quick global search & replace.

core/src/i18n/en.yaml
```diff
@@ -558,6 +559,7 @@ deployments:
       image-pull-secret: Image Pull Secret
       image-pull-secret-tooltip: The secret for authorization to the private Docker repository.
     title: Create Deployment
+  description: ' is a Kubernetes object that represents a replicated application running on your cluster.'
```
majakurcius, line 562: Do you mean `application` in a general sense, or `Application`, as in our custom resource representing an external solution?

core/src/i18n/en.yaml
```diff
@@ -960,6 +962,7 @@ jobs:
     status: Status
     title: Conditions
     type: Type
+  description: ' is a finite or batch task that runs to completion.'
```
majakurcius, line 965: What do you mean by "runs to completion"?

core/src/i18n/en.yaml
```diff
@@ -485,7 +485,7 @@ cron-jobs:
       month: 1 to 12
       starting-deadline: Optional deadline in seconds for starting the job if it misses the scheduled time.
       suspend: 'If true, the controller suspends subsequent executions (default: false). It does not apply to started executions.'
-  description: CronJobs are for regular scheduled actions. A CronJob object is like one line of a crontab file, running a job periodically on a given schedule.
+  description: ' is for regular scheduled actions. A CronJob object is like one line of a crontab file, running a job periodically on a given schedule.'
```
majakurcius, line 488:
```suggestion
  description: ' is for regular scheduled actions. A CronJob object is like one line of a crontab file, running a job periodically on a given schedule.'
```
I'm not 100% sure about the explanation. It references another concept (`crontab`), which to me is not clear BTW, and thus assumes a certain level of understanding from the user, but if we expect the user to know certain things, then why do we need to explain others? 
How would you explain CronJob to someone who doesn't know a first thing about it in a conversation?

core/src/i18n/en.yaml
grego952: According to my understanding it is general.

core/src/i18n/en.yaml
grego952: It runs until it is terminated. It is taken from the K8s glossary. Maybe it will be better to write sth like: `creates one or more Pods and ensures that a specified number of them successfully terminate.`?

majakurcius: I think more specific info could be useful here, 'cause I'd generally assume that any task eventually runs to completion and any Pod eventually terminates (hopefully successfully 🤞 ), so I'd try to emphasize what's special here.

core/src/i18n/en.yaml
grego952: Maybe `allows to execute actions on a regular basis`?

majakurcius: Also, note that you should not take stuff off another website (Kubernetes) without quoting it, otherwise it might be plagiarism. 
So here, you'd want to use your own words.

Wawrzyn321 (dismissed):

Wawrzyn321 (dismissed):

Wawrzyn321 (approved):

majakurcius (approved):

PR #512, opened 2021-11-08
add Service definition

grego952:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -1214,6 +1214,7 @@ service-management:
   title: Service Management
 services:
   cluster-ip: Cluster IP
+  description: <0>Service </0> enables access to an application running on a set of Pods.
   external-ips: External IPs
   labels:
     service: Service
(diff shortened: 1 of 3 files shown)


core/src/i18n/en.yaml
```diff
@@ -1214,6 +1214,7 @@ service-management:
   title: Service Management
 services:
   cluster-ip: Cluster IP
+  description: <0>Service </0> enables access to an application running on a set of Pods.
```
alexandra-simeonova, line 1217:
```suggestion
  description: <0>Service</0> enables access to an Application running on a set of Pods.
```

See https://kyma-project.io/community/guidelines/content-guidelines/08-style-and-terminology/#terminology for Application vs. application.
Also, since the other descriptions containing `<0>` ([example](https://github.com/kyma-project/busola/blob/2a2030193f0c5516512b4d86156b56009bc1d872/core/src/i18n/en.yaml#L28)) don't leave an empty space, I think it's best to delete the extra space.

Wawrzyn321: In this case the `application` is right.

Wawrzyn321 (approved):

Wawrzyn321 (approved):

alexandra-simeonova (approved):

PR #513, opened 2021-11-09
add integration components defintions

grego952:
**Description**

- add defintion of Application and Addons Configuration
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -158,7 +158,7 @@ btp-service-bindings:
     secret: The name of the secret where the credentials are stored, defaults to the ServiceBinding name if not specified.
     secret-not-found: This Secret doesn't exist.
 catalog:
-  description: <0>Service Catalog</0> provides an easy way to access services that the Service Brokers manage and use them in an application.
+  description: <0>Service Catalog</0> provides an easy way to access services that the Service Brokers manage, and to use them in an application.
   menu-title: Catalog
   subtitle: Enrich your experience with additional services that are installed outside the cluster.
   title: Service Catalog
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -1,5 +1,6 @@
 addons:
   addons-status: Addons Status
+  description: <0>Addons Configuration</0> defines Namespace-scoped addons that provide all information required to convert Helm charts into Service Classes in Service Catalog.
   errors:
     cannot-create: Cannot create the AddonsConfiguration.
   headers:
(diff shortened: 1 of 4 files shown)


core/src/i18n/en.yaml
```diff
@@ -156,6 +158,7 @@ btp-service-bindings:
     secret: The name of the secret where the credentials are stored, defaults to the ServiceBinding name if not specified.
     secret-not-found: This Secret doesn't exist.
 catalog:
+  description: <0>Service Catalog</0> provides an easy way to access services that the Service Brokers manage and use them in an application.
```
NHingerl, line 161: ambiguous.

If it's the human who uses the services, I suggest
```suggestion
  description: <0>Service Catalog</0> provides an easy way to access services that the Service Brokers manage, and to use them in an application.
```
If it's the Service Brokers who use the services, I suggest

```suggestion
  description: <0>Service Catalog</0> provides an easy way to access services that the Service Brokers manage and use in an application.
```

core/src/i18n/en.yaml
```diff
@@ -1,5 +1,6 @@
 addons:
   addons-status: Addons Status
+  description: <0>Addons Configuration</0> defines Namespace-scoped addons that provide all information required to convert Helm charts into Service Classes in Service Catalog.
```
Wawrzyn321, line 3: What about
```suggestion
  description: <0>Addons</0> define Namespace-scoped addons that provide all information required to convert Helm charts into Service Classes in Service Catalog.
```

And the same with Cluster Addons?

NHingerl, line 3: `Namespace-scoped addons` I'm not a fan of the capitalized hyphenated adjective, but I guess character constraints made that necessary?

Wawrzyn321 (dismissed):

core/src/i18n/en.yaml
```diff
@@ -198,6 +201,7 @@ certificates:
     with-csr: Enable to use a Certificate Signing Request instead of providing domain names manually.
   with-csr: Use a CSR
 cluster-addons:
+  description: <0>Cluster Addons Configuration</0> defines cluster-wide addons that provide all information required to convert Helm charts into Service Classes in Service Catalog.
```
NHingerl, line 204: Aaaah I see, you went for consistent sentence structure here and in line 3, makes sense.

core/src/i18n/en.yaml
klaudiagrz: We also capitalize the term "Namespace" in our docs, since it's a Kubernetes term. See the [guidelines](https://github.com/kyma-project/community/blob/main/docs/guidelines/content-guidelines/02-style-and-terminology.md#capitalized-terms) for reference.

Wawrzyn321 (approved):

core/src/i18n/en.yaml
NHingerl: I know :sob: it just looks less weird if it's a noun, not part of an adjective phrase. But I'm not going to object :D

NHingerl (approved):

PR #519, opened 2021-11-10
Create replica set form

akucharska:
**Description**

- Create replica set form

**Related issue(s)**


core/src/i18n/en.yaml
```diff
@@ -1110,13 +1110,24 @@ preferences:
...
   cpu: CPU
+  create-modal:
+    labels:
+      min-ready-seconds: Minimum ready seconds
+      replicas: Replicas
+    placeholders:
+      min-ready-seconds: Minimum number of seconds
+      replicas: Replicas
+    tooltips:
+      min-ready-seconds: Minimum number of seconds for which a newly created pod should be ready without any of its container crashing, for it to be considered available.
```
grego952, line 1121:
```suggestion
      min-ready-seconds: Minimum number of seconds for which a newly created Pod should be ready without any of its container crashing, to be considered available.
```

alexandra-simeonova, line 1121:
```suggestion
      min-ready-seconds: Minimum number of seconds for which a newly created Pod should be ready without any of its containers crashing in order to be considered available.
```

alexandra-simeonova (approved):

core-ui/src/shared/components/K8sResourceSelect.js
```diff
@@ -98,6 +98,7 @@ export function K8sResourceSelect({
         })}
```
dariadomagala, line 98: I cannot add a comment there but shouldn't we switch the selectionType to 'manual' (line 93)? this one doesn't work good

dariadomagala (approved):

alexandra-simeonova (approved):

PR #538, opened 2021-11-12
Hide edit and remove buttons at the Function list

wchrapka:

- Hide edit and remove buttons at the Function list
- Add tooltip informing why


valentinvieriu:
Let's move those settings to the management-plane config

PR #539, opened 2021-11-15
Cluster config items

grego952:
**Description**

- add definitions of cluster configuration items
- add definition of Role
- ...
- ...

**Related issue(s)**


core/src/i18n/en.yaml
```diff
@@ -514,6 +516,7 @@ cron-jobs:
   schedule: Schedule
   title: Cron Jobs
 custom-resource-definitions:
+  description: <0>Custom Resource Definition</0> is a file used to define a custom resource that allows you to extend the Kubernetes API to cover use cases that are not directly covered by core Kubernetes.
```
majakurcius, line 519:
```suggestion
  description: <0>Custom Resource Definition</0> is an object used to define a custom resource that allows you to extend the Kubernetes API for use cases that are not directly covered by core Kubernetes.
```
Using the `for` you can get rid of the double `cover` and `to`.
Also, CRD is not necessarily a file, it can be defined in various ways, e.g. as a Go struct, via an HTTP call, etc.

majakurcius (approved):

Wawrzyn321 (approved):

PR #569, opened 2021-11-17
Test Jobs

### LINKED ISSUE #558: Automate tests for Jobs
**Description**

Automate tests for Jobs: 
- create view
- list view
- details view
- edit view ( edit button is added to the detail view )

Wawrzyn321:
**Description**

- Add Edit mode to Jobs, make readonly fields readonly
- Test Jobs

**Related issue(s)**

Closes #558


tests/tests/test-jobs.spec.js
```diff
@@ -0,0 +1,206 @@
...
+      showLogsSelector: `[aria-label="view-logs-for-${JOB_NAME}"]`,
+      expectedLogs: 'Busola test',
+    });
+    checkJobLogs({
+      showLogsSelector: `[aria-label="view-logs-for-${SECOND_CONTAINER_NAME}"]`,
+      expectedLogs: 'Node image test',
+    });
+
+    // sometimes ControlledBy component still renders CRD(?) path instead of our predefined path - wait until it loads
+    cy.wait(500);
```
Sawthis, line 145: Can we wait for status or something else?

Sawthis (changes requested):
Please look on comment.

Sawthis (approved):

Sawthis (approved):

core/src/i18n/en.yaml
```diff
@@ -988,6 +988,7 @@ jobs:
     at-least-one-container-required: Cron Job must have at least one container.
     container: Container
     containers: Containers
+    containers-readonly-in-edit: Containers are readonly in edit mode.
```
NHingerl, line 991:
```suggestion
    containers-readonly-in-edit: In edit mode, containers are read-only.
```

Isn't this the opposite of what "edit mode" means? Or can other resources be changed, just not containers? Why do we have to mention edit mode at all? Why not something like:

```suggestion
    containers-readonly-in-edit: You can't edit containers.
 ```
or 
```suggestion
    containers-readonly-in-edit: Containers are read-only.
```

Wawrzyn321: You can change containers, but only during resource creation. Should we go with your first comment, or try out something else, like `After a Job is created the containers are read-only.`?

NHingerl: Your suggestion explains it best! 
Adding a comma and I don't think Job needs a capital J.

```suggestion
    containers-readonly-in-edit: After a job is created, the containers are read-only.
```

Sawthis (approved):

core/src/i18n/en.yaml
```diff
@@ -906,7 +906,7 @@ generic-list:
 git-repositories:
   auth:
     basic: Basic
-    public: Public
+    public: PublicAfter a job is created, the containers are read-only.
```
NHingerl, line 909:
```suggestion
    public: After a Job is created, the containers are read-only.
```

I removed the "Public", typo?

NHingerl: Ah wait, that's the only difference between 909 and 991. Bamboozled again.

PR #578, opened 2021-11-18
Cronjobs test

### LINKED ISSUE #562: Automate CronJobs
**Description**

Automate tests for Jobs and CronJobs: 
- create view
- list view
- details view
- Edit view

mrCherry97:
**Description**

- Added test for:
   - create Cron Job,
   - display information,
   - edit details,
- Allowed edit for Cron Jobs.

**Related issue(s)**
Closes #562

### DIFF
--- a/core-ui/src/components/Predefined/Create/Jobs/CronJobs.create.js
+++ b/core-ui/src/components/Predefined/Create/Jobs/CronJobs.create.js
@@ -27,6 +27,7 @@ function CronJobsCreate({
   namespace,
   onChange,
   setCustomValid,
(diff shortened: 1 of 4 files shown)


Wawrzyn321 (changes requested):
It works fine, but I think that testing also Jobs list on CronJob details view could bring even more value.

Could you (e.g. on Edit step) change the schedule to like `every 5 seconds` and check if the new Job appears on list? Then it would be good to delete the CronJob because it would keep spamming every 5 seconds.

Wawrzyn321 (approved):

PR #615, opened 2021-11-26
Improve validation in ports and issuers requests

Lyczeq:
**Description**

- Validation for ports to not be able to use the `0` PORT
- Requests in the Issuers shouldn't be negative numbers


core-ui/src/components/Predefined/Create/Issuers/Issuers.create.js
```diff
@@ -319,6 +319,8 @@ function IssuersCreate({
         tooltipContent={t('issuers.tooltips.requests')}
         placeholder={t('issuers.placeholders.requests-per-day')}
         input={Inputs.Number}
+        validate={requests => typeof requests === 'number' && requests >= 0}
```
dariadomagala, line 322: Can we move this validation to the `Number.js` file, so it works for every such input?

dariadomagala (approved):

PR #639, opened 2021-12-01
Add word wrapping in the generic table

wchrapka:
**Description**

Add a change to wrap long phrases in table cells.

**Related issue(s)**


shared/components/GenericList/GenericList.scss
```diff
@@ -65,6 +65,7 @@
   td {
     vertical-align: middle;
     min-width: 3em;
+    word-break: break-all;
```
akucharska, line 68: Maybe break0word would be better?

akucharska (approved):

PR #641, opened 2021-12-01
Fix scrolling page when mouse is over monaco container

wchrapka:
**Description**

Fixes scrolling page when mouse is over a monaco container and removes the scrollbars from the main iframe container to disallow double scrolls.

**Related issue(s)**

### DIFF
--- a/core-ui/src/shared/components/ReadonlyEditorPanel.js
+++ b/core-ui/src/shared/components/ReadonlyEditorPanel.js
@@ -10,6 +10,9 @@ export function ReadonlyEditorPanel({ title, value, editorProps }) {
     minimap: {
       enabled: false,
     },
+    scrollbar: {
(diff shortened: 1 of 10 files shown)


core-ui/src/shared/components/ReadonlyEditorPanel.js
```diff
@@ -10,6 +10,9 @@ export function ReadonlyEditorPanel({ title, value, editorProps }) {
     minimap: {
       enabled: false,
     },
+    scrollbar: {
```
dariadomagala, line 13: We have more Monaco editors in the app (like lambda details view), we should add it everywhere.

dariadomagala (changes requested):

dariadomagala (approved):

dariadomagala (approved):

dariadomagala (approved):

PR #672, opened 2021-12-06
Add filtering and links to event messages

wchrapka:
**Description**

Adds message filtering to Cluster Overview, Node Overview and Namespace Overview. Resources mentioned in the messages are linked to pages in busola

**Related issue(s)**

### DIFF
--- a/core-ui/src/components/Clusters/views/ClusterOverview/ClusterNodes/NodeWarningsList.js
+++ b/core-ui/src/components/Clusters/views/ClusterOverview/ClusterNodes/NodeWarningsList.js
@@ -3,22 +3,23 @@ import LuigiClient from '@luigi-project/client';
 import { Link } from 'fundamental-react';
 import { useGet, GenericList, ReadableCreationTimestamp } from 'react-shared';
 import { useTranslation } from 'react-i18next';
+import { useMessageList, EVENT_MESSAGE_TYPE } from 'hooks/useMessageList';
 
-export function ClusterNodesWarnings({ nodesNames }) {
-  const { i18n } = useTranslation();
+export function ClusterNodesWarnings() {
+  const { i18n, t } = useTranslation();
   const { data, loading, error } = useGet('/api/v1/events');
 
-  const formatInvolvedObject = obj => {
-    if (obj.namespace) {
-      return `${obj.kind} ${obj.namespace}/${obj.name}`;
-    } else {
-      return `${obj.kind} ${obj.name}`;
-    }
-  };
+  const {
+    displayType,
+    sortedItems,
+    formatInvolvedObject,
+    messageSelector,
+  } = useMessageList(data?.items);
 
-  const warnings = data?.items
-    .filter(e => e.type === 'Warning')
-    .filter(e => nodesNames.includes(e.source.host));
+  const entries =
+    displayType === EVENT_MESSAGE_TYPE.ALL
+      ? sortedItems
+      : sortedItems.filter(e => e.type === displayType.type);
 
   const navigateToNodeDetails = nodeName => {
     LuigiClient.linkManager().navigate(`nodes/${nodeName}`);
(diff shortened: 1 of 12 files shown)


core-ui/src/hooks/useMessageList.js
```diff
@@ -0,0 +1,83 @@
...
+    );
+  };
+
+  const messageSelector = (
+    <Dropdown
+      compact
+      options={Object.values(EVENT_MESSAGE_TYPE).map(el => ({
+        key: el,
+        text: t(`node-details.${el.label}`),
+      }))}
```
Wawrzyn321, line 69: Could you use `string`s as a `key` and `selectedKey`?
```suggestion
      options={Object.values(EVENT_MESSAGE_TYPE).map(el => ({
        key: el.type,
        text: t(`node-details.${el.label}`),
        data: el,
      }))}
```

core-ui/src/hooks/useMessageList.js
```diff
@@ -0,0 +1,83 @@
...
+import { useTranslation } from 'react-i18next';
+
+export const EVENT_MESSAGE_TYPE = {
+  ALL: { type: 'All', label: 'all' },
+  NORMAL: { type: 'Normal', label: 'normal' },
+  WARNING: { type: 'Warning', label: 'warnings' },
+};
+
+export const RESOURCE_PATH = {
+  Pod: 'pods',
```
Wawrzyn321, line 14: Would it be (easily) possible to handle `Node`s here, too?

core/src/i18n/en.yaml
```diff
@@ -1115,13 +1115,15 @@ namespaces:
 no-permissions:
   title: No Permissions
 node-details:
+  all: All Messages
```
Wawrzyn321, line 1118: Tbh instead of (all messages, normal messages, warnings) I'd go with (all, info, warnings), but it's just my idea, let's see wait for TWs opinion.

core-ui/src/components/Clusters/views/ClusterOverview/ClusterNodes/NodeWarningsList.js
```diff
@@ -3,22 +3,23 @@ import LuigiClient from '@luigi-project/client';
 import { Link } from 'fundamental-react';
 import { useGet, GenericList, ReadableCreationTimestamp } from 'react-shared';
 import { useTranslation } from 'react-i18next';
+import { useMessageList, EVENT_MESSAGE_TYPE } from 'hooks/useMessageList';
 
-export function ClusterNodesWarnings({ nodesNames }) {
-  const { i18n } = useTranslation();
+export function ClusterNodesWarnings() {
```
Wawrzyn321, line 8: I think we can rename this component now :)

core-ui/src/hooks/useMessageList.js
wchrapka: Nodes have a separate column at the cluster/node view, but added them anyway.

Wawrzyn321 (approved):

Wawrzyn321:
Please fix the missing translation
image

Wawrzyn321 (dismissed):

tests/tests/smoke-tests.spec.js
```diff
@@ -25,7 +25,7 @@ context('Smoke Tests', () => {
       .should('be.visible');
 
     cy.getIframeBody()
-      .contains('Warnings')
+      .contains('All Messages')
```
valentinvieriu, line 28: How do this work with Translations? What happens if translation changes? Do we need to adjust our tests?

Wawrzyn321: If translation changes the tests will fail - as far as I can remember we considered this a _good thing_, as text is the contract and if contract changes, the tests would fail.

Wawrzyn321 (approved):

Wawrzyn321 (approved):

Issue #673, opened 2021-12-06
Repository URL input loses focus after typing the first letter


dariadomagala:
**Description**

`Repository URL` input in the create repositories view loses focus after typing one letter.

**Expected result**

The user should be able to type the whole address at once.

**Actual result**

The user has to click on the input again to have the focus on the input and be able to finish typing the address.

**Steps to reproduce**

Go to create a git repository modal and start adding a `Repository URL`

**Troubleshooting**

PR #675, opened 2021-12-07
Fix blur issue at the git repos URL field

wchrapka:
**Description**

Fixes the issue of repository-url input at *Git Repositories*. The input was losing focus after the first character was typed. 

**Related issue(s)**
#673


Wawrzyn321 (changes requested):
Damn, I was expecting the need of contribution to Fundamental or something...

Would you kindly update our e2e tests? We don't need that workaround anymore in `tests/tests/test-git-repositories.spec.js:44`

Wawrzyn321 (approved):

Issue #680, opened 2021-12-08
Add the Edit Deployment view


dariadomagala:
**Description**

Reuse the create view to also edit a Deployment. Replace the 'edit yaml' button with the side drawer, for the 'edit' button with the create-like view.

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

PR #683, opened 2021-12-08
Add a possibility to edit a deployment using new edit modal

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Resolves https://github.com/kyma-project/busola/issues/680


qbalukom:
what about the "separate service" option? I think it was agreed it should be removed when editing (similar to how limits/quotas are not present when editing a namespace)

qbalukom (approved):

PR #729, opened 2021-12-17
Istio virtual services

qbalukom:


core/src/i18n/en.yaml
```diff
@@ -1568,6 +1570,75 @@ stateful-sets:
...
+    add: add
+    remove: remove
+    request: Request
+    response: Response
+    set: set
+    title: Headers
+  hosts: Hosts
+  http-routes:
+    cors:
+      allow-credentials: Allow Credential
```
grego952, line 1586:
```suggestion
      allow-credentials: Allow Credentials
```

core-ui/src/components/Predefined/Create/VirtualServices/templates.js
```diff
@@ -0,0 +1,11 @@
+export function createTemplate(namespace) {
+  return {
+    apiVersion: 'networking.istio.io/v1beta1',
+    kind: 'VirtualService',
+    metadata: {
+      name: '',
+      namespace: namespace,
+    },
+    spec: {},
```
dariadomagala, line 9: why don't we have anything in the spec? we should have here the minimum to be able to create the VS, name and namespace are not enough.

core/src/i18n/en.yaml
```diff
@@ -1568,6 +1570,75 @@ stateful-sets:
...
+    remove: remove
+    request: Request
+    response: Response
+    set: set
+    title: Headers
+  hosts: Hosts
+  http-routes:
+    cors:
+      allow-credentials: Allow Credentials
+      allow-method: Allow Method
```
dariadomagala, line 1587: we're missing the `allow-header` here as well.

core-ui/src/components/Predefined/Details/VirtualServices/HttpRoutes.js
```diff
@@ -0,0 +1,532 @@
...
+      </ul>
+    </>
+  );
+}
+
+function HttpMatchRequests({ matches }) {
+  const { t } = useTranslation();
+
+  return (
+    <GenericList
```
dariadomagala, line 187: Do we really need a generic list for that? it seems like a big component for such a small view. why won't we use the same approach as we have for cors policies?

dariadomagala: Also, the search is still visible and doesn't work.

dariadomagala: But, with this, you've revealed another tiny bug :)
Please fix this message:
Screen Shot 2021-12-23 at 13 45 35

dariadomagala (changes requested):
1. Why do we display every `Allow Method/Headers` separately like 
```
Allow Method GET
Allow Method POST
...
```
Instead of just 
```
Allow Methods GET, POST, ...
```
?

2. Please change this margin to be the default size
Screen Shot 2021-12-23 at 13 32 45

core-ui/src/components/Predefined/Details/VirtualServices/HttpRoutes.js
qbalukom: the idea of using a generic list is because the matches can be more complex, and as such list entries separate matches, while several rules within a match are line-by-line: 
image

qbalukom: > 2. Please change this margin to be the default size
> 
> Screen Shot 2021-12-23 at 13 32 45

This "margin" stems from the fact that there's an extra panel body in there for additional route info. The condition to check this would be unnecessarily long... 

image

dariadomagala: we can remove the double margin for the matches panel (or wherever the margin comes from) and add it for the additional info panel, so there is no complex condition.

qbalukom: but we're using generic styling from fundamental components for this...

core/src/i18n/en.yaml
```diff
@@ -1603,9 +1603,10 @@ virtualservices:
   http-routes:
     cors:
       allow-credentials: Allow Credentials
-      allow-method: Allow Method
+      allow-headers: Allow Headers
+      allow-methods: Allow Method
```
dariadomagala, line 1607:
```suggestion
      allow-methods: Allow Methods
```

core/src/i18n/en.yaml
```diff
@@ -1603,9 +1603,10 @@ virtualservices:
...
       allow-origin: Allow Origin
```
dariadomagala, line 1608:
```suggestion
      allow-origin: Allow Origins
```

core-ui/src/components/Predefined/Details/VirtualServices/common.js
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <>
+      {match.port && (
+        <>
+          <dd>{t('virtualservices.matches.port')}</dd>
+          <dt>{match.port}</dt>
+        </>
+      )}
+      {match.sourceLabels &&
```
dariadomagala, line 48: Why don't we have it as a list with commas as well?

core-ui/src/components/Predefined/Details/VirtualServices/HttpRoutes.js
```diff
@@ -0,0 +1,532 @@
...
+    operator = '~';
+    value = def.regex;
+    tooltip = t('virtualservices.matches.regex');
+  } else {
+    return '-';
+  }
+
+  return (
+    <>
+      <Tooltip className="match-operator" content={tooltip}>
```
dariadomagala, line 399: I had to read the code to realize there's a tooltip here. In my opinion, it should be more obvious for the user. Maybe we could add a question mark icon?

core-ui/src/components/Predefined/Details/VirtualServices/HttpRoutes.js
```diff
@@ -375,31 +381,24 @@ function HttpFaultInjectionAbort({ abort }) {
 export function StringMatch({ def, ignoreCase }) {
   const { t } = useTranslation();
 
-  let operator;
+  let label;
   let value;
-  let tooltip;
   if (def.exact) {
```
dariadomagala, line 386: Can we use Object.keys instead of so many similar ifs?

core-ui/src/components/Predefined/Details/VirtualServices/HttpRoutes.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { GenericList, Tooltip } from 'react-shared';
-import { LayoutPanel } from 'fundamental-react';
+import { LayoutPanel, Icon } from 'fundamental-react';
```
dariadomagala, line 4: We don't use it here

core-ui/src/components/Predefined/Details/VirtualServices/TlsRoutes.js
```diff
@@ -11,11 +11,15 @@ function TlsMatchAttributesItem({ match }) {
   return (
     <dl>
       <dd>{t('virtualservices.matches.sni-hosts')}</dd>
-      <dt>{match.sniHosts.join(', ')}</dt>
+      {match.sniHosts.map(host => (
```
dariadomagala, line 14:
```suggestion
      {match.sniHosts?.map(host => (
```
Better safe than sorry

core/src/i18n/en.yaml
```diff
@@ -410,7 +410,7 @@ common:
     resource-namespace: '{{resource}} $t(common.labels.namespace)'
     secret: Secret
     service: Service
-    version: Version
+    version: Kyma Dashboard Version
```
dariadomagala, line 417: Where did it come from?

qbalukom: hmm... from the merge... interesting

core-ui/src/components/Predefined/Details/VirtualServices/TlsRoutes.js
qbalukom: technically it's required...

dariadomagala (approved):

core/src/i18n/en.yaml
```diff
@@ -1610,6 +1612,76 @@ stateful-sets:
 top-nav:
   profile:
     preferences: $t(preferences.title)
+virtualservices:
+  description: <0>Virtual Service</0> describes a configuration affecting traffic routing.
```
klaudiagrz, line 1616:
```suggestion
  description: <0>Virtual Service</0> describes a configuration that affects traffic routing.
```

klaudiagrz (approved):

akucharska (approved):

PR #737, opened 2021-12-20
Feature create event subscription and list with details

### LINKED ISSUE #702: Introduce Events Listing and Events Detail View
As [Sebastian](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#sebastian---tech-lead) and [Mathew](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#matthew---cloud-native-software-engineer) I would like to be able to be able to see all the Events in my namespace and be able to modify them using the Busola UI.

**Acceptance Criteria**

- [x] Follow Project DOD
- [x] Events listing and detail follow the same pattern as other List Views and Detail View
- [x] Edit mode is present in the Detail View - creating the edit view is not part of this task
- [x] TW are consulted regarding the location in the Menu
- [x] Events in Funtions will provide deep links
- [x] Tests are present testing the functionality

**Tasks**

- [ ] Add Tasks

mrCherry97:
**Description**
Feature create event subscription and list with details

- Added Event Subscription to Left Nav
- Added Event Subscription list and details
- Added Event Subscription create
- Added Event Subscription deep linking
- Added Event Subscription translations
- Added Event Subscription test for:
   - create via YAML
   - list details

**Related issue(s)**
Closes #702


core-ui/src/shared/components/EventSubscriptionConditionStatus.js
```diff
@@ -0,0 +1,29 @@
...
+          type: 'error',
+          text: t('common.statuses.error'),
+          tooltipMessage: condition?.message,
+        };
+  return (
+    <StatusBadge
+      type={statusBadgeProperties.type}
+      additionalContent={statusBadgeProperties.tooltipMessage}
+      i18n={i18n}
+      noTooltip={condition?.status !== 'False' && !isListView}
```
dariadomagala, line 24: Why don't we show the ready status for the details view?

dariadomagala (changes requested):
One of the DOD says `Events in Funtions will provide deep links` yet there are no deep links on the functions view?
(Additionally, I would also add the links to the list of subscriptions on the services details view since it's the same list)

core-ui/src/shared/components/EventSubscriptionConditionStatus.js
Lyczeq: I asked @Wawrzyn321 what he thought about it and he said that there was no need to inform a user twice because the `Type` information was already there (e.g. 'Subscription Active'). We can discuss it with other team members.

Lyczeq: Referring to your first comment, I'll add deep links for functions and services.

core-ui/src/components/Predefined/Details/EventSubscriptions/EventSubscriptions.details.js
```diff
@@ -0,0 +1,161 @@
...
+  const conditions = eventSubscription?.status?.conditions;
+  const headerRenderer = _ => [
+    t('event-subscription.conditions.last-transition-time'),
+    t('event-subscription.conditions.reason'),
+    t('event-subscription.conditions.status'),
+    t('event-subscription.conditions.type'),
+  ];
+
+  const rowRenderer = condition => [
+    <FormattedDatetime
```
dariadomagala, line 27: Instead of the `FormattedDatetime` component, we should use the `ReadableCreationTimestamp` one here, as we use everywhere else (we use FormattedDatetime only to display the expiration date, which should be displayed differently).

core/src/i18n/en.yaml
```diff
@@ -780,6 +781,12 @@ domains:
     label: Include Domains
     placeholder: Domain that is allowed
 event-subscription:
+  conditions:
+    last-transition-time: Last transition time
```
dariadomagala, line 785: We have `Last transition` in the jobs view and `Last transition time` here. You can pick any of those, but let's just agree to use the same one in both views.

shared/components/ResourceDetails/ResourceDetails.js
```diff
@@ -134,8 +134,11 @@ function Resource({
   silentRefetch,
   updateResourceMutation,
   windowTitle,
+  resourceTitle,
```
dariadomagala, line 137: I don't know what was the agreement, so correct me if I'm wrong. 
Shouldn't we display in the Busola exactly what's on a cluster? So if a resource is called `Subscription` then we should show `Subscription`, not `Event Subscription`? In theory, it's a small difference, I just wonder if it could confuse the user that those are different resources.

core-ui/src/components/Predefined/Create/EventSubscriptions/EventSubscriptions.create.js
```diff
@@ -0,0 +1,30 @@
...
+  resourceUrl,
+}) => {
+  const [eventSubscription, setEventSubscription] = useState(
+    createEventSubscriptionTemplate(namespace),
+  );
+
+  return (
+    <ResourceForm
+      pluralKind="eventsubscriptions"
+      singularName="Event Subscription"
```
dariadomagala, line 20: I believe it should be translated

dariadomagala (changes requested):

core-ui/src/shared/components/EventSubscriptionConditionStatus.js
```diff
@@ -0,0 +1,28 @@
...
+      : {
+          type: 'error',
+          text: t('common.statuses.error'),
+          tooltipMessage: condition?.message,
+        };
+  return (
+    <StatusBadge
+      type={statusBadgeProperties.type}
+      additionalContent={statusBadgeProperties.tooltipMessage}
+      i18n={i18n}
```
dariadomagala, line 23: Let's add a translation to the `ready` state, instead of getting it from the condition type. 
To do that, remove the `condition?.type` from the tooltipMessage above, add `resourceKind='event-subscription'` here, and add a transtaltion to the `en.yaml` file.

core-ui/src/components/Predefined/Details/EventSubscriptions/EventFilters.scss
```diff
@@ -0,0 +1,29 @@
+.event-filters-panel {
```
dariadomagala, line 1: please add a margin above the `property`

core-ui/src/components/Predefined/Details/EventSubscriptions/EventSubscriptions.details.js
```diff
@@ -0,0 +1,125 @@
...
+    },
+    {
+      header: t('event-subscription.sink'),
+      value: ({ spec }) => {
+        const index = spec?.sink.lastIndexOf('/') + 1;
+
+        const firstDot = spec?.sink.indexOf('.');
+        const serviceName = spec?.sink.substring(index, firstDot);
+        return spec?.sink ? (
+          <Link
```
dariadomagala, line 99: Just a thought, maybe we should keep the sink just as a text, without a redirection, and add a link to the service in parenthesis (like it is in the 'controlled by' columns) or just to a separate column? 
If I click on something that's an address, I would expect to be redirected to this address, not to a different view.

dariadomagala (dismissed):

core/src/i18n/en.yaml
```diff
@@ -795,12 +802,24 @@ event-subscription:
...
+  description: <0>Event Subscription</0> is used to subscribe to events.
   errors:
     invalid: The eventType field must have a proper value.
     not-matching-search-query: Couldn't find Event Subscriptions matching this query.
     server: Server error. Contact your cluster admin.
     subscriptions-not-found: There are no Event Subscriptions yet.
+  filters:
+    event-source: Event Source
+    event-type: Event Type
+    property: property
```
klaudiagrz, line 814: Shouldn't `property`, `type` and `value` be capitalized as well?

core/src/i18n/en.yaml
```diff
@@ -421,6 +421,7 @@ common:
     error: 'Error: {{error}}'
     fill-required-data: Fill the remaining required data.
     loading-name: Loading name...
+    no-entries-found: No entries found
```
klaudiagrz, line 424:
```suggestion
    no-entries-found: No entries found.
```

core/src/i18n/en.yaml
```diff
@@ -1194,7 +1213,7 @@ jobs:
     complete: Job completed.
     not-complete: Job in progress.
 kubeconfig-id:
-  error: "Couldn't load kubeconfig ID; configuration not changed (Error: ${{error}})."
+  error: 'Couldn''t load kubeconfig ID; configuration not changed (Error: ${{error}}).'
```
klaudiagrz:
```suggestion
  error: 'Couldn't load a kubeconfig ID; configuration not changed (Error: ${{error}}).'
```

klaudiagrz (approved):

dariadomagala (approved):

PR #758, opened 2021-12-28
Protect Applications on SKRs

### LINKED ISSUE #712: Busola allows to edit Applications (External systems) in Managed Kyma Runtime
**Description**

Busola allows user to Create / Delete and Connect Application representing external systems in Kyma.
This is a standard feature of Kyma Open Source but it should not be the case for Managed Kyma Runtime (SKR). 
In SKR mode applications are created and managed centrally with Kyma Management Plane and propagated to subscribed runtimes with a help of `Compass Runtime Agent` service. User should not be able to modify them at the level of Kyma Runtime.

**Expected result**

Following UI operations for Applications managed by the Runtime Agent should not be possible in Busola:
- Delete Application
- Connect Application 

On SKR clusters there should also be no possibility to use the `Connect Application` button

**Actual result**

Busola allows to: 

- Modify existing Compass application.
- Delete existing Compass application.
- Start procedure to generate connection token for existing application. The procedure fails since it is not possible to generate token in managed Kyma Runtime. After clicking the button `Connect Application` the dialog pops up and does not contain any valid token value.

**Steps to reproduce**

1. Provision Kyma 2.0 as SKR from BTP (using preview plan)
2. Open Busola
3. Observe UI controls that allows to edit applications. You can also enter the application and try to generate token for connection.

Wawrzyn321:
**Description**

- Protect applications
- Adjust messages

**Testing**

- Disabling `Connect Application` - I've added configmap on `sundown` cluster (`kubectl -n kyma-system get configmap skr-configmap -oyaml`), edit the `data` to disable protection.
- Protecting Applications - add
```
"PROTECTED_RESOURCES": {
        "isEnabled": true,
        "config": {
          "resources": [
            {
              "match": { "$.metadata.labels['applicationconnector.kyma-project.io/managed-by']": "compass-runtime-agent" }
            }
          ]
        }
      }
```
to local `config.json`.

**Related issue(s)**

Resolves #712


akucharska (dismissed):

core/src/i18n/en.yaml
```diff
@@ -108,12 +108,16 @@ applications:
     namespace: Namespace
   messages:
     app-created: Application {{applicationName}} created.
+    binding-create-failed: Failed to create the Binding.
+    binding-created: Binding created.
     binding-deleted: Binding deleted.
     binding-not-found: No bindings.
```
klaudiagrz, line 114:
```suggestion
    binding-not-found: No Bindings.
```

klaudiagrz: Please be consistent and use either upper- or lowercase 🙂

core/src/i18n/en.yaml
```diff
@@ -108,12 +108,16 @@ applications:
...
     cannot-create: Cannot create the Application.
+    connect-disabled: Connecting an Application is disabled for SKR runtime.
```
klaudiagrz, line 116:
```suggestion
    connect-disabled: Connecting the Application is disabled for SAP Kyma Runtime.
```

klaudiagrz: I'd either go with `SAP Kyma Runtime` or `SKR` because the SKR abbreviation already has the word "runtime" included.

core/src/i18n/en.yaml
```diff
@@ -108,12 +108,16 @@ applications:
...
     copy-url: 'Copy the following URL and use it at the external system that you would like to connect to:'
-    create-failed: Failed to create the Binding.
-    created: Binding created.
+    create-disabled: Creating an Application is disabled for SKR runtime.
```
klaudiagrz, line 118:
```suggestion
    create-disabled: Creating the Application is disabled for SAP Kyma Runtime.
```

core/src/i18n/en.yaml
klaudiagrz: I'd also be consistent with the articles. In previous messages, you use definite articles (`the` Bindings, `the` Application), so either change all of them to indefinite articles, or to definite ones. Let's discuss 🙂

Wawrzyn321: I thought `the` is ok for one-and-only application we were working on and something went wrong (thus `the Application`, `that Application`), and `a` is good for just _any_ application -it's a general rule, so it's not about any particular app.

klaudiagrz: Yes, you are right 🙂 That is why I also thought it would be better to use an indefinite article in all the cases above (so: `a` Binding, `an` Application). I'll adjust the suggestions then!

klaudiagrz: Ok, I didn't realize the subtle differences between these messages. Thanks for clarifying!

klaudiagrz (approved):

akucharska (dismissed):

klaudiagrz (approved):

qbalukom (approved):

PR #773, opened 2022-01-05
Visualize crds

qbalukom:


mrCherry97:
spec and DNS just go for a walk 
image

mrCherry97:
Some strings have [] is it neccesary?
image

mrCherry97 (commented):
Some improvement idea, maybe for a separate task. Can we make all the links real references?
image

qbalukom:
> Some strings have [] is it neccesary? image With objects is the same problem image

yes, because those are an array of strings and an array of objects

akucharska (dismissed):

NHingerl (dismissed):

NHingerl (approved):

PR #791, opened 2022-01-11
Refactor and repair checking logs for eventing

### LINKED ISSUE #777: Check function logs for Cypress in-cluster Eventing flow
**Description**

Loading the logs for function doesn't work in Cypress. [We have commented this step.](https://github.com/kyma-project/busola/blob/main/tests/tests/in-cluster-eventing.spec.js#L131) 
Maybe we don’t need to use the UI for reading the logs? Can we do api calls and just check the json?

**Reasons**

This is important flow and we should test it regularly in automated way.

**Attachments**

https://github.com/cypress-io/cypress/issues/1185#issuecomment-856745211
It didn't help too: cypress-io/cypress#15259 (comment)

mrCherry97:
**Description**
We missed tests for checking logs in the eventing flow.

- Added tests for checking logs after triggering function

**Related issue(s)**
Closes #777


tests/tests/in-cluster-eventing.spec.js
```diff
@@ -121,17 +119,30 @@ context('In-cluster eventing', () => {
...
+          cy.request({
+            method: 'GET',
+            url: kubeconfig.clusters[0].cluster.server + requestUrl,
+            timeout: 10000,
+            headers: {
+              authorization: 'Bearer ' + kubeconfig.users[0].user.token,
+            },
+          }).then(response => {
+            // response.body is automatically serialized into JSON
+            expect(response.body).to.match(/^Payload/);
```
Sawthis, line 142: Shouldn't we check if we received "Event received" log entry?

Sawthis (changes requested):

tests/tests/in-cluster-eventing.spec.js
mrCherry97: No, we test eventing for publisher function, not for receiver.

mrCherry97: We have ready `API Rule` for the publisher so that's why I'm testing publisher, not reciver.

Sawthis: But we should test if receiver received the event.

Wawrzyn321: In the end, we test both.

Wawrzyn321 (approved):

Sawthis (approved):

PR #816, opened 2022-01-17
Improve validation

### LINKED ISSUE #774: Mark wrong inputs in Subscriptions
**Description**
As a user, I want to know where I made a mistake in creating or editing form for Subscription.

**Reasons**
Fields with wrong data aren't marked. The user has to be informed where he made a mistake.

**Acceptance Criteria**
- [ ] Validation for all fields
- [ ] Fields with wrong input is marked on red
- [ ] Error Message should be displayed under the field - use fundamental method

We should validate the inputs in the way they're validated on the backend. The following links might be very helpful: 
- https://github.com/kyma-project/kyma/blob/95e0208f3536003cec82877049b16416bcc92671/components/eventing-controller/pkg/handlers/nats.go#L279
- https://github.com/kyma-project/kyma/blob/95e0208f3536003cec82877049b16416bcc92671/components/eventing-controller/pkg/handlers/eventtype/clean.go#L46
- https://github.com/kyma-project/kyma/blob/95e0208f3536003cec82877049b16416bcc92671/components/eventing-controller/pkg/handlers/eventtype/parse.go#L13
- https://github.com/kyma-project/kyma/blob/95e0208f3536003cec82877049b16416bcc92671/components/eventing-controller/pkg/handlers/eventtype/clean.go#L15

qbalukom:
Closes #774


dariadomagala (changes requested):
1. When I try to add a new filter:
Screen Shot 2022-01-19 at 11 56 23
2. We should also add validation in the simple view for the event name to consist of a dot.
3. Should we disable an option to use dashes in the app name or should we allow it, and the backend will change it?

Wawrzyn321:
1. Fixed
2. Done
3. I have no idea, backend regex does not mention '-', but subscription created with dash is errorless... I've added it to whitelist.

core/src/i18n/en.yaml
```diff
@@ -1722,6 +1723,12 @@ subscriptions:
   description: <0>Subscription</0> is used to subscribe to events.
   errors:
     empty-event-types: You must provide at least one event type.
+    event-name-required: Event name is required.
+    event-name-segments: Event name must consist of at least 2 non-empty segments, separated by "." character.
```
dariadomagala, line 1727: Wouldn't that be simplier?
```suggestion
    event-name-segments: Event name must consist of at least 2 non-empty segments, separated by a dot.
```

dariadomagala (commented):
If I add a new filter, the 'Create' button is still active, even tho the new filter with a default value (so only a prefix), shouldn't be considered valid.

dariadomagala (dismissed):

grego952 (dismissed):

dariadomagala (dismissed):

grego952 (approved):

dariadomagala (approved):

PR #819, opened 2022-01-17
Changed delete modal

### LINKED ISSUE #801: Deletion from the details view uses wrong confirmation modal
**Description**

After clicking a `delete` button, we should show the confirmation modal with the option to `don't show it again`. When deleting a resource from its details page, we show an old confirmation modal, without this option.

**Expected result**

* we should have the same UI everywhere
* the user should be able to use `don't show it again` with any deletion
* if the user has this option already enabled, they should now see the confirmation modal anywhere

**Actual result**

* we don't use the same UI everywhere
* the user is not able to use `don't show it again` with deletions from the details views
* if the user has this option already enabled, they still see the confirmation modal with deletions from the details views

**Steps to reproduce**

**Troubleshooting**

mrCherry97:
**Description**
On the details page, we had the old delete modal.

- Refactor of deleting modal,
- Move deletion modal to separated hook
- ...

**Related issue(s)**
Closes #801


shared/hooks/useDeleteResource.js
```diff
@@ -0,0 +1,111 @@
...
+      value: dontConfirmDelete,
+    });
+    try {
+      await deleteResourceMutation(url);
+      notification.notifySuccess({
+        content: t('components.resources-list.messages.delete.success', {
+          resourceType: prettifiedResourceName,
+        }),
+      });
+      navigateToList(resourceType);
```
dariadomagala, line 42: it doesn't work well if we delete a resource from a list that's on a details page. for example - go to the details of a function, and delete a pod/api rule that's connected to it. it should stay in the same view, yet we're redirected to the list of functions.

dariadomagala (changes requested):

dariadomagala (approved):

PR #825, opened 2022-01-18
Fix roles submit button disable state

### LINKED ISSUE #771: In Roles user cannot click Create in Create Role view after entering all required data
**Description**

When user click **Create Role** in **Roles** and provide all required data, they cannot click **Create** straight away. Instead they have to click somewhere outside the placeholder for the button to work.

**Expected result**
User should be able to click **Create** straight away after entering all required data.

**Actual result**

**Steps to reproduce**
1. Go to **Roles**
2. Click **Create Role**
3. Enter the name
4. Choose required **API Groups**, **Resources** and **Verbs**
5. **Create** button remains inactive until you click somewhere outside the placeholder.

wchrapka:
**Description**

Fixes a bug in Roles with submit button

**Related issue(s)**
Fixes #771


dariadomagala (changes requested):
The problem occurs with an edit. If I would want to add a Resource to an existing Role, the first time I open a dropdown, I can only see the default `*` option. I would have to click on the API group dropdown, to get the list of available Resources next.

wchrapka:
> Now on edit, we make every call twice, can we reduce it?

Hmm, it does so as ResourceFormWrapper mounts (not renders) all its children twice (so it resets all the states, refs etc. to default). I would start a new bug with a request to optimize that behavior. Locally I can introduce only a nasty workaround

dariadomagala (approved):

PR #826, opened 2022-01-19
Events

akucharska:
**Description**

- Add events

**Related issue(s)**


core-ui/src/components/Predefined/Details/Namespace/Namespace.details.js
```diff
@@ -63,11 +64,7 @@ export const NamespacesDetails = ({ DefaultRenderer, ...otherParams }) => {
     />
   ) : null;
 
-  const eventsParams = {
-    namespace: otherParams.resourceName,
-  };
-
-  const Events = <ComponentForList name="eventsList" params={eventsParams} />;
+  const Events = <EventsList namespace={otherParams.resourceName} />;
```
dariadomagala, line 67: maybe we could hide the `namespace` column when on a namespace details? it would give us more space for the remaining fields, that are all pretty wide

dariadomagala: Same thing in other places - do we need the `namespace` and `involved object` columns when being on the for example pods details page?

dariadomagala (changes requested):
1. Just a question - maybe we could sort it by the last seen 
Screen Shot 2022-01-20 at 10 55 41

2. Missing events in the Storage Class details

core-ui/src/components/Predefined/List/Events.list.js
```diff
@@ -4,68 +4,128 @@ import {
...
-} from 'hooks/useMessageList';
+  Link as DescriptionLink,
+  PageHeader,
+  prettifyNamePlural,
+} from 'react-shared';
+import { Icon, Link } from 'fundamental-react';
+import { useMessageList, EVENT_MESSAGE_TYPE } from 'hooks/useMessageList';
+
+export const EventsList = ({ ...otherParams }) => {
+  // TODO EVENTS DESCRIPTION
```
dariadomagala, line 17: todo?

core-ui/src/components/Predefined/Details/Pod/Pod.details.js
```diff
@@ -29,6 +30,14 @@ function goToSecretDetails(resourceKind, name) {
 
 export const PodsDetails = ({ DefaultRenderer, ...otherParams }) => {
   const { t, i18n } = useTranslation();
+
+  const filterByHost = e =>
+    e.involvedObject.name === otherParams.resourceName &&
+    e.involvedObject.kind === 'Pod';
```
dariadomagala, line 36: We have this function used many times, the only difference is the kind used. Can we move it to shared and import it, instead of writing the same code multiple times?

dariadomagala: Also, I would add `?` just in case the `involvedObject` is undefined - we know that k8s likes to surprise us this way from time to time

core-ui/src/components/Predefined/List/Events.list.js
```diff
@@ -4,68 +4,128 @@ import {
...
-    t('namespaces.events.headers.object'),
-    t('namespaces.events.headers.object-type'),
-    t('namespaces.events.headers.created'),
+    t('common.labels.name'),
+    t('events.headers.type'),
+    t('events.headers.message'),
+    t('common.labels.namespace'),
+    t('events.headers.involved-object'),
+    t('events.headers.source'),
+    t('events.headers.count'),
```
dariadomagala, line 75: Maybe we could move this column to the details only?

akucharska: Together with Valentin, we decided to reduce the space needed for type columb (use icon not text) and to remove the namespace completely because the same information is in the involved object column

core-ui/src/components/Predefined/Details/Namespace/Namespace.details.js
akucharska: I removed the namespace but left the involved object. 

Moreover, a namespace with the deep link has been added to the details view. The reason for it is that going from cluster-wide event list to details will redirect users to details in a specific namespace... since every event has a namespace. And we don't want them to overlook it

core-ui/src/components/Predefined/Details/Events.details.js
```diff
@@ -0,0 +1,113 @@
...
+      value: event => (
+        <p>
+          {event.type}{' '}
+          {event.type === 'Warning' ? (
+            <Tooltip content={event.type}>
+              <Icon
+                ariaLabel="Warning"
+                glyph="message-warning"
+                size="s"
+                className="fd-has-color-status-2 cursor-pointer"
```
dariadomagala, line 77: why do we have cursor-pointer instead of a question mark as we have in other places with tooltips?

akucharska: Because I used wrong class by mistake

dariadomagala (approved):

grego952 (dismissed):

dariadomagala (approved):

grego952 (approved):

dariadomagala (approved):

PR #872, opened 2022-01-25
Plural form of some of the options in Cron Job, Job, Gateway, Oath2 Client, and DNE Entry views

### LINKED ISSUE #817: Components that allow for choosing/typing more than one option should be in plural form
**Description**
Not all components that allow for choosing/typing more than one option are in plural form all around Busola. It should unified. For example:
Zrzut ekranu 2022-01-17 o 11 18 32
Slectors

**Expected result**
Whenever a component has a possibility to choose/type multiple items, it should be written in plural form.

**Actual result**

**Steps to reproduce**

**Troubleshooting**

grego952:
**Description**

- add plural form to CRD with multiple choice/type option
- small text refactor

**Related issue(s)**
Resolves #817

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -506,7 +506,7 @@ common:
     duplicate-key: Duplicate key
     error: Error
     hide: Hide read-only fields
-    k8s-name-input: 'Name must consist of lowercase alphanumeric characters, can contain ''-'' (e.g.: ''my-name1'').'
+    k8s-name-input: "Name must consist of lowercase alphanumeric characters, can contain '-' (e.g.: 'my-name1')."
     key: Key must start and end with an alphanumeric character, can contain '-', '_' or '.'.
     key-value: key=value, must start and end with an alphanumeric character, can contain '-', '_' or '.'.
     read-file: Read file content as a single value, with file name as a key.
@@ -1273,7 +1273,7 @@ jobs:
     complete: Job completed.
     not-complete: Job in progress.
 kubeconfig-id:
-  error: 'Couldn''t load kubeconfig ID; configuration not changed (Error: ${{error}}).'
+  error: "Couldn't load kubeconfig ID; configuration not changed (Error: ${{error}})."
   must-be-an-object: Kubeconfig must be a JSON or YAML object.
 legal:
   copyright: Copyright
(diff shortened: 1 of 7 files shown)


core/src/i18n/en.yaml
```diff
@@ -1244,7 +1244,7 @@ jobs:
     containers-readonly-in-edit: After a Job is created, the containers are read-only.
     labels:
       args: Args
-      command: Command
+      command: Commands
```
Wawrzyn321, line 1247: It's a single command, consisting of multiple parts.
```suggestion
      command: Command
```

core/src/i18n/en.yaml
```diff
@@ -1407,7 +1407,7 @@ persistent-volume-claims:
     match-expressions: Match expressions
     match-labels: Match labels
     operator: Operator
-    selector: Selector
+    selector: Selectors
```
Wawrzyn321, line 1410:
```suggestion
    selector: Selector
```
K8s calls it "selector".

core/src/i18n/en.yaml
```diff
@@ -1378,7 +1378,7 @@ oauth2-clients:
     post-logout-redirect-uris: Post-Logout Redirect URIs
     redirect-uris: Redirect URIs
     response-types: Response Types
-    scope: Scope
+    scope: Scopes
```
Wawrzyn321, line 1381: I'd go with "scope", for example Luigi calls it like that https://docs.luigi-project.io/docs/authorization-configuration, even though there are multiple 
```suggestion
    scope: Scope
```

Wawrzyn321: `scope defines the permissions to request at login.`
yeah, it's a single scope with a set of permissions.

grego952: But here you have scopes - https://docs.apigee.com/api-platform/security/oauth/working-scopes

Wawrzyn321: Yup, and there's a `scope` in JSON
```
  "scope" : "A",
```
I guess they named the article like that, just like https://www.ocg.nsw.gov.au/child-safe-organisations/working-with-children-check, even though you could also work with one child.

Wawrzyn321 (approved):

klaudiagrz (approved):

PR #916, opened 2022-02-04
implement storage of hiding read-only fields

qbalukom:


dariadomagala (changes requested):
It doesn't work as expected.
Example:
* go to pods details
* click edit yaml
* click hide
* `managedFields` are hidden
* close the drawer
* click edit yaml again
* `managedFields` are not hidden, yet `annotations` are even tho they shouldn't be

dariadomagala (approved):

dariadomagala (approved):

PR #937, opened 2022-02-08
Initial version

### LINKED ISSUE #879: Introduce Feature to protect /backend calls by JWT token
At this moment we are using ISTIO to protect /backend to be reachable with a valid JWT token. As we plan to move to a Google Cloud Run hosting, we will need to support this feature at our backend level.

Acceptance Criteria:

- [ ] This is feature flagged so it can be turned on or off via env variables
- [ ] technical solution will have performance in mind
- [ ] we will just validate if we received a valid jwt token

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Load testing**
For testing I added an early return just before performing actual request to apiserver - we're comparing the jwt-check, so hitting k8s is not needed. 
```
from locust import HttpUser, between, task

class WebsiteUser(HttpUser):
    wait_time = between(5, 15)

    @task
    def index(self):
        self.client.get("/backend/apis", headers=dict({
            'Authorization': 'Bearer yadada',
            'X-Cluster-Certificate-Authority-Data': 'yadada',
            'X-Cluster-Url': 'https://api.kmain.hasselhoff.shoot.canary.k8s-hana.ondemand.com',
            'X-K8s-Authorization': '<copy token from browser>'
        }))
```

**Related issue(s)**

Closes #879


core/src/assets/config/config.json
```diff
@@ -66,6 +66,13 @@
             }
           ]
         }
+      },
+      "JWT_CHECK_CONFIG": {
```
dariadomagala, line 70: this needs to be properly documented

dariadomagala (commented):
You also need to bump the web deployment

docs/features.md
```diff
@@ -127,6 +127,18 @@ The document lists and describes all the feature flags that are available in Kym
   },
   ```
 
+- **JWT_CHECK_CONFIG** – is used to configure data necessary for the backend authentication, such as an issuer and JWKS (JSON Web Key Set) address.
```
dariadomagala, line 130: Should we describe the behavior if we don't have this feature enabled?

grego952 (approved):

PR #940, opened 2022-02-09
Add possibility to disable sidecars for jobs and cronjobs

akucharska:
**Description**

- Add possibility to disable sidecars for jobs and cronjobs

**Related issue(s)**


core/src/i18n/en.yaml
```diff
@@ -615,6 +615,7 @@ cron-jobs:
   create-modal:
     day-of-month: Day of Month
     day-of-week: Day of Week
+    disable-sidecar: Disable sidecar
```
Wawrzyn321, line 618: Maybe remove that translation and use the one from `jobs` section? Common stuff for Jobs and CronJobs is here anyway (see `ContainerSection.js`, which uses only `jobs.` prefixed translation keys and is used for both Jobs and CJs).

Wawrzyn321 (changes requested):
View breaks on entering the CronJob create modal
image

core-ui/src/components/Predefined/Create/Jobs/CronJobs.create.js
```diff
@@ -29,15 +34,71 @@ function CronJobsCreate({
...
+                ...cronJob.spec.jobTemplate.spec.template,
+                metadata: {
+                  ...cronJob.spec.jobTemplate.spec.template.metadata,
+                  annotations: jobTemplateAnnotations,
+                },
+              },
+            },
+          },
+        },
+      });
```
Wawrzyn321, line 85: Maybe use
```
jp.value(cronJob, 'turbo-long-path', jobTemplateAnnotations);
setCronJob({...cronJob});
```
?

Wawrzyn321 (approved):

grego952 (approved):

PR #949, opened 2022-02-09
Add error boundary to monaco editors

### LINKED ISSUE #909: Use error boundaries in Busola
**Description**

Introduce "catch-all" style handlers in various places of Busola - especially for create/edit modals.Those handlers would show small message or handle the error gracefully instead of crashing the entire MF.

**Reasons**

Even though we validate our inputs, users can type anything into the YAML editor - and for example if we rely on some value being of particular type, then an error may arise when user types e.g. object instead of an array.

Remember to rethrow caught errors again, so that Sentry can catch them and display on dashboard - they are _not_ expected (or use Sentry functionality to capture the error, if such functionality exists).

wchrapka:
**Description**

Adds Error boundary component for monaco editors

**Related issue(s)**

Resolves #909


dariadomagala (changes requested):
We don't have the boundaries for the 'add cluster' editor.

dariadomagala:
Also, could we add a test (might be just a unit test) for that?

shared/components/ErrorBoundary/ErrorBoundary.js
```diff
@@ -0,0 +1,40 @@
...
+    // Update state so the next render will show the fallback UI.
+    return { hasError: true, error };
+  }
+
+  render() {
+    if (this.state.hasError) {
+      return (
+        <div role="alert">
+          <p className="fd-margin-top-bottom--sm">
+            {this.props.t('err-boundary.restored-initial-form')}
```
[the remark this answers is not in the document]
wchrapka: option added

Wawrzyn321 (changes requested):
Would you kindly fix this case? When I paste such "kubeconfig", the error boundary is triggered, but then the error show up anyway:
```apiVersion: v1
kind: Config
current-context: shoot--hasselhoff--kmain
contexts:
  - name: shoot--hasselhoff--kmain
    context:
      cluster: shoot--hasselhoff--kmain
      user: shoot--hasselhoff--kmain-token
clusters:
  - name: shoot--hasselhoff--kmain
    cluster:
      server: https://api.kmain.hasselhoff.shoot.canary.k8s-hana.ondemand.com
      certificate-authority-data: 'yadadada'
users: 2 
```

wchrapka:
> Would you kindly fix this case? When I paste such "kubeconfig", the error boundary is triggered, but then the error show up anyway:
> 
> ```
> kind: Config
> current-context: shoot--hasselhoff--kmain
> contexts:
>   - name: shoot--hasselhoff--kmain
>     context:
>       cluster: shoot--hasselhoff--kmain
>       user: shoot--hasselhoff--kmain-token
> clusters:
>   - name: shoot--hasselhoff--kmain
>     cluster:
>       server: https://api.kmain.hasselhoff.shoot.canary.k8s-hana.ondemand.com
>       certificate-authority-data: 'yadadada'
> users: 2 
> ```

That's the expected behavior on dev server, I haven't observed this issue in other envrionments. Could you describe how to reproduce?

Wawrzyn321 (approved):

core/src/i18n/en.yaml
```diff
@@ -862,6 +862,9 @@ domains:
   include:
     label: Include Domains
     placeholder: Domain that is allowed
+err-boundary:
+  go-back: Go Back
```
grego952, line 866:
```suggestion
  go-back: Go back
```

core/src/i18n/en.yaml
```diff
@@ -862,6 +862,9 @@ domains:
...
+  restored-initial-form: Something went wrong, we restored the initial form for you.
```
grego952, line 867:
```suggestion
  restored-initial-form: Something went wrong We restored the initial form for you.
```

Wawrzyn321 (approved):

grego952 (approved):

PR #957, opened 2022-02-11
Add links to CRs with custom busola pages

### LINKED ISSUE #950: The Deep links for the CR instances that have a custom view should point to the right landing page
**Description**

For some of the CR in our system ( like Function, VirtualServices, API Rule ... ) we have specific views. Would be great to deep link to those pages and not to the general pages.
**Reasons**

**Attachments**

image
image

wchrapka:
**Description**

Recognizes if a CRD has its custom busola page (like destination rules) and creates an appropriate link in the CRD view.

**Related issue(s)**
Resolves #950

### DIFF
--- a/backend/config/config.json
+++ b/backend/config/config.json
@@ -1,0 +1,1 @@
+../../core/src/assets/config/config.json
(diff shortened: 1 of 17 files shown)


Wawrzyn321 (changes requested):
I regret to inform you it doesn't work for addons.

PR #999, opened 2022-02-22
remove unnecessary texts from placeholders

### LINKED ISSUE #978: Information in placeholders should not repeat information from the label
**Description**
According to the [Fiori UI text guidelines](https://experience.sap.com/internal/fiori-design-web/ui-text-guidelines-for-sap-fiori/#user-input), placeholders should not repeat information provided in the label. For example: 
placeholder repeat
Since label says `Name` there is no need to repeat that in the placeholder.

**Expected result**
There should be no placeholders if the text in it repeats the the text from the label.

**Actual result**
Placeholders texts sometimes repeat the labels' texts.

grego952:
**Description**

- remove unnecessary texts from placeholders
- small fixes

**Related issue(s)**
Resolves #978


core-ui/src/components/Predefined/Create/Gateways/Forms/HostsForm.js
```diff
@@ -14,6 +14,7 @@ export const HostsForm = ({ server = {}, servers, setServers }) => {
     <TextArrayInput
       advanced
       required
+      ariaLabel="Host"
```
qbalukom, line 17: Shouldn't Aria label be taken from i18n as well? this is formally meant for accessibility purposes, after all

qbalukom: (this applies to all cases of ariaLabel appearing in the PR)

core/src/i18n/en.yaml
```diff
@@ -550,7 +547,7 @@ common:
     hide: Hide Read-Only Fields
     k8s-name-input: "Name must consist of lowercase alphanumeric characters, can contain '-' (e.g.: 'my-name1')."
     key: Key must start and end with an alphanumeric character, can contain '-', '_' or '.'.
-    key-value: key=value, must start and end with an alphanumeric character, can contain '-', '_' or '.'.
+    key-value: key-value, must start and end with an alphanumeric character, can contain '-', '_' or '.'.
```
qbalukom, line 550: I think it would be better to properly form it as "Both key and value must start", or something similar (also, it seems it's not entirely accurate - as KeyValueField is used both for labels and annotations... and while it's true for labels, it's not true for annotations... ugh)

core/src/i18n/en.yaml
```diff
@@ -592,7 +587,7 @@ components:
     value: Value
   label-selector-input:
     click-to-delete: Click to delete
-    enter-key-value: Enter {{type}} key=value
+    enter-key-value: Enter key-value labels
```
qbalukom, line 590: Not sure if we use it anymore (might require some component purge to happen - i'll investigate it independently)

qbalukom: I mean, we DO use it, but in a component that might be not necessary anymore :)

qbalukom (changes requested):

core/src/i18n/en.yaml
grego952: Since it is available in the Dashboard I'd suggest to keep it :)

core/src/i18n/en.yaml
grego952: Ok, I will rephrase it :) Any suggestions how to make it more accurate?

core-ui/src/components/Predefined/Create/Gateways/Forms/PortsForm.js
```diff
@@ -62,17 +59,15 @@ export const PortsForm = ({ server = {}, servers, setServers }) => {
...
 
       <ResourceForm.FormField
         tooltipContent={t('gateways.create-modal.tooltips.target-port')}
         label={t('gateways.create-modal.advanced.port.target-port')}
         propertyPath="$.port.targetPort"
-        placeholder={t(
-          'gateways.create-modal.advanced.placeholders.port.target-port',
-        )}
         input={Inputs.Port}
+        ariaLabel="Target port"
```
qbalukom, line 70: one last untranslated aria-label

qbalukom (changes requested):

qbalukom (dismissed):

core/src/i18n/en.yaml
```diff
@@ -1659,7 +1655,7 @@ replica-sets:
       replicas: Replicas
     tooltips:
       min-ready-seconds: Minimum number of seconds for which a newly created Pod should be ready without any of its containers crashing in order to be considered available.
-      replicas: Replicas is the number of desired replicas.
+      replicas: Number of desired replicas.
```
klaudiagrz, line 1658:
```suggestion
      replicas: Number of desired replicas
```

klaudiagrz (approved):

qbalukom (approved):

PR #1012, opened 2022-02-24
implement grafana logs link

qbalukom:


shared/components/ResourceDetails/ResourceDetails.js
```diff
@@ -164,6 +168,20 @@ function Resource({
 
   const protectedResource = isProtected(resource);
 
+  const grafanaLogsAction = () => {
```
dariadomagala, line 171: why won't we just add it to the `headerActions` instead? I don't think it will be used in so many places that we should make it generic, let's not make this file any more chaotic.

core-ui/src/components/Predefined/Details/Pod/Pod.details.js
```diff
@@ -126,6 +126,7 @@ export const PodsDetails = ({ DefaultRenderer, ...otherParams }) => {
         Events,
       ]}
       customColumns={customColumns}
+      grafanaQuery={`{pod="${otherParams.resourceName}"}`}
```
dariadomagala, line 129: we should also add a namespace to the query, you can have pod with the same name in multiple namespaces

shared/components/LogsLink/LogsLink.scss
```diff
@@ -0,0 +1,3 @@
+.logs-link {
+  align-self: center;
+}
```
dariadomagala, line 3: we should also add a question mark on hover, as everywhere else where we have tooltips

dariadomagala (changes requested):

dariadomagala (approved):

dariadomagala (dismissed):

dariadomagala (approved):

grego952 (approved):

PR #1019, opened 2022-02-27
Introduce watch to the resource lists

### LINKED ISSUE #985: SPIKE:  ?watch=true parameter in api calls instead of polling
Research the use of ?watch=true parameter in api calls instead of polling.

Acceptance Criteria

- [x] Evaluate if we can use ?watch
- [x] Evaluate and document the Performance Gains / Losses and possible scaling issues when we use ?watch

### LINKED ISSUE #986: SPIKE: Have a Visual Way to display the connected resources
Acceptance Criteria
- [x] We research if there is a universal way to detect related / connected resources
- [x] We evaluate if this can be represented in a Visual Way - see https://reactflow.dev/ as an option
- [x] We identify and document some of the shortcomings

Relevant Projects for inspiration:

- https://www.npmjs.com/package/k8s-dot-graph
- https://github.com/benc-uk/kubeview
- https://github.com/ahmetb/kubectl-tree
- https://github.com/viralpoetry/kubesurveyor
- https://github.com/steveteuber/kubectl-graph
- https://github.com/alcideio/rbac-tool

Examples:
image
image
image
image
image
image

wchrapka:
**Description**

Introduce watch to the resource lists

**Related issue(s)**
Closes #985


backend/index.js
```diff
@@ -18,7 +18,23 @@ try {
...
-if (gzipEnabled) app.use(compression());
+if (gzipEnabled)
+  app.use(
+    compression({
+      filter: (req, res) => {
+        //TODO change it to a custom header, pass the header from the front
+        if (
+          req.originalUrl.includes('watch=') ||
+          req.originalUrl.includes('follow=')
+        ) {
```
Wawrzyn321, line 29: Since we're only proxing k8s requests, I'd say we don't even need a custom header - we can take it from query params instead of an entire URL.

PR #1021, opened 2022-02-28
Fix repeating status names in tooltips

### LINKED ISSUE #981: In many cases information in tooltips repeats the status name
**Description**
In many cases the information in the tooltip is preceded by the repetition of the status or includes only the repetition of the status. For example:
Zrzut ekranu 2022-02-21 o 10 58 55
Zrzut ekranu 2022-02-21 o 09 31 38
Zrzut ekranu 2022-02-21 o 09 32 17
Zrzut ekranu 2022-02-21 o 09 38 22

 

**Expected result**
Tooltips should not include the repetition of the already visible status.

**Actual result**
The tooltips include repetition of the status.

Lyczeq:
**Description**

- Fix repeating status names in tooltips
- Fix status bug on Brokers list

**Related issue(s)**
Fixes #981


shared/components/StatusBadge/StatusBadge.js
```diff
@@ -125,8 +125,7 @@ export const StatusBadge = ({
 
   if (additionalContent) {
     // Remove the dot at the end of the sentence if we add a colon afterwards
-    if (content?.endsWith('.')) content = content?.slice(0, -1);
-    content = `${content}: ${additionalContent}`;
+    content = additionalContent;
```
qbalukom, line 128: are we sure `content` will always be the same as `badgeContent`? maybe checking `if (content === badgeContent)` would work better?

also after the changes, the comment in line 127 doesn't make sense

qbalukom (changes requested):

qbalukom (approved):

PR #1039, opened 2022-03-02
Refactor of pod and containers statuses

### LINKED ISSUE #1024: Improve Pod statuses
**Description**

In case of Pods, we display the general status as a convoluted result of `pod.status.containerStatuses`. It's not required - we could do just fine with `pod.status.phase`
image.

What's more, it would be nice to show similar statuses on Pod's container list - for now on we show nothing:
image

Here we could actually use `pod.status.containerStatuses` and `pod.status.initContainerStatuses`.

**Reasons**

Simplify the "main" status, add more info for containers.

mrCherry97:
**Description**
Pod statuses were not loaded properly and we didn't display containers statuses.

- Changed pod statuses
- Added containers statuses
- ...

**Related issue(s)**
Closes #1024


tests/tests/test-replica-sets.spec.js
```diff
@@ -56,7 +56,9 @@ context('Test Replica Sets', () => {
       .contains(`${REPLICA_SET_NAME}-`)
       .click();
 
-    cy.getIframeBody().contains(`Name${REPLICA_SET_NAME}`);
+    cy.getIframeBody().contains(REPLICA_SET_NAME);
```
dariadomagala, line 59: Why those changes?

core-ui/src/components/Predefined/Details/Pod/ContainerStatus.js
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <div>
+      <StatusBadge
+        i18n={i18n}
+        resourceKind="containers"
+        additionalContent={message}
+        type={badgeType(state.reason || 'Running')}
+      >
+        {state.reason || 'Running'}
```
dariadomagala, line 41: We don't show tooltips for running/completed statuses.
We have labels like 'imagepullbackoff' even tho we agreed at some point that those should be simplified to just 'error'.

We should use the same approach as we have for pods statuses. Statuses for pods are calculated based on container statuses, so I believe they should display similar messages.

tests/tests/test-replica-sets.spec.js
mrCherry97: Because for now, we aren't displaying names in the table, only in the header of the container.

core/src/i18n/en.yaml
```diff
@@ -628,6 +628,54 @@ config-maps:
...
+    registryunavailable: Error
+    runcontainererror: Error
+    running: Running
+    terminated: Terminated
+    terminating: Terminating
+  tooltips:
+    completed: Container completed
+    containercreating: Creating container
+    containersnotinitialized: Containers not initialized
+    containersnotready: Containers not ready
```
dariadomagala, line 659:
```suggestion
    containersnotready: Container not ready
```

core/src/i18n/en.yaml
```diff
@@ -628,6 +628,54 @@ config-maps:
...
+    createcontainererror: Create container error
+    dockerdaemonnotready: Docker daemon not ready
+    errimageneverpull: Image has never been pulled
+    errimagepull: Image pull error
+    error: Container failed to start
+    imageinspecterror: Image inspect error
+    imagepullbackoff: Could not pull a container image
+    invalidimagename: Invalid image name
+    networkpluginnotready: Network plugin not ready
+    no-restarts: None of the containers has restarted
```
dariadomagala, line 671:
```suggestion
    no-restarts: Container has not restarted
```

grego952 (approved):

dariadomagala (approved):

PR #1041, opened 2022-03-02
Add min value to limit ranges

### LINKED ISSUE #1029: Min limit is missing from Limit Ranges
**Description**

**Expected result**

We should see the Min Limit Range

**Actual result**

**Steps to reproduce**

Add the next limit range and notice the UI

```
metadata:
  name: default-limit-range
  namespace: default
spec:
  limits:
    - type: Container
      max:
        cpu: 500m
        memory: 512Mi
      min:
        cpu: 50m
        memory: 50Mi
      default:
        cpu: 100m
        memory: 100Mi
      defaultRequest:
        cpu: 50m
        memory: 50Mi
```        

**Troubleshooting**

image
Min

mrCherry97:
**Description**
Users should see the Min Limit Range.

- Added `min` value to the template
- Added `min` field in the list
- ...

**Related issue(s)**
Closes #1029

### DIFF
--- a/core-ui/src/components/Predefined/Create/LimitRanges/templates.js
+++ b/core-ui/src/components/Predefined/Create/LimitRanges/templates.js
@@ -1,6 +1,7 @@
 export function createLimitRangeTemplate({
   max = '1100Mi',
-  defaultVal = '511Mi',
+  min = '32Mi',
+  defaultVal = '512Mi',
   defaultRequest = '32Mi',
   name = '',
   namespaceName = '',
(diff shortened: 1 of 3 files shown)


core-ui/src/components/Predefined/Create/LimitRanges/templates.js
```diff
@@ -1,6 +1,7 @@
 export function createLimitRangeTemplate({
   max = '1100Mi',
-  defaultVal = '511Mi',
+  min = '32Mi',
```
akucharska, line 3: Maybe we should modify tests to check if created limit range uses these defaults values? For now, we only check the name.

mrCherry97: I didn't check tests, I will look at it later.

akucharska (dismissed):

grego952 (approved):

akucharska (approved):

PR #1069, opened 2022-03-09
Repair uppercase name for Role Bindigs

mrCherry97:
**Description**
Role Bindings could have uppercase in the name.

- Added if for `toLocaleLowerCase`, because we don't want lower case names.
- ...
- ...

**Related issue(s)**
Closes [Issue](https://github.tools.sap/kyma/backlog/issues/2410)


core-ui/src/shared/helpers/getResourceUrl.js
```diff
@@ -3,8 +3,20 @@ export function getResourceUrl() {
   const fullResourceApiPath = queryParams.get('fullResourceApiPath');
   const resourceApiPath = queryParams.get('resourceApiPath');
 
+  // resource names can't be lowercased - Role Bindings there could be name with upper case
+  const tokens = window.location.pathname.split('/');
+  let tokensLength = tokens.length;
+
+  if (tokensLength === 3) tokensLength = 2;
+  else if (tokensLength === 5) tokensLength = 4;
```
dariadomagala, line 11: what about number 7? 🙃 
we have a longer path for example here: `namespaces/NamespaceName/customresourcedefinitions/details/CRDName/v1alpha1/CRName`
Also, we don't lowercase just the last resourceName in this case, shouldn't we apply it to all in the path (so, in the example above - NamespaceName, CRDName and CRName)?

dariadomagala (commented):
`Rules` instead of `API Rules` in the api rules details

grego952 (approved):

Issue #982, opened 2022-02-21
Introduce Lighthouse Regression Testing in our Pipeline


valentinvieriu:
**Description**

We need to protect our App against performance regressions.

Acceptance Criteria

- [x] We reject all PR that harm in a negative way ( deviation of < 3% ) the lighthouse score from the current status quo

**Reasons**

**Attachments**

[headless-chrome](https://github.com/GoogleChrome/lighthouse/blob/master/docs/headless-chrome.md)

Sawthis:
I think we should try to introduce https://github.com/GoogleChrome/lighthouse-ci especially if we have problems with integration with Cypress. @Wawrzyn321 @qbalukom what do you think?

qbalukom:
If there's a way of reliably logging into a cluster... then sure

Wawrzyn321:
Then let's try this https://www.npmjs.com/package/playwright-lighthouse

Wawrzyn321:
There's no additional iframe, so it'll be an advantage. We can add our code to pick the kubeconfig and then run Lighthouse audit.

PR #1072, opened 2022-03-10
Lighthouse CI

Wawrzyn321:
**Description**

- Remember to copy a kubeconfig to `fixtures/kubeconfig.yaml`

**Related issue(s)**

Connected #982
[test-infra](https://github.com/kyma-project/test-infra/pull/5101)


lighthouse/playwright.config.js
```diff
@@ -0,0 +1,106 @@
...
+  projects: [
+    {
+      name: 'chromium',
+      use: {
+        ...devices['Desktop Chrome'],
+      },
+    },
+
+    // {
+    //   name: 'firefox',
```
dariadomagala, line 54: can we also add scans for firefox, or does it need additional configurations?

lighthouse/lighthouse.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+  await expect(
+    page.frameLocator('iframe').locator('text=Cluster Overview'),
+  ).toBeVisible();
+
+  await playAudit({
+    page,
+    port: 9222,
+    thresholds: {
+      performance: 3,
```
dariadomagala, line 59: Can we set higher values here? Checking if it's higher than 3 is not valuable at all, and on my machine, the performance value is only 5, which is also pretty far from what the lighthouse shows normally.

lighthouse/lighthouse.spec.js
```diff
@@ -0,0 +1,68 @@
...
+test('Busola Lighthouse audit', async () => {
+  const context = await chromium.launchPersistentContext(tmpdir(), {
+    args: ['--remote-debugging-port=9222'],
+  });
+  const page = await context.newPage();
+
+  await page.goto('http://localhost:8080');
+
+  // on dev chromium may remember previous state - get back to clusters list
+  if (!page.url().endsWith('/clusters')) {
```
dariadomagala, line 15: it doesn't work on a first run. Address is just `localhost:8080`, and only after a while it gets redirected to `.../clusters`.

Wawrzyn321: Woah, I forgot that /clusters actually **doesn't** redirect you to current cluster... Fixed!

Wawrzyn321: Or actually... no, it does redirect to current cluster. I don't know why.

lighthouse/playwright.config.js
Wawrzyn321: Nope, Lighthouse is Chrome(chromium?)-specific.

Wawrzyn321: I mean, there's a Lighthouse extension for Firefox

dariadomagala (approved):

grego952 (approved):

PR #1087, opened 2022-03-15
Fix lowercase in the 'resource not found page'

### LINKED ISSUE #1090: Wrong resource type in the 'resource not found' page
**Description**

With the latest changes, we introduced a bug where we don't show a proper message when the user is on a 'not found' page.
Screen Shot 2022-03-16 at 07 29 54
We should show 'Cluster Role Bindings' instead of 'Clusterrolebindings'

**Expected result**

Proper resource type should be shown.

**Actual result**

Resource type is hard to read.

**Steps to reproduce**

**Troubleshooting**

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Resolves #1090

### DIFF
--- a/core-ui/src/components/Predefined/Details/RoleBindings/RoleBindings.details.js
+++ b/core-ui/src/components/Predefined/Details/RoleBindings/RoleBindings.details.js
@@ -25,3 +25,21 @@ export function ClusterRoleBindingsDetails(props) {
     />
   );
 }
+
+function GenericRoleBindingDetails({ DefaultRenderer, ...otherParams }) {
+  const { t } = useTranslation();
+
+  const customColumns = [
+    {
+      header: t('role-bindings.headers.role-ref'),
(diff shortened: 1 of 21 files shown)


core-ui/src/components/Predefined/Details/RoleBindings/RoleBindings.details.js
```diff
@@ -23,6 +23,7 @@ function GenericRoleBindingDetails({ DefaultRenderer, ...otherParams }) {
   ];
   return (
     <DefaultRenderer
+      resourceTitle={t('cluster-role-bindings.title')}
```
qbalukom, line 26: this displays "Cluster Role Bindings" even for regular Role Bindings

qbalukom (changes requested):

qbalukom (approved):

PR #1091, opened 2022-03-16
Graph colors

qbalukom:


shared/components/ResourceGraph/buildGraph/helpers.js
```diff
@@ -9,15 +9,20 @@ export function wrap(str) {
 export function makeNode(resource) {
   const { kind, metadata } = resource;
   const { name, uid } = metadata;
+  console.log('makeNode', resource, metadata);
```
Wawrzyn321, line 12:
```suggestion
```

Wawrzyn321 (changes requested):
I feel like it looks better with that bolder text, but graphviz doesn't get it and sometimes it overflows:
image

Why don't we add `margin="0.2,0.2` param to each node?

Wawrzyn321 (approved):

PR #1107, opened 2022-03-18
Added inspecting list function and added inspecting with deleting to tests

### LINKED ISSUE #969: [TESTS] Move Inspecting created resource on list to separated function
**Description**
We have to clean up tests more and we have to inspect created resource on the list in every test. We have to also investigate  if `search-input` needs `{ force: true }`, Wojtek mention that `open-search` is working without it.

**Reasons**
In every test we have to inspect resource on list so we can move it to separated function:
```
it('Inspect list and delete resource', () => {
    cy.getIframeBody()
      .contains(RESOURCE)
      .click();

    cy.getIframeBody()
      .find('[role="search"] [aria-label="search-input"]')
      .type(RESOURCE_NAME, { force: true });

    cy.getIframeBody()
      .contains('tbody tr td a', RESOURCE_NAME)
      .click({ force: true });

    cy.deleteInDetails();
  });
```

mrCherry97:
**Description**
We have to more cover code with tests, so we have to add an inspecting list for every resource.
- Added `inspectList` function,
- Moved PVC tests to namespace zone and trigger it,
- Added deletion for all cluster resources and to single namespace resource.

**Related issue(s)**
Closes #969


mrCherry97:
End on `in-cluster-eventing`

tests/tests/namespace/test-config-maps.spec.js
```diff
@@ -108,4 +110,18 @@ context('Test Config Maps', () => {
       .contains('.fd-layout-panel', ENTRY_KEY2)
       .contains(ENTRY_VALUE2);
   });
+
+  it('Delete Config Maps', () => {
```
dariadomagala, line 114:
```suggestion
  it('Delete Config Map', () => {
```

tests/tests/namespace/test-config-maps.spec.js
```diff
@@ -108,4 +110,18 @@ context('Test Config Maps', () => {
...
+    cy.getLeftNav()
+      .contains('Config Maps')
+      .click();
+
+    deleteFromGenericList(CLONE_NAME);
+
+    cy.getLeftNav()
```
dariadomagala, line 121: Why do we click it twice? one should be enough

tests/tests/namespace/test-cron-jobs.spec.js
```diff
@@ -168,4 +169,12 @@ context('Test Cron Jobs', () => {
       .contains('0 * * * *')
       .should('be.visible');
   });
+
+  it('Inspect list and delete', () => {
+    cy.getIframeBody()
```
dariadomagala, line 174: Cannot we use the `navigateTo` function?

tests/tests/namespace/test-config-maps.spec.js
```diff
@@ -108,4 +110,18 @@ context('Test Config Maps', () => {
       .contains('.fd-layout-panel', ENTRY_KEY2)
       .contains(ENTRY_VALUE2);
   });
+
+  it('Delete Config Maps', () => {
+    cy.getLeftNav()
+      .contains('Config Maps')
```
dariadomagala, line 116: Cannot we use the `navigateTo` function?

tests/tests/namespace/test-api-rules.spec.js
```diff
@@ -145,4 +146,9 @@ context('Test API Rules in the Function details view', () => {
       .contains('Allow')
       .should('exist');
   });
+
+  it('Inspect list and delete', () => {
+    cy.navigateTo('Discovery and Network', 'API Rules');
+    deleteFromGenericList(API_RULE_NAME);
```
dariadomagala, line 152: I don't think it's required to delete it, since it will be deleted together with a function (I believe)

tests/tests/namespace/test-config-maps.spec.js
```diff
@@ -108,4 +110,18 @@ context('Test Config Maps', () => {
...
+      .click();
+
+    deleteFromGenericList(CLONE_NAME);
+
+    cy.getLeftNav()
+      .contains('Config Maps')
+      .click();
+
+    deleteFromGenericList(CONFIG_MAP_NAME);
```
dariadomagala, line 125: I question if we should delete all namespace-wide resources. It would be nice to test it for one of them to check the mechanics, but those resources will be deleted together with a namespace, so there's no need to run more tests that are not very useful.

mrCherry97:
Passed, but for sure:
/test pre-busola-integration-test-k3s-namespace

mrCherry97:
Passed, but for sure:
/test pre-busola-integration-test-k3s-namespace

dariadomagala (dismissed):

dariadomagala (approved):

PR #1123, opened 2022-03-22
All api groups

qbalukom:


dariadomagala (changes requested):
You cannot type anything in the input now, it's automatically added as a nwe api gruop etc

core-ui/src/components/Predefined/Create/Roles/helpers.js
```diff
@@ -1,3 +1,21 @@
...
+export const EMPTY_API_GROUP_KEY = 'core-api-group';
+export const extractApiGroup = groupVersion => {
+  // handle core ('') group
+  if (groupVersion === 'v1') return '';
+  const [apiGroup] = groupVersion.split('/');
+  return apiGroup;
+};
+
+export const getApiGroupInputOptions = groupVersions =>
+  unique(groupVersions.map(extractApiGroup)).map(g =>
```
dariadomagala, line 13: can we please add either `?` before map, or `|| []` at the end? I think it's the most common cause when it comes to crashing UIs in our code 😄 

(please also apply in other places, it doesn't make sense to comment it all - also for 'some', 'filter' etc)

core-ui/src/shared/ResourceForm/fields/ComboboxArrayInput.js
```diff
@@ -41,38 +43,59 @@ export function ComboboxArrayInput({
...
+              _ref={ref}
+              selectedKey={value}
+              typedValue={value || ''}
+              selectionType="manual"
+              setValue={setValue}
+              options={filteredOptions()}
+              onKeyDown={focus}
+              onBlur={onBlur}
+              onSelectionChange={(_, selected) => {
+                console.log('onSelectionChange', selected);
```
dariadomagala, line 83: nope

dariadomagala (changes requested):
Now you can type in some values, but those are not being added to the list, nothing happens. I cannot add a verb that's not on the list and it was a reason why we've chosen this component.

core-ui/src/shared/ResourceForm/fields/ComboboxArrayInput.js
qbalukom: oops

dariadomagala (approved):

grego952 (approved):

Issue #1135, opened 2022-03-22
Prometheus Graphs and Visual Resources are missing from the documentation


valentinvieriu:
**Description**
We have a documentation for our features, and seems the latest ones are missing from the documentation
https://github.com/kyma-project/busola/blob/main/docs/features.md

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

dariadomagala:
Added in https://github.com/kyma-project/busola/pull/1139

PR #1139, opened 2022-03-23
Add default prometheus address

### LINKED ISSUE #1128: Prometheus Graphs component should be configurable via feature toggle
**Description**

The API proxy endpoint for Prometheus Graphs is not configurable at the Feature Toggle level

**Expected result**

We should be able to configure and change the  API proxy endpoint for Prometheus Graphs

**Actual result**

The API proxy endpoint for Prometheus Graphs is not configurable at the Feature Toggle level

**Steps to reproduce**

**Troubleshooting**

dariadomagala:
**Description**

test it also on `nkyma` where the path is `api/v1/namespaces/kyma-system/services/monitoring-prometheus:http-web/proxy/api/v1`

- ...
- ...
- ...

**Related issue(s)**

Resolves #1128 #1135


core-ui/src/shared/components/StatsGraph/StatsPanel.js
```diff
@@ -136,6 +144,10 @@ export function StatsPanel({ type, ...props }) {
     return '';
   }
 
+  let path = features.PROMETHEUS?.config?.path;
```
qbalukom, line 147: wouldn't it be better to read the path inside the `usePrometheus` hook? It's read from microfrontend context anyway - passing the path through these components seems to only add unnecessary complexity

core/src/assets/defaultConfig.json
```diff
@@ -61,7 +61,10 @@
         }
       },
       "PROMETHEUS": {
-        "isEnabled": true
+        "isEnabled": true,
+        "config": {
+          "path": "api/v1/namespaces/kyma-system/services/monitoring-prometheus:web/proxy/api/v1"
```
qbalukom, line 66: do we need actually the whole path?
both the prefix (`api/v1/namespaces/kyma-system/services/`) and the suffix (`/proxy/api/v1`) are very unlikely to change - as they're part of the kubernetes and prometheus implementations respectively.
The only two parts that can easily change on a per-cluster basis are service name and port name. I feel that would be enough. Either as a single value (`monitoring-prometheus:web`) or two, one for service name (`monitoring-prometheus`) one for port (`web`)

qbalukom (changes requested):

docs/features.md
```diff
@@ -342,19 +342,38 @@ The **match** keys and **messageSrc** must use the format described in the [`jso
...
+  ```bash
+  "PROMETHEUS": {
+    "isEnabled": true,
+    "config": {
+      "path": "api/v1/namespaces/kyma-system/services/monitoring-prometheus:web/proxy/api/v1"
+    }
+  },
+  ```
+
+- **VISUAL_RESOURCES** – configures if the resource graphs should be rendered at resources details views.
```
grego952, line 373:
```suggestion
- **VISUAL_RESOURCES** – determines if the resource graphs should be rendered at a resource details view.
```

grego952 (approved):

qbalukom (approved):

PR #1151, opened 2022-03-25
Add stats for deployments

akucharska:
**Description**

- add CPU, Memory, Network graph for deployments
- show graph values divided by containers
- add values divided by containers also to pods

**Related issue(s)**


dariadomagala (commented):
the tooltips are confusing. if the graph has multiple entries, the one that is on the bottom of the graph, is mentioned on the top in the tooltip.

dariadomagala (approved):

Issue #1085, opened 2022-03-15
Hide the PVC and SC links if they don't exist


dariadomagala:
**Description**

When the user is on a Persistent Volume list, they see a link to the bound Persistent Volume Claim. When the user is on the PV details, they see also a link to the bound Storage Class. 
Although, they also see it even if PVC or SC do not exist. And after clicking on a link, they get redirected to the page that says `Such Persistent Volume Claim / Storage Class doesn't exist`. 
We should hide links that are not valid.

<img width="637" alt="Screen Shot 2022-03-15 at 13 19 02" src="https://user-images.githubusercontent.com/40663133/158378589-43a44906-bd4b-4e91-a3c8-84832941a13a.png">

**Expected result**

Hide invalid links.

**Actual result**

The user sees and can click invalid links.

**Steps to reproduce**

Create a PV with `storageClassName` and `claimRef` in spec that point to non-existing resources.

**Troubleshooting**

dariadomagala:
Duplicate of https://github.com/kyma-project/busola/issues/1156

PR #1157, opened 2022-03-25
Added validation if storage class and pvc exist

### LINKED ISSUE #1156: Deep link in Persistent Volume to Storage Class provide to no existing resource
**Description**
When we create Persisten Volume with relation to a Storage Class that doesn't exist there couldn't be a deep link.

**Expected result**
When storage class doesn't exist there couldn't be the deep link.

**Actual result**
image
After clicking on the link:
image

**Steps to reproduce**
1. Go to Persistent Volume
2. Create Persistent Volume -> provide the name of storage class that doesn't exist
3. Check the link

mrCherry97:
**Description**
When Storage Class or PErsistentVolumeClaim doesn't exist UI can't show links to the non-existing resources.
- Added validation for existing SC,
- Added validation for existing PVC,
- ...

**Related issue(s)**
Closes #1156


core-ui/src/components/Predefined/List/PersistentVolumes.list.js
```diff
@@ -14,23 +15,35 @@ import { PersistentVolumesCreate } from '../Create/PersistentVolumes/PersistentV
 const PersistentVolumesList = props => {
   const { t } = useTranslation();
 
+  const { data: storageClasses } = useGetList()(
+    '/apis/storage.k8s.io/v1/storageclasses',
+  );
+
+  const { data: persistentVolumeClaims } = useGetList()(
```
dariadomagala, line 22: We might want to double-check it, but AFAIK, if the bound PVC doesn't exist anymore, the PV has a `released` status. If that's true, then we don't need to make another call to get a list of PVC, just check the status (SC are not the same story, unfortunately).

dariadomagala:
Closes also #1085

dariadomagala (approved):

dariadomagala (approved):

PR #1167, opened 2022-03-29
fix namespaced user navigation

### LINKED ISSUE #1109: Some Cluster Links don't redirect to namespace
**Description**

Links on the top menu and from the command pallete menu don't redirect to namespace in the case of namespace only kubeconfig file

**Expected result**

The navigation behaviour should be the same as clicking on the Cluster List. Those redirect to namespace

**Actual result**

**Steps to reproduce**

**Troubleshooting**

image
image

qbalukom:
- fix cluster navigation in top bar for namespaced users
- fix cluster navigation from command palette for namespaced users
- display user namespace in top bar namespace dropdown for namespaced users

fixes #1109


dariadomagala (changes requested):
If I have two clusters with different permissions, and I switch from the one with limited permissions to the admin one, I still get for example 
```User "system:serviceaccount:qqq:ns-user" cannot get resource "services/proxy" in API group "" in the namespace "kyma-system"```
even though after reload, everything works ok.

dariadomagala (approved):
😢

dariadomagala (approved):
/woof

PR #1175, opened 2022-03-30
Change base image to unprivileged nginx

KaiReichart:
Currently, the web container is running as root. With usage of the nginx unprivileged image, the container can run with an arbitrary user-id and does not need to be run as root anymore.

- Change base image to `nginx-unprivileged`

### DIFF
--- a/resources/backend/deployment.yaml
+++ b/resources/backend/deployment.yaml
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
         - name: backend
-          image: eu.gcr.io/kyma-project/busola-backend:PR-1145
+          image: eu.gcr.io/kyma-project/busola-backend:PR-1175
           imagePullPolicy: Always
           resources:
             limits:
--- a/resources/backend/deployment.yaml
+++ b/resources/backend/deployment.yaml
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
         - name: backend
-          image: eu.gcr.io/kyma-project/busola-backend:PR-1175
+          image: eu.gcr.io/kyma-project/busola-backend:PR-1145
           imagePullPolicy: Always
           resources:
             limits:
--- a/resources/web/deployment.yaml
+++ b/resources/web/deployment.yaml
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
         - name: busola
-          image: eu.gcr.io/kyma-project/busola-web:PR-1175
+          image: eu.gcr.io/kyma-project/busola-web:PR-1174
           imagePullPolicy: Always
           resources:
             requests:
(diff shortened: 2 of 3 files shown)


KaiReichart:
@mrCherry97 It looks like the docker image is not being built, am I missing something here?

resources/backend/deployment.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
         - name: backend
-          image: eu.gcr.io/kyma-project/busola-backend:PR-1145
```
mrCherry97, line 23: You have to bring back `PR-1145`

resources/web/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
         - name: busola
-          image: eu.gcr.io/kyma-project/busola-web:PR-1174
```
mrCherry97, line 19: Yout have to bring back `PR-1174`

mrCherry97 (changes requested):
Hi @KaiReichart !
When you are editing just Dockerfile you don't have to bu,p deployments, let's bring PR numbers back and we can check the tests. Thanks!

resources/backend/deployment.yaml
KaiReichart: done thx

resources/web/deployment.yaml
KaiReichart: done thx

KaiReichart:
@mrCherry97 changed it back, thx. 
However, I don't really understand how the tests will bring any value if the Dockerfile is not being used for the tests... 

Maybe the images should also be built and tested when the Dockerfile changes?
Because now the tests don't really test the changes of the PR, right?

mrCherry97:
@KaiReichart 
When you change Dockerfile we need to test `pre-busola-web` because we are using docker images for testing and development, so we need to know that images are building and deploying well.

KaiReichart:
@mrCherry97 can this be merged now? Or is it missing something?

Wawrzyn321:
@KaiReichart yeah, it's missing something - my review. On it right now!

Wawrzyn321 (approved):

KaiReichart:
Thanks @Wawrzyn321 ! Now its saying `Requested changes must be addressed to merge this pull request.`  
How do we resolve this? I had incorporated the requested changes already, but it seems like GitHub didn't notice...

mrCherry97 (approved):

Wawrzyn321:
It looks like @mrCherry97 needs to approve, as he requested some changes.

mrCherry97:
Hi @KaiReichart and @Wawrzyn321 
Changes approved 😄

KaiReichart:
nice, thanks guys! See you in the next PR ;)

PR #1216, opened 2022-04-05
Resource modules

qbalukom:


core-ui/src/resources/APIRules/APIRuleList.js
```diff
@@ -1,15 +1,17 @@
...
 import ApiRuleStatus from 'components/ApiRules/ApiRuleStatus/ApiRuleStatus';
-import { useTranslation, Trans } from 'react-i18next';
 import { ResourcesList } from 'shared/components/ResourcesList/ResourcesList';
 import { Link } from 'shared/components/Link/Link';
-import { APIRulesCreate } from '../Create/ApiRules/ApiRules.create';
 
-const ApiRulesList = props => {
+import { APIRuleCreate } from './APIRuleCreate';
+
+export function APIRuleList(props) {
```
wchrapka, line 14: Why are there doubled exports in every list and detail file?

core-ui/src/resources/other.js
```diff
@@ -17,7 +16,7 @@ const other = (
   <>
     {customResourceDetails}
     {customResourceListOfType}
-    {customResourceDefinitions}
+    {/*TODO customResourceDefinitions*/}
```
wchrapka: todo?

wchrapka: FYI
main chunk size (gzipped): 387.15KB
image

PR #1241, opened 2022-04-12
Improve caching for static files

### LINKED ISSUE #1165: Improve static files caching
**Description**

We need to cache the static resources as long as possible. We still have static resources showing up in Lighthouse that are 1-7.
We need to make sure that all resources use hashes, and they follow the Lighthouse recommendations https://web.dev/uses-long-cache-ttl/?utm_source=lighthouse&utm_medium=devtools

- Reduce caching for index.html to less than 1h ( maybe use etags ?) - purpose is to make sure when we deploy when we make a push , the user gets the latest version

**Reasons**

**Attachments**

image

Lyczeq:
**Description**

- Bundle `index.css` to the css file with contenthash
- Adjust time period for static files we have
- Add `reset.css` and other css files to the main bundle to not send unnecessary requests for them
- Remove needless manifest and favicon

**Related issue(s)**
Closes #1165


nginx/nginx.conf
```diff
@@ -53,22 +53,22 @@ http {
...
         location /assets {
             include /etc/nginx/core.conf;
-            expires 7d;
+            expires 1y;
         }
 
         location /core-ui/themes {
             include /etc/nginx/core-ui.conf;
-            expires 1d;
+            expires 30d;
```
Wawrzyn321, line 71:
```suggestion
            expires 1m;
```

Wawrzyn321 (changes requested):
Looks good, one tiny comment.

Wawrzyn321 (approved):

PR #1266, opened 2022-04-20
Load graphs when 20% of the graph container is visible

### LINKED ISSUE #1256: Resource graphs don't always load
**Description**
The resource graph doesn't load if its whole container is not visible. On a regular screen, the user needs to scroll all the way down to load the graph (even 1px less will cause it not to load). On a smaller screen (or when the dev tools are open), the graph never loads.
Screen Shot 2022-04-19 at 10 08 21

**Expected result**

Graphs should start to load when the container is at least partially visible. It should always be able to load if it's not hidden.

**Actual result**

Graphs don't always load

**Steps to reproduce**

**Troubleshooting**

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Fixes #1256

### DIFF
--- a/core-ui/src/shared/hooks/useIntersectionObserver.js
+++ b/core-ui/src/shared/hooks/useIntersectionObserver.js
@@ -3,7 +3,7 @@ import { useEffect, useMemo, useState } from 'react';
 const defaultOptions = {
   root: null,
   rootMargin: '0px',
-  threshold: 1.0,
+  threshold: 0.5,
 };
 
 export const useIntersectionObserver = (target, options = {}) => {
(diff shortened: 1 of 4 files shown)


core-ui/src/shared/hooks/useIntersectionObserver.js
```diff
@@ -3,7 +3,7 @@ import { useEffect, useMemo, useState } from 'react';
 const defaultOptions = {
   root: null,
   rootMargin: '0px',
-  threshold: 1.0,
+  threshold: 0.5,
```
mrCherry97, line 6: Investigate `0.2`, I think it could be better value.

mrCherry97 (approved):

akucharska (approved):

PR #1270, opened 2022-04-20
Introduce Volume Usage Graph for PVCs

### LINKED ISSUE #1259: Introduce Prometheus Graphs for Persistent Volumes Claims
As [Sebastian](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#sebastian---tech-lead) and [Mathew](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#matthew---cloud-native-software-engineer) I would like to be able to see Prometheus Metrics regarding Persistent Volumes Claims so that I can make informed decisions about possible space allocation problems.
**Acceptance Criteria**

- [x] Follow Project DOD
- [x] We use prometheus as a source for the metrics
- [x] The feature is enabled when prometheus is available in the cluster, and replaced by the current graphs when prometheus is disabled.
- [x] We will follow Fiori Guidelines when comes to the style of the graph
- [x] Our users are used to Grafana dashboards, so  when choosing a graph style use https://grafana.com/docs/grafana/latest/visualizations/ as inspiration
- [x] We show Volume Usage Limits
- [x] We use our existing dual graph to display free and used space

**Tasks**

- [ ] task 1

Lyczeq:
**Description**

- Add `Volume Space Usage` Graph on the PVC view
- Add translations
- Modify StatsPanel to add next graphs easier

**Related issue(s)**
Closes #1259


core-ui/src/shared/components/StatsGraph/StatsPanel.js
```diff
@@ -179,23 +221,23 @@ export function StatsPanel({ type, mode = 'single', ...props }) {
...
         <LayoutPanel.Filters>
-          <Dropdown
-            selectedKey={metric}
-            onSelect={(e, val) => setMetric(val.key)}
-            options={graphOptions.map(option => ({
-              key: option,
-              text: t(`graphs.${option}`),
-            }))}
-          />
+          {props?.title ? (
```
dariadomagala, line 229: Maybe we could just add an option for PVC to the `getGraphOptions`, and then here show a title if there is only 1 option, and a dropdown if there are multiple? Also, we could instead of passing the title from the parent component, use the title same approach as we do for the dropdown `t('graphs.${option}')`.

Wawrzyn321 (approved):

grego952 (approved):

PR #1280, opened 2022-04-25
Keep edit form open after a server error

### LINKED ISSUE #1100: Handle duplicate names in create form
**Description**

When I want to create a resource with a name that is already taken, entire form content is lost - all I get is an error prompt:
image

Would be nice to either keep the modal open until the requests succeeds, or display a warning/error on form - ResourceForm is on list views, so passing already fetched resources (to get their names) would be trivial. 

**Reasons**

This situation doesn't happen often, but our effort to fix this is rather minimal

**Acceptance Criteria**

- Those type or errors will not close the Modal - we display the error, but the user is not loosing the content of the widow
- We cover all the Resources.

wchrapka:
**Description**
don't close edit/add form after a server error

**Related issue(s)**
Closes #1100


Wawrzyn321 (approved):
props

PR #1309, opened 2022-05-09
Preferences

akucharska:
**Description**

- use local.storage in preferences modal view
- ...
- ...

**Related issue(s)**


akucharska:
Found better way in #1310

PR #1316, opened 2022-05-10
Hidden namespaces, disabled nodes as features

### LINKED ISSUE #1273: Move Disabled Nodes as a Feature
**Description**

At this moment to disable nodes we have to change a special node in the config. This should be a FEATURE with a FEATURE_FLAG and configuration
**Reasons**

We need to have a unified way to control features

**Reminder**

Update the documentation :)

**Attachments**

### LINKED ISSUE #1311: Move hiddenNamespaces as a feature
**Description**

At this moment to add hiddenNamespaces we have to change a special node in the config. This should be a FEATURE with a FEATURE_FLAG and configuration
**Reasons**

We need to have a unified way to control features

**Attachments**

wchrapka:
**Description**

- Moves the hidden namespace list to the config features.
- Moves the disabled nodes list to the config features.

**Related issue(s)**
Resolves #1311, resolves #1273

### DIFF
--- a/core/src/assets/defaultConfig.json
+++ b/core/src/assets/defaultConfig.json
@@ -64,7 +64,7 @@
         "selectors": []
       },
       "HIDDEN_NAMESPACES": {
-        "selector": [
+        "selectors": [
           "compass-system",
           "istio-system",
           "knative-eventing",
--- a/core/src/assets/defaultConfig.json
+++ b/core/src/assets/defaultConfig.json
@@ -60,6 +60,9 @@
           ]
         }
       },
+      "DISABLED_NODES": {
+        "selectors": []
+      },
       "HIDDEN_NAMESPACES": {
         "selector": [
           "compass-system",
--- a/core/src/assets/defaultConfig.json
+++ b/core/src/assets/defaultConfig.json
@@ -67,7 +67,7 @@
       },
       "HIDDEN_NAMESPACES": {
         "isEnabled": true,
-        "selectors": [
+        "config": [
           "compass-system",
           "istio-system",
           "knative-eventing",
--- a/core/src/assets/defaultConfig.json
+++ b/core/src/assets/defaultConfig.json
@@ -60,29 +60,25 @@
           ]
         }
       },
-      "DISABLED_NODES": {
-        "isEnabled": false,
-        "config": {
-          "nodes": []
-        }
-      },
       "HIDDEN_NAMESPACES": {
         "isEnabled": true,
-        "config": [
-          "compass-system",
-          "istio-system",
-          "knative-eventing",
-          "knative-serving",
-          "kube-public",
-          "kube-system",
-          "kyma-backup",
-          "kyma-installer",
-          "kyma-integration",
-          "kyma-system",
-          "natss",
-          "kube-node-lease",
-          "serverless-system"
-        ]
+        "config": {
+          "namespaces": [
+            "compass-system",
+            "istio-system",
+            "knative-eventing",
+            "knative-serving",
+            "kube-public",
+            "kube-system",
+            "kyma-backup",
+            "kyma-installer",
+            "kyma-integration",
+            "kyma-system",
+            "natss",
+            "kube-node-lease",
+            "serverless-system"
+          ]
+        }
       },
       "PROMETHEUS": {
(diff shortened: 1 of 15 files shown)


core/src/assets/defaultConfig.json
```diff
@@ -60,6 +60,26 @@
           ]
         }
       },
+      "DISABLED_NODES": {
+        "selectors": []
+      },
+      "HIDDEN_NAMESPACES": {
```
dariadomagala, line 66: every feature should have `isEnabled` field as well.

valentinvieriu: I agree

docs/features.md
```diff
@@ -110,6 +110,17 @@ The document lists and describes all the feature flags that are available in Kym
   },
   ```
 
+- **DISABLED_NODES** - an array of IDs of navigation nodes that are hidden from navigation. Format: `<category>.<nodeName>` or `<namespace>.<category>.<nodeName>`).
+
+  Default settings:
+
+  ```bash
+  "DISABLED_NODES": {
+    "isEnabled": true,
```
dariadomagala, line 119: I believe that by default it should be disabled.

docs/features.md
```diff
@@ -177,7 +213,7 @@ The document lists and describes all the feature flags that are available in Kym
   }
   ```
 
-- **KUBECONFIG_ID** – is used to configure the URL to which Busola sends a request to download a kubeconfig file. If you add `?kubeconfigID={your ID}` to the Busola URL, Busola tries to download the kubeconfig from `{kubeconfigUrl}/{yourID}`. If the operation succeeds, Busola adds the kubeconfing file to the cluster.
+* **KUBECONFIG_ID** – is used to configure the URL to which Busola sends a request to download a kubeconfig file. If you add `?kubeconfigID={your ID}` to the Busola URL, Busola tries to download the kubeconfig from `{kubeconfigUrl}/{yourID}`. If the operation succeeds, Busola adds the kubeconfing file to the cluster.
```
dariadomagala, line 216: Why those changes?

core/src/assets/defaultConfig.json
```diff
@@ -60,6 +60,28 @@
           ]
         }
       },
+      "DISABLED_NODES": {
+        "isEnabled": true,
```
dariadomagala, line 64: I don't think we need to add it to the default config since it does nothing here. I think it should be disabled by default.

core/src/assets/defaultConfig.json
```diff
@@ -60,6 +60,28 @@
...
+        "config": []
+      },
+      "HIDDEN_NAMESPACES": {
+        "isEnabled": true,
+        "config": [
```
dariadomagala, line 69: Everywhere `config` is an object, I believe we should stick to that approach, in case we would want to add something else to this configuration in the future.

docs/features.md
wchrapka: autochanges made by prettify

docs/features.md
```diff
@@ -110,6 +110,17 @@ The document lists and describes all the feature flags that are available in Kym
   },
   ```
 
+- **DISABLED_NODES** - an array of IDs of navigation nodes that are hidden from navigation. Format: `<category>.<nodeName>` or `<namespace>.<category>.<nodeName>`).
```
dariadomagala, line 113: I know it's not your code, it's been here before, but I think the pattern here is not correct. It suggests that the `<namespace>` is a variable when it's not.

```suggestion
- **DISABLED_NODES** - an array of IDs of navigation nodes that are hidden from navigation. Format: `<category>.<nodeName>` or `namespace.<category>.<nodeName>`).
```

dariadomagala (approved):
Small suggestion, you can discuss it with technical writers. From the code perspective, it looks good.

dariadomagala (approved):

docs/features.md
grego952: It looks like it does not affect the doc structure. Will it stay this way after the merge? :)

wchrapka: it will. That's what prettify does. It has a set of own formatting rules (that can be adjusted to some extent), but it doesn't change how the code works

grego952 (approved):

PR #1337, opened 2022-05-13
Remove Kyma version auto-detection when Kyma 2.1.1. is live on all clusters

### LINKED ISSUE #1204: Remove Kyma version auto-detection when Kyma 2.1.1. is live on all clusters 😈
**Description**
Remove autodetection of kyma version at graphs when 2.1.1. is live on all clusters.

**Reasons**
Promethus query for CPU changed in kyma 2.1.1. We added detection of kyma version to check which query should be used but It won't be needed after 2.1.1. is live on all clusters.

**Attachments**

Wawrzyn321:
**Description**

- As in title

**Related issue(s)**

Closes #1204 1204


Wawrzyn321:
`/hold`-ed for the time being while we're verifying if we don't need to support the previous versions.

akucharska (approved):

PR #1346, opened 2022-05-17
Added case for wrong accesstrategy type

### LINKED ISSUE #1349: UI crashed when Access strategy is wrong
**Description**
UI crashes when the user creates an API Rule using YAML and writes in the `Access strategies` wrong handler.

**Expected result**
UI is not crashing and displaying what is wrong.

**Actual result**
UI crashed.
image

**Steps to reproduce**
1. Create API Rule
2. Write in `accessStrategies` in handler something wrong (for example: test)

**Troubleshooting**

mrCherry97:
**Description**

- Changed displayed info from the handler,
- Added tooltip for wrong value,
- Added color changer for wrong value.

**Related issue(s)**
Closes #1349

### DIFF
--- a/core/src/i18n/en.yaml
+++ b/core/src/i18n/en.yaml
@@ -34,6 +34,7 @@ api-rules:
     messages:
       one-method: This access strategy requires at least one method.
       one-strategy-required: Provide at least one strategy.
+      unaccepted-type: Make sure, that Access Strategies has accepted values (noop, allow, oauth2_introspection, jwt) in YAML.
     title: Access Strategies
     tooltips:
       path: Path can contain alphanumeric characters and  '/', '.', '*', '?', '!', '-', '(', and ')'.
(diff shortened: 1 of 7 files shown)


core/src/i18n/en.yaml
```diff
@@ -34,6 +34,7 @@ api-rules:
     messages:
       one-method: This access strategy requires at least one method.
       one-strategy-required: Provide at least one strategy.
+      unaccepted-type: Make sure that Access Strategies has accepted values (noop, allow, oauth2_introspection, jwt) in YAML.
```
grego952, line 37:
```suggestion
      unaccepted-type: Make sure that Access Strategies have the accepted values (noop, allow, oauth2_introspection, jwt) in YAML.
```

Wawrzyn321 (approved):

grego952 (approved):

PR #1351, opened 2022-05-18
Removal of old Application Connector flow from Busola

### LINKED ISSUE #1250: Removal of old Application Connector flow from Busola
**Description**

Right now we are working on removing old and unused flows from Kyma, one of them includes registering and connecting Applications through Connector Service and Application Registry, those were deprecated around 2 years ago and we would like to get rid of them completely.

Due to removing some of the functionalities, we should align what is available in Busola - we need to drop the button for connecting Applications, and their Status field - those will no longer be used.

I've included screenshots of what needs to be removed in both the Application List and the Application Details views.

**Reasons**

We want to keep the UI consistent with what is available in Kyma.

**Attachments**

The issue for [removing App Registry and Connector Service](https://github.com/kyma-project/kyma/issues/13718)

App Details view:
application_details

App List view:
app_list

Wawrzyn321:
**Description**

- Introduce `APPLICATION_CONNECTOR_FLOW` feature that enables features mentioned in #1250 

**Related issue(s)**

Closes #1250


akucharska (approved):

akucharska (changes requested):
Something is not right witj cluster/test-applications.spec.js (it tries to test status even though we don't display one) and cluster/test-applications.spec.js

akucharska (approved):

PR #1357, opened 2022-05-19
Fix sentry 

Lyczeq:
**Description**

- ...
- ...
- ...

**Related issue(s)**


core-ui/src/components/Clusters/views/ClusterList.js
```diff
@@ -43,13 +43,13 @@ function ClusterList() {
   }
 
   const styleActiveCluster = entry => {
-    return entry.kubeconfig['current-context'] === activeClusterName
+    return entry?.kubeconfig['current-context'] === activeClusterName
```
wchrapka: if the error is "cannot read kubeconfig of undefined", then "current-context" won't be defined as well 
```suggestion
    return entry?.kubeconfig?.['current-context'] === activeClusterName
```

core/src/luigi-config/feature-discovery.js
```diff
@@ -115,7 +115,7 @@ export async function updateFeature(featureName) {
       const activeCluster = getActiveCluster();
       const permissionSet = await fetchPermissions(
         getAuthData(),
-        getCurrentContextNamespace(activeCluster.kubeconfig),
+        getCurrentContextNamespace(activeCluster?.kubeconfig),
```
wchrapka, line 118: that function is trying to get current-context from this argument, it will lead to another error
```
const currentContextName = kubeconfig['current-context'];
```

core/src/luigi-config/navigation/navigation-data-init.js
```diff
@@ -210,12 +210,12 @@ export async function createNavigation() {
     const activeCluster = getActiveCluster();
     const permissionSet = await fetchPermissions(
       authData,
-      getCurrentContextNamespace(activeCluster.kubeconfig),
+      getCurrentContextNamespace(activeCluster?.kubeconfig),
```
wchrapka, line 213: it will lead to the same error as above

core/src/luigi-config/navigation/navigation-data-init.js
```diff
@@ -425,7 +425,7 @@ export async function createNavigationNodes(
 
 async function getNamespaces() {
   const activeCluster = getActiveCluster();
-  const namespace = getCurrentContextNamespace(activeCluster.kubeconfig);
+  const namespace = getCurrentContextNamespace(activeCluster?.kubeconfig);
```
wchrapka, line 428: same here

core/src/luigi-config/communication.js
```diff
@@ -97,7 +97,7 @@ export const communication = {
     'busola.addCluster': async ({ params, switchCluster = true }) => {
       await saveClusterParams(params);
       if (switchCluster) {
-        await setCluster(params.kubeconfig['current-context']);
+        await setCluster(params?.kubeconfig['current-context']);
```
wchrapka: it will lead to an error in the setCluster function: "Cannot read undefined of clusters"

wchrapka (changes requested):
I've added some options chainings in a separate commit (now a child would throw an error previously thrown by its parent). Please review what happens when we pass undefined to some functions. Some of the proposed corrections are going to just move the errors down the line.

wchrapka (approved):

grego952 (approved):

wchrapka (approved):

PR #1362, opened 2022-05-23
Clean json packages

### LINKED ISSUE #1274: Clean up package.jsons
**Description**

When we've had multiple microfrontends, we've had a shared package.json with the packages that are used among more than just one microfrontend. Now, we only have two frontends, and not many packages are actually shared between them, yet they are still in the main package.json. Please move the packages that are used only in one of those frontends, to theirs "sub-package.jsons".

**Reasons**

It's easier to read, and easier to maintain. It might also decrease the time to bootstap the app, and size of specific applications since not so many packages will be in their node_modules.

**Attachments**

wchrapka:
**Description**

Moves some of the dependencies from master to core-ui/core package. 

**Related issue(s)**
Closes  #1274


Wawrzyn321 (approved):

Wawrzyn321 (approved):

wchrapka:
<html>
<body>


/test pre-busola-integration-test-k3s-namespace
--



</body>
</html>

Wawrzyn321 (approved):

PR #1363, opened 2022-05-24
Add feature tests

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**


wchrapka:
IMO testing it for cluster and namespace falls onto DRY (testing it twice). 

I would test if app behaves normally, if a flag like "PROMETHEUS" is not defined at all, against "reading isEnabled of undefined"

Are we treating other flags separately?

tests/tests/cluster/test-navigation-features.spec.js
```diff
@@ -0,0 +1,55 @@
...
+
+    // visual resources
+    cy.navigateTo('Configuration', 'Cluster Role Bindings');
+
+    cy.getIframeBody()
+      .contains('cluster-admin')
+      .click();
+
+    // wait for view to stabilize
+    cy.wait(5000);
```
wchrapka, line 49: maybe 

```suggestion
      cy.getIframeBody()
      .contains('system:masters') // means the view is rendered
      .should('be.visible');
```

Wawrzyn321: I get it, but for me it navigates to cluster roles... I'll try to find a better selector.

wchrapka (approved):

PR #1378, opened 2022-06-02
Mock backend for tests

### LINKED ISSUE #1223: SPIKE: Evaluate The possibility of using Mock Backend
**Description**

To speed up our testing would be great to use mocks instead of real k8s api server. Investigate what tools can be used to facilitate this

Minimum target is to mock first pace ( cluster overview ) so we can run Lighthouse tests on this in a predictable way.

**Possible tools to investigate**
- https://mswjs.io/

**Target timing**
2 MD

wchrapka:
**Description**

Mock backend requests with MSW JS

**Related issue(s)**

Resolves #1223


core-ui/src/mocks/data/queryRange2.js
```diff
@@ -0,0 +1,31 @@
+import { mockGetRequest } from 'mocks/createMock';
+
+export const clusterDetailsQueryRange2 = mockGetRequest(
```
wchrapka, line 3: Actually there should be only 2 queryRange mocks returning different data based on the GET req.parameters

PR #1379, opened 2022-06-02
Sorting on list

### LINKED ISSUE #1127: List View should have sorting by column
**Description**

All the list views should have a sorting option by column. This sorting can be ascending or descending when user clicks on header 

**Reasons**

**Attachments**

image

mrCherry97:
**Description**

- ...
- ...
- ...

**Related issue(s)**
Closes #1127


mrCherry97:
TODO:
- [ ] move sorting to `GenericList`,
- [ ] dropdown in `GenericList`,
- [ ] move sorting functions to separate files,
- [ ] time sorting in `ResourceList` transferred to `GenericList`.

dariadomagala (changes requested):
1. The icon should be disabled when there are no entries on the list.
2. It would be nice to have a tooltip for the icon (maybe we could add one for the search icon as well)
3. I believe we should add it to all lists, so that would be also events for example (sorted by last seen and types maybe)

dariadomagala (approved):

grego952 (dismissed):

dariadomagala (approved):

grego952 (approved):

Issue #1402, opened 2022-06-09
Extensibility - list property as an object instead of an array breaks the view.


dariadomagala:
**Description**

If the user puts `"list": {}` instead of `"list": []` in the extensibility configuration, it breaks the view (although the Error Boundaries work well here, it doesn't break the whole app).

```Uncaught TypeError: (resMetaData.list || []).map is not a function```

**Expected result**

It could show a warning, but shouldn't break the view.

**Actual result**

It breaks the view, with no explanation why.

**Steps to reproduce**

**Troubleshooting**

PR #1423, opened 2022-06-10
Fix crashing view if a user puts an object instead of an array

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Fixes https://github.com/kyma-project/busola/issues/1402

### DIFF
--- a/core-ui/src/components/Extensibility/ExtensibilityDetails.js
+++ b/core-ui/src/components/Extensibility/ExtensibilityDetails.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { isArray } from 'lodash';
 
 import { usePrepareDetailsProps } from 'resources/helpers';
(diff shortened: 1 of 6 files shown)


core-ui/src/components/Extensibility/ExtensibilityDetails.js
```diff
@@ -43,12 +44,16 @@ export const ExtensibilityDetailsCore = ({ resMetaData }) => {
...
+              header: widgetT(def),
+              value: resource => (
+                <Widget value={resource} structure={def} schema={schema} />
+              ),
+            }))
+          : []
+      }
       customComponents={[
         resource => (
           <Widget value={resource} structure={body} schema={schema} />
```
akucharska, line 59: Shouldn't we check if `body` is an array as well? If it is an object we display it on the page.
Screen Shot 2022-06-10 at 12 14 22
Screen Shot 2022-06-10 at 12 15 15

akucharska (approved):

PR #1424, opened 2022-06-10
added props for displaying button in ErrorBoundary

### LINKED ISSUE #1403: EXT: Incorrect configuration displays a button that doesn't work
**Description**
When the extension configuration is incorrect I get information and a not-working `Go back` button.

**Expected result**
Clicking on the `Go back` button should redirect me to e.g. list view of the customized resources.

**Actual result**
Clicking on the `Go back` button does nothing.
image

**Steps to reproduce**
My configuration of the Vertical Pod Autoscaler:
The details configuration is the cause of an error.
**Resource**

```
{
  "kind": "VerticalPodAutoscalers",
  "group": "autoscaling.k8s.io",
  "version": "v1"
}
```

**Navigation**

```
{
  "path": "vertical-pod-autoscalers",
  "scope": "namespace"
}
```

**Details**
```
{
"header": [
{ "path": "spec.updatePolicy.updateMode" , "name": "Update Mode"}
],
"body":[
    {
      "name": "Container Policies",
      "path": "spec.resourcePolicy.containerPolicies",
      "widget": "Table",
      "columns": [{ "path": "containerName" }, { "path": "controlledValues" }]
    }
]
}
```

### LINKED ISSUE #1409: Extensibility - list doesn't refresh immediately
**Description**

1. When you go to the Custom Resources page and then go to the different Custom Resources page the data are not refreshed immediately. You  still see old entries in the table.

Screen Shot 2022-06-09 at 11 28 16
Screen Shot 2022-06-09 at 11 28 56

2. When you go the the list view that has wrong definition, you see `Your extension configuration is incorrect. [Go back button]`, but then after changing page, you have to click on `[Go back button]` even though you are on different page.
Screen Shot 2022-06-09 at 11 27 09
Screen Shot 2022-06-09 at 11 27 27

**Expected result**
You see proper data on each list view

**Actual result**

You see a list with entries from the previous resource 

**Steps to reproduce**

* Go to Custom Resources -> Pizzas
* Then go to Custom Resources -> GitRepository
* 💥 For a couple of seconds you still see pizzas in the table even if the title page says GitRepository

**Troubleshooting**

mrCherry97:
**Description**

- ...
- ...
- ...

**Related issue(s)**
Closes #1403 
Fixes #1409 (Closes 2 topic in that issue)

### DIFF
--- a/core-ui/src/components/Extensibility/ExtensibilityList.js
+++ b/core-ui/src/components/Extensibility/ExtensibilityList.js
@@ -53,6 +53,7 @@ export const ExtensibilityList = () => {
       <ErrorBoundary
         customMessage={t('extensibility.error')}
         displayButton={false}
+        key={resMetaData.navigation.path}
       >
         <ExtensibilityListCore resMetaData={resMetaData} />
       </ErrorBoundary>
--- a/core-ui/src/components/Extensibility/ExtensibilityList.js
+++ b/core-ui/src/components/Extensibility/ExtensibilityList.js
@@ -50,7 +50,10 @@ export const ExtensibilityList = () => {
 
   return (
     <TranslationBundleContext.Provider value={resMetaData.navigation.path}>
-      <ErrorBoundary customMessage={t('extensibility.error')}>
+      <ErrorBoundary
+        customMessage={t('extensibility.error')}
+        displayButton={false}
+      >
         <ExtensibilityListCore resMetaData={resMetaData} />
       </ErrorBoundary>
     </TranslationBundleContext.Provider>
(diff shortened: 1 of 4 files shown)


core-ui/src/components/Extensibility/ExtensibilityList.js
```diff
@@ -50,7 +50,10 @@ export const ExtensibilityList = () => {
 
   return (
     <TranslationBundleContext.Provider value={resMetaData.navigation.path}>
-      <ErrorBoundary customMessage={t('extensibility.error')}>
+      <ErrorBoundary
```
dariadomagala, line 53: We also need it for details and forms

dariadomagala (changes requested):

dariadomagala (dismissed):

dariadomagala (approved):

PR #1426, opened 2022-06-10
Add nodejs16 runtime

kwiatekus:
**Description**

- add nodejs16 as non-default runtime

**Related issue(s)**
https://github.com/kyma-project/kyma/issues/14067


kwiatekus:
@valentinvieriu @majakurcius @Sawthis 
Nodejs16 was introduced in 2.3 version
This should not get deployed on prod until kyma 2.3 is available on prod

mrCherry97:
If you want lastest `nodejs16` as default in Create Modal you have to change in `Functions/helpers.js` value for runtime.
We have to investigate how to fix this validation warning:
image

kwiatekus:
> If you want lastest `nodejs16` as default in Create Modal you have to change in `Functions/helpers.js` value for runtime. We have to investigate how to fix this validation warning: image

@mrCherry97 
nodejs16 runtime is very different from the other nodejs runtimes (it is not kubeless based). Therefore we dont want to do it default yet. We will do that later

mrCherry97 (approved):

Wawrzyn321 (approved):

PR #1452, opened 2022-06-15
Fix incorrect Table Widget to not break the page 

### LINKED ISSUE #1440: Extensibility - incorrectly configured Table widget breaks the page
**Description**

The table widget breaks the page if it's configured incorrectly.
Example 
```
{
      "path": "metadata",
      "widget": "Table",
      "children": [{ "path": "name" }]
}
```

**Expected result**

It should show an error, or display the table partially.

**Actual result**

It breaks the page, with errors in the console
```Invalid prop `entries` of type `object` supplied to `GenericList`, expected an array.```
```entries.filter is not a function```

**Steps to reproduce**

**Troubleshooting**

### PR DESCRIPTION
**Pull request by Lyczeq**
**Description**

- Fixed the bug of incorrect Table Widget breaking the details page
- Added a key prop to remove console errors
- Added a prop to the GenericList component that is displayed when there's an incorrect extensibility configuration 
- Added translations

**Related issue(s)**
Resolves #1440

### DIFF
--- a/core-ui/src/components/Extensibility/components/Plain.js
+++ b/core-ui/src/components/Extensibility/components/Plain.js
@@ -6,7 +6,7 @@ export function Plain({ value, structure, schema }) {
   return (
     <div>
       {structure.children?.map(def => (
-        <Widget value={value} structure={def} schema={schema} />
+        <Widget value={value} structure={def} schema={schema} key={def} />
       ))}
     </div>
   );
--- a/core-ui/src/components/Extensibility/components/Plain.js
+++ b/core-ui/src/components/Extensibility/components/Plain.js
@@ -5,9 +5,11 @@ import { Widget } from './Widget';
 export function Plain({ value, structure, schema }) {
   return (
     <div>
-      {structure.children?.map(def => (
-        <Widget value={value} structure={def} schema={schema} key={def} />
-      ))}
+      {structure.children?.map((def, idx) => {
+        return (
+          <Widget value={value} structure={def} schema={schema} key={idx} />
+        );
+      })}
     </div>
   );
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: core-ui/src/components/Extensibility/components/Plain.js [hunk 1]
```diff
@@ -6,7 +6,7 @@ export function Plain({ value, structure, schema }) {
   return (
     <div>
       {structure.children?.map(def => (
-        <Widget value={value} structure={def} schema={schema} />
+        <Widget value={value} structure={def} schema={schema} key={def} />
```
Wawrzyn321, line 9: This won't do - since `def` is an object, the key would be always `[Object object]`.

core-ui/src/components/Extensibility/components/Table.js
```diff
@@ -1,12 +1,33 @@
...
+import { useTranslation } from 'react-i18next';
 
 import { GenericList } from 'shared/components/GenericList/GenericList';
 
 import { useGetTranslation } from '../helpers';
 import { Widget } from './Widget';
 
+const handleTableValue = (value, t) => {
+  switch (true) {
+    case value === undefined: {
```
Wawrzyn321, line 11: Maybe use `isNil` from lodash to cover the `null` value?

core-ui/src/components/Extensibility/components/Table.js
```diff
@@ -1,12 +1,33 @@
...
+    }
+    case Array.isArray(value): {
+      return {
+        entries: value,
+      };
+    }
+    default: {
+      return {
+        entries: [],
+        invalidConfigurationMessage: t('extensibility.widgets.table.error'),
```
Wawrzyn321, line 22: I'd go with less specific name here, so we can reuse this prop elsewhere. `genericErrorMessage`?

Wawrzyn321 (approved):

PR #1468, opened 2022-06-21
Refactor Edit Config Map to allow code editors

### LINKED ISSUE #1442: Refactor Edit Config Map to allow code editors
As [Sebastian](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#sebastian---tech-lead) and [Mathew](https://github.com/kyma-project/community/blob/main/manifesto/personas.md#matthew---cloud-native-software-engineer) I would like to be able to edit comfortably and with syntax highlighting a multiline config map

**Acceptance Criteria**

- [x] The config map editor supports Monaco editor
- [x] Allow user to choose one from dropdown list so that  syntax highlighting works
- [x] We introduce collapsible panels to try to limit the clutter 
- [x] Follow Project DOD

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Closes #1442


core-ui/src/shared/components/MonacoEditorESM/Editor.js
```diff
@@ -50,6 +55,37 @@ export function Editor({
...
+      return () => {
+        focusRef.current.dispose();
+        blurRef.current.dispose();
+      };
+    }
+  }, [onBlur, onFocus, editorRef]);
+
+  if (editorRef.current) {
+    changeRef.current?.dispose();
+    changeRef.current = editorRef.current.onDidChangeModelContent(() => {
```
wchrapka, line 83: it installs onDidChangeModelContent listener on every content change, while the same listener is installed in line 150. Can we remove this code?

core-ui/src/shared/components/MonacoEditorESM/Editor.js
```diff
@@ -50,6 +55,37 @@ export function Editor({
...
+        if (typeof onFocus === 'function') {
+          onFocus();
+        }
+      });
+      return () => {
+        focusRef.current.dispose();
+        blurRef.current.dispose();
+      };
+    }
+  }, [onBlur, onFocus, editorRef]);
```
wchrapka, line 79: onBlur is recreated with new identity on every render, hence this code is executed on every render. If it's supposed to work that way, it shouldn't be placed in useEffect.

Wawrzyn321: That's true, but it wouldn't be a case if someone wraps `onBlur` in `useCallback`, right?

core-ui/src/shared/components/MonacoEditorESM/Editor.js
Wawrzyn321: Right, I'll remove it and the blur/focus one.

Wawrzyn321: Or NO, if I remove the handler in function near `onMount` making changes in YAML tab won't work...

core-ui/src/shared/ResourceForm/fields/RichEditorDataField/RichEditorSection.js
```diff
@@ -0,0 +1,93 @@
...
+      options={getAvailableLanguages().map(l => ({
+        key: l.id,
+        text: l.aliases?.[0] || l.id,
+      }))}
+      selectedKey={language || ''}
+      onSelect={(e, { key: language }) => {
+        e.stopPropagation(); // don't collapse the section
+        onChange({ language });
+      }}
+      placeholder={t('common.statuses.unknown')}
```
dariadomagala, line 37: Could we add a more descriptive placeholder like 'choose a language' etc? We have something like this for presets.

dariadomagala (commented):
1. 
* Choose a language
* start typing something
* click outside Monaco
* the chosen language might change if it doesn't match to the previously chosen language structure
was it intended?

2. 
Not in this task, but should we also add language detection to the editor in the details view, or is it too much?

core-ui/src/shared/ResourceForm/fields/RichEditorDataField/RichEditorSection.js
Wawrzyn321: 1. Me and @wchrapka are on it.
2. Damn, I was thinking about this, but I didn't want to add too many changes in this PR. I'll do it here then, sure.
3. I'll add a better placeholder.

Wawrzyn321:
Known issue - changes to languages aren't propagated between _Simple_ and _Advanced_. To achieve this, we would need to somehow share the state between them. I've outlined it in https://github.com/kyma-project/busola/issues/1481

dariadomagala (commented):
1. did we lose the language detection in the crate form or was it on purpose?
2. also, can we remove repeated values (or describe the difference if there is any)
Screen Shot 2022-06-24 at 10 36 13

Wawrzyn321:
1. There was no language detection in the create form - just like there's no detection when we add new keys in edit mode.
2. Fixed.

dariadomagala (dismissed):

grego952 (approved):

dariadomagala (approved):

PR #1469, opened 2022-06-21
read extensions from all namespaces

qbalukom:


core/src/luigi-config/customResources.js
```diff
@@ -24,15 +28,25 @@ async function loadBusolaClusterCRs() {
 }
 
 async function loadTargetClusterCRs(authData) {
+  const activeCluster = getActiveCluster();
+  const namespace = getCurrentContextNamespace(activeCluster?.kubeconfig);
```
Wawrzyn321, line 32:
```suggestion
  const namespace = getCurrentContextNamespace(activeCluster?.kubeconfig) || 'kube-public';
```

If clusterwide call fails and we don't have a context namespace, we don't want to make a call for `undefined` namespace. I'd say `kube-public` is a good enough fallback.

Wawrzyn321 (dismissed):

Wawrzyn321 (approved):

PR #1488, opened 2022-06-23
Fix allowEdit in Functions

### LINKED ISSUE #1487: Functions can't be edit by UI Form
**Description**
In functions is only `Edit YAML` 

**Expected result**
Functions can be edit by `UI Form`

**Actual result**
image

**Steps to reproduce**
1. Create a Function
2. Go to Details

mrCherry97:
**Description**
I can't edit functions using UI Form, so I bring back allowEdit flag.

- ...
- ...
- ...

**Related issue(s)**
Closes #1487

### DIFF
--- a/core-ui/src/resources/Functions/FunctionCreate.js
+++ b/core-ui/src/resources/Functions/FunctionCreate.js
@@ -248,3 +248,5 @@ export function FunctionCreate({
     </ResourceForm>
   );
 }
+
+FunctionCreate.allowEdit = true;


mrCherry97:
TO DO:
`Edit` button is opening like Create Modal and isn't getting data.
image

core-ui/src/resources/Functions/FunctionCreate.js
```diff
@@ -1,6 +1,7 @@
 import React, { useEffect, useState, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import * as jp from 'jsonpath';
+import * as _ from 'lodash';
```
qbalukom, line 4: I believe there's a bit of a push towards using
```suggestion
import { cloneDeep } from 'lodash';
```
instead - makes for a somewhat cleaner code.
It's not done so everywhere, but I think it's good practice to at least use it like this in new places.

core-ui/src/resources/Functions/FunctionCreate.js
```diff
@@ -132,8 +139,12 @@ export function FunctionCreate({
       setResource={setFunction}
       onChange={onChange}
       formElementRef={formElementRef}
-      createUrl={`/apis/serverless.kyma-project.io/v1alpha1/namespaces/${namespace}/functions`}
+      createUrl={
+        resourceUrl ||
+        `/apis/serverless.kyma-project.io/v1alpha1/namespaces/${namespace}/functions`
```
qbalukom, line 144: i know we've had a discussion about this - but maybe it might be possible to cleanly move this to namespace overview? (same with deployments)
if it can't be done cleanly, then let's keep it as-is

qbalukom (changes requested):

PR #1495, opened 2022-06-27
Extensibility - show connected resources

### LINKED ISSUE #1471: Extensibility - Show connected resources
**Description**

Add a possibility to show a list of connected resources (for example subscriptions in the Service details view).

**Reasons**

Sometimes it is useful to see the connection between resources.

**Examples**

- you can attach k8s events
- configmaps
- pods ( based on )
- not Visual Graphs

**Attachments**

Wawrzyn321:
**Description**

- As in commit messages

**Testing**

<details>
 <summary>Drop it into customResources.json.</summary>

```
{
    "services": {
      "resource": {
        "kind": "Service",
        "group": "core",
        "version": "v1",
        "scope": "namespace",
        "path": "services-new"
      },
      "details": {
        "header": [
          {
            "path": "$apiRules",
            "name": "API Rule names & statuses, this is so  f a n c y",
            "valuePreprocessor": "PendingWrapper",
            "formula": "$join(data[*].(metadata.name & ' (' & status.APIRuleStatus.code & ')'), ', ')"
          }
        ],
        "body": [
          {
            "widget": "ResourceList",
            "path": "$apiRules"
          },
          {
            "name": "Related Subscriptions",
            "widget": "ResourceList",
            "path": "$subscriptions",
            "columns": [{
              "path": "status.cleanEventTypes[0]",
              "name": "My very own event types",
              "widget": "Badge"
            }],
            "showSearchField": false
          }
        ]
      },
      "relations": {
        "apiRules": {
          "kind": "APIRule",
          "group": "apis/gateway.kyma-project.io",
          "version": "v1alpha1",
          "selector": "$filter(data, function($item) {resource.metadata.name = $item.spec.service.name})"
        },
        "subscriptions": {
          "kind": "Subscription",
          "group": "apis/eventing.kyma-project.io",
          "version": "v1alpha1",
          "selector": "$filter(data, function($item) {resource.metadata.name = $substringBefore($substringAfter($item.spec.sink, 'http://'), '.')})"
        }
      },
      "list": [
        {
          "path": "metadata.ownerReferences",
          "widget": "ControlledByKind"
        },
        {
          "path": "spec.type"
        },
        {
          "path": "spec.clusterIP"
        },
        {
          "path": "spec.externalIPs"
        },
        {
          "path": "spec.ports"
        }
      ],
      "form": [
        {
          "path": "spec.ports.name",
          "simple": true
        }, {
          "path": "spec.ports.protocol",
          "simple": true
        }, {
          "path": "spec.ports.port",
          "simple": true
        }, {
          "path": "spec.ports.targetPort",
          "simple": true
        }, {
          "path": "spec.clusterIP"
        }, {
          "path": "spec.externalIPs"
        }, {
          "path": "spec.type"
        }
      ],
      "schema": {
        "description": "Service is a named abstraction of software service (for example, mysql) consisting of local port (for example 3306) that the proxy listens on, and the selector that determines which pods will answer requests sent through the proxy.",
        "type": "object",
        "properties": {
          "apiVersion": {
            "description": "APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources",
            "type": "string"
          },
          "kind": {
            "description": "Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds",
            "type": "string"
          },
          "metadata": {
            "description": "ObjectMeta is metadata that all persisted resources must have, which includes all objects users must create.",
            "type": "object",
            "properties": {
              "annotations": {
                "description": "Annotations is an unstructured key value map stored with a resource that may be set by external tools to store and retrieve arbitrary metadata. They are not queryable and should be preserved when modifying objects. More info: http://kubernetes.io/docs/user-guide/annotations",
                "type": "object",
                "additionalProperties": {
                  "type": "string"
                }
              },
              "clusterName": {
                "description": "The name of the cluster which the object belongs to. This is used to distinguish resources with same name and namespace in different clusters. This field is not set anywhere right now and apiserver is going to ignore it if set in create or update request.",
                "type": "string"
              },
              "creationTimestamp": {
                "description": "Time is a wrapper around time.Time which supports correct marshaling to YAML and JSON.  Wrappers are provided for many of the factory methods that the time package offers.",
                "format": "date-time",
                "type": "string"
              },
              "deletionGracePeriodSeconds": {
                "description": "Number of seconds allowed for this object to gracefully terminate before it will be removed from the system. Only set when deletionTimestamp is also set. May only be shortened. Read-only.",
                "format": "int64",
                "type": "integer"
              },
              "deletionTimestamp": {
                "description": "Time is a wrapper around time.Time which supports correct marshaling to YAML and JSON.  Wrappers are provided for many of the factory methods that the time package offers.",
                "format": "date-time",
                "type": "string"
              },
              "finalizers": {
                "description": "Must be empty before the object is deleted from the registry. Each entry is an identifier for the responsible component that will remove the entry from the list. If the deletionTimestamp of the object is non-nil, entries in this list can only be removed. Finalizers may be processed and removed in any order.  Order is NOT enforced because it introduces significant risk of stuck finalizers. finalizers is a shared field, any actor with permission can reorder it. If the finalizer list is processed in order, then this can lead to a situation in which the component responsible for the first finalizer in the list is waiting for a signal (field value, external system, or other) produced by a component responsible for a finalizer later in the list, resulting in a deadlock. Without enforced ordering finalizers are free to order amongst themselves and are not vulnerable to ordering changes in the list.",
                "type": "array",
                "items": {
                  "type": "string"
                },
                "x-kubernetes-patch-strategy": "merge"
              },
              "generateName": {
                "description": "GenerateName is an optional prefix, used by the server, to generate a unique name ONLY IF the Name field has not been provided. If this field is used, the name returned to the client will be different than the name passed. This value will also be combined with a unique suffix. The provided value has the same validation rules as the Name field, and may be truncated by the length of the suffix required to make the value unique on the server.\n\nIf this field is specified and the generated name exists, the server will NOT return a 409 - instead, it will either return 201 Created or 500 with Reason ServerTimeout indicating a unique name could not be found in the time allotted, and the client should retry (optionally after the time indicated in the Retry-After header).\n\nApplied only if Name is not specified. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#idempotency",
                "type": "string"
              },
              "generation": {
                "description": "A sequence number representing a specific generation of the desired state. Populated by the system. Read-only.",
                "format": "int64",
                "type": "integer"
              },
              "labels": {
                "description": "Map of string keys and values that can be used to organize and categorize (scope and select) objects. May match selectors of replication controllers and services. More info: http://kubernetes.io/docs/user-guide/labels",
                "type": "object",
                "additionalProperties": {
                  "type": "string"
                }
              },
              "managedFields": {
                "description": "ManagedFields maps workflow-id and version to the set of fields that are managed by that workflow. This is mostly for internal housekeeping, and users typically shouldn't need to set or understand this field. A workflow can be the user's name, a controller's name, or the name of a specific apply path like \"ci-cd\". The set of fields is always in the version that the workflow used when modifying the object.",
                "type": "array",
                "items": {
                  "description": "ManagedFieldsEntry is a workflow-id, a FieldSet and the group version of the resource that the fieldset applies to.",
                  "type": "object",
                  "properties": {
                    "apiVersion": {
                      "description": "APIVersion defines the version of this resource that this field set applies to. The format is \"group/version\" just like the top-level APIVersion field. It is necessary to track the version of a field set because it cannot be automatically converted.",
                      "type": "string"
                    },
                    "fieldsType": {
                      "description": "FieldsType is the discriminator for the different fields format and version. There is currently only one possible value: \"FieldsV1\"",
                      "type": "string"
                    },
                    "fieldsV1": {
                      "description": "FieldsV1 stores a set of fields in a data structure like a Trie, in JSON format.\n\nEach key is either a '.' representing the field itself, and will always map to an empty set, or a string representing a sub-field or item. The string will follow one of these four formats: 'f:<name>', where <name> is the name of a field in a struct, or key in a map 'v:<value>', where <value> is the exact json formatted value of a list item 'i:<index>', where <index> is position of a item in a list 'k:<keys>', where <keys> is a map of  a list item's key fields to their unique values If a key maps to an empty Fields value, the field that key represents is part of the set.\n\nThe exact format is defined in sigs.k8s.io/structured-merge-diff",
                      "type": "object"
                    },
                    "manager": {
                      "description": "Manager is an identifier of the workflow managing these fields.",
                      "type": "string"
                    },
                    "operation": {
                      "description": "Operation is the type of operation which lead to this ManagedFieldsEntry being created. The only valid values for this field are 'Apply' and 'Update'.",
                      "type": "string"
                    },
                    "time": {
                      "description": "Time is a wrapper around time.Time which supports correct marshaling to YAML and JSON.  Wrappers are provided for many of the factory methods that the time package offers.",
                      "format": "date-time",
                      "type": "string"
                    }
                  }
                }
              },
              "name": {
                "description": "Name must be unique within a namespace. Is required when creating resources, although some resources may allow a client to request the generation of an appropriate name automatically. Name is primarily intended for creation idempotence and configuration definition. Cannot be updated. More info: http://kubernetes.io/docs/user-guide/identifiers#names",
                "type": "string"
              },
              "namespace": {
                "description": "Namespace defines the space within which each name must be unique. An empty namespace is equivalent to the \"default\" namespace, but \"default\" is the canonical representation. Not all objects are required to be scoped to a namespace - the value of this field for those objects will be empty.\n\nMust be a DNS_LABEL. Cannot be updated. More info: http://kubernetes.io/docs/user-guide/namespaces",
                "type": "string"
              },
              "ownerReferences": {
                "description": "List of objects depended by this object. If ALL objects in the list have been deleted, this object will be garbage collected. If this object is managed by a controller, then an entry in this list will point to this controller, with the controller field set to true. There cannot be more than one managing controller.",
                "type": "array",
                "items": {
                  "description": "OwnerReference contains enough information to let you identify an owning object. An owning object must be in the same namespace as the dependent, or be cluster-scoped, so there is no namespace field.",
                  "type": "object",
                  "required": [
                    "apiVersion",
                    "kind",
                    "name",
                    "uid"
                  ],
                  "properties": {
                    "apiVersion": {
                      "description": "API version of the referent.",
                      "type": "string"
                    },
                    "blockOwnerDeletion": {
                      "description": "If true, AND if the owner has the \"foregroundDeletion\" finalizer, then the owner cannot be deleted from the key-value store until this reference is removed. Defaults to false. To set this field, a user needs \"delete\" permission of the owner, otherwise 422 (Unprocessable Entity) will be returned.",
                      "type": "boolean"
                    },
                    "controller": {
                      "description": "If true, this reference points to the managing controller.",
                      "type": "boolean"
                    },
                    "kind": {
                      "description": "Kind of the referent. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds",
                      "type": "string"
                    },
                    "name": {
                      "description": "Name of the referent. More info: http://kubernetes.io/docs/user-guide/identifiers#names",
                      "type": "string"
                    },
                    "uid": {
                      "description": "UID of the referent. More info: http://kubernetes.io/docs/user-guide/identifiers#uids",
                      "type": "string"
                    }
                  }
                },
                "x-kubernetes-patch-merge-key": "uid",
                "x-kubernetes-patch-strategy": "merge"
              },
              "resourceVersion": {
                "description": "An opaque value that represents the internal version of this object that can be used by clients to determine when objects have changed. May be used for optimistic concurrency, change detection, and the watch operation on a resource or set of resources. Clients must treat these values as opaque and passed unmodified back to the server. They may only be valid for a particular resource or set of resources.\n\nPopulated by the system. Read-only. Value must be treated as opaque by clients and . More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency",
                "type": "string"
              },
              "selfLink": {
                "description": "SelfLink is a URL representing this object. Populated by the system. Read-only.\n\nDEPRECATED Kubernetes will stop propagating this field in 1.20 release and the field is planned to be removed in 1.21 release.",
                "type": "string"
              },
              "uid": {
                "description": "UID is the unique in time and space value for this object. It is typically generated by the server on successful creation of a resource and is not allowed to change on PUT operations.\n\nPopulated by the system. Read-only. More info: http://kubernetes.io/docs/user-guide/identifiers#uids",
                "type": "string"
              }
            }
          },
          "spec": {
            "description": "ServiceSpec describes the attributes that a user creates on a service.",
            "type": "object",
            "properties": {
              "allocateLoadBalancerNodePorts": {
                "description": "allocateLoadBalancerNodePorts defines if NodePorts will be automatically allocated for services with type LoadBalancer.  Default is \"true\". It may be set to \"false\" if the cluster load-balancer does not rely on NodePorts. allocateLoadBalancerNodePorts may only be set for services with type LoadBalancer and will be cleared if the type is changed to any other type. This field is alpha-level and is only honored by servers that enable the ServiceLBNodePortControl feature.",
                "type": "boolean"
              },
              "clusterIP": {
                "description": "clusterIP is the IP address of the service and is usually assigned randomly. If an address is specified manually, is in-range (as per system configuration), and is not in use, it will be allocated to the service; otherwise creation of the service will fail. This field may not be changed through updates unless the type field is also being changed to ExternalName (which requires this field to be blank) or the type field is being changed from ExternalName (in which case this field may optionally be specified, as describe above).  Valid values are \"None\", empty string (\"\"), or a valid IP address. Setting this to \"None\" makes a \"headless service\" (no virtual IP), which is useful when direct endpoint connections are preferred and proxying is not required.  Only applies to types ClusterIP, NodePort, and LoadBalancer. If this field is specified when creating a Service of type ExternalName, creation will fail. This field will be wiped when updating a Service to type ExternalName. More info: https://kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies",
                "type": "string"
              },
              "clusterIPs": {
                "description": "ClusterIPs is a list of IP addresses assigned to this service, and are usually assigned randomly.  If an address is specified manually, is in-range (as per system configuration), and is not in use, it will be allocated to the service; otherwise creation of the service will fail. This field may not be changed through updates unless the type field is also being changed to ExternalName (which requires this field to be empty) or the type field is being changed from ExternalName (in which case this field may optionally be specified, as describe above).  Valid values are \"None\", empty string (\"\"), or a valid IP address.  Setting this to \"None\" makes a \"headless service\" (no virtual IP), which is useful when direct endpoint connections are preferred and proxying is not required.  Only applies to types ClusterIP, NodePort, and LoadBalancer. If this field is specified when creating a Service of type ExternalName, creation will fail. This field will be wiped when updating a Service to type ExternalName.  If this field is not specified, it will be initialized from the clusterIP field.  If this field is specified, clients must ensure that clusterIPs[0] and clusterIP have the same value.\n\nUnless the \"IPv6DualStack\" feature gate is enabled, this field is limited to one value, which must be the same as the clusterIP field.  If the feature gate is enabled, this field may hold a maximum of two entries (dual-stack IPs, in either order).  These IPs must correspond to the values of the ipFamilies field. Both clusterIPs and ipFamilies are governed by the ipFamilyPolicy field. More info: https://kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies",
                "type": "array",
                "items": {
                  "type": "string"
                },
                "x-kubernetes-list-type": "atomic"
              },
              "externalIPs": {
                "description": "externalIPs is a list of IP addresses for which nodes in the cluster will also accept traffic for this service.  These IPs are not managed by Kubernetes.  The user is responsible for ensuring that traffic arrives at a node with this IP.  A common example is external load-balancers that are not part of the Kubernetes system.",
                "type": "array",
                "items": {
                  "type": "string"
                }
              },
              "externalName": {
                "description": "externalName is the external reference that discovery mechanisms will return as an alias for this service (e.g. a DNS CNAME record). No proxying will be involved.  Must be a lowercase RFC-1123 hostname (https://tools.ietf.org/html/rfc1123) and requires `type` to be \"ExternalName\".",
                "type": "string"
              },
              "externalTrafficPolicy": {
                "description": "externalTrafficPolicy denotes if this Service desires to route external traffic to node-local or cluster-wide endpoints. \"Local\" preserves the client source IP and avoids a second hop for LoadBalancer and Nodeport type services, but risks potentially imbalanced traffic spreading. \"Cluster\" obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading.",
                "type": "string"
              },
              "healthCheckNodePort": {
                "description": "healthCheckNodePort specifies the healthcheck nodePort for the service. This only applies when type is set to LoadBalancer and externalTrafficPolicy is set to Local. If a value is specified, is in-range, and is not in use, it will be used.  If not specified, a value will be automatically allocated.  External systems (e.g. load-balancers) can use this port to determine if a given node holds endpoints for this service or not.  If this field is specified when creating a Service which does not need it, creation will fail. This field will be wiped when updating a Service to no longer need it (e.g. changing type).",
                "format": "int32",
                "type": "integer"
              },
              "internalTrafficPolicy": {
                "description": "InternalTrafficPolicy specifies if the cluster internal traffic should be routed to all endpoints or node-local endpoints only. \"Cluster\" routes internal traffic to a Service to all endpoints. \"Local\" routes traffic to node-local endpoints only, traffic is dropped if no node-local endpoints are ready. The default value is \"Cluster\".",
                "type": "string"
              },
              "ipFamilies": {
                "description": "IPFamilies is a list of IP families (e.g. IPv4, IPv6) assigned to this service, and is gated by the \"IPv6DualStack\" feature gate.  This field is usually assigned automatically based on cluster configuration and the ipFamilyPolicy field. If this field is specified manually, the requested family is available in the cluster, and ipFamilyPolicy allows it, it will be used; otherwise creation of the service will fail.  This field is conditionally mutable: it allows for adding or removing a secondary IP family, but it does not allow changing the primary IP family of the Service.  Valid values are \"IPv4\" and \"IPv6\".  This field only applies to Services of types ClusterIP, NodePort, and LoadBalancer, and does apply to \"headless\" services.  This field will be wiped when updating a Service to type ExternalName.\n\nThis field may hold a maximum of two entries (dual-stack families, in either order).  These families must correspond to the values of the clusterIPs field, if specified. Both clusterIPs and ipFamilies are governed by the ipFamilyPolicy field.",
                "type": "array",
                "items": {
                  "type": "string"
                },
                "x-kubernetes-list-type": "atomic"
              },
              "ipFamilyPolicy": {
                "description": "IPFamilyPolicy represents the dual-stack-ness requested or required by this Service, and is gated by the \"IPv6DualStack\" feature gate.  If there is no value provided, then this field will be set to SingleStack. Services can be \"SingleStack\" (a single IP family), \"PreferDualStack\" (two IP families on dual-stack configured clusters or a single IP family on single-stack clusters), or \"RequireDualStack\" (two IP families on dual-stack configured clusters, otherwise fail). The ipFamilies and clusterIPs fields depend on the value of this field.  This field will be wiped when updating a service to type ExternalName.",
                "type": "string"
              },
              "loadBalancerClass": {
                "description": "loadBalancerClass is the class of the load balancer implementation this Service belongs to. If specified, the value of this field must be a label-style identifier, with an optional prefix, e.g. \"internal-vip\" or \"example.com/internal-vip\". Unprefixed names are reserved for end-users. This field can only be set when the Service type is 'LoadBalancer'. If not set, the default load balancer implementation is used, today this is typically done through the cloud provider integration, but should apply for any default implementation. If set, it is assumed that a load balancer implementation is watching for Services with a matching class. Any default load balancer implementation (e.g. cloud providers) should ignore Services that set this field. This field can only be set when creating or updating a Service to type 'LoadBalancer'. Once set, it can not be changed. This field will be wiped when a service is updated to a non 'LoadBalancer' type.",
                "type": "string"
              },
              "loadBalancerIP": {
                "description": "Only applies to Service Type: LoadBalancer LoadBalancer will get created with the IP specified in this field. This feature depends on whether the underlying cloud-provider supports specifying the loadBalancerIP when a load balancer is created. This field will be ignored if the cloud-provider does not support the feature.",
                "type": "string"
              },
              "loadBalancerSourceRanges": {
                "description": "If specified and supported by the platform, this will restrict traffic through the cloud-provider load-balancer will be restricted to the specified client IPs. This field will be ignored if the cloud-provider does not support the feature.\" More info: https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/",
                "type": "array",
                "items": {
                  "type": "string"
                }
              },
              "ports": {
                "description": "The list of ports that are exposed by this service. More info: https://kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies",
                "type": "array",
                "items": {
                  "description": "ServicePort contains information on service's port.",
                  "type": "object",
                  "required": [
                    "port"
                  ],
                  "properties": {
                    "appProtocol": {
                      "description": "The application protocol for this port. This field follows standard Kubernetes label syntax. Un-prefixed names are reserved for IANA standard service names (as per RFC-6335 and http://www.iana.org/assignments/service-names). Non-standard protocols should use prefixed names such as mycompany.com/my-custom-protocol. This is a beta field that is guarded by the ServiceAppProtocol feature gate and enabled by default.",
                      "type": "string"
                    },
                    "name": {
                      "description": "The name of this port within the service. This must be a DNS_LABEL. All ports within a ServiceSpec must have unique names. When considering the endpoints for a Service, this must match the 'name' field in the EndpointPort. Optional if only one ServicePort is defined on this service.",
                      "type": "string"
                    },
                    "nodePort": {
                      "description": "The port on each node on which this service is exposed when type is NodePort or LoadBalancer.  Usually assigned by the system. If a value is specified, in-range, and not in use it will be used, otherwise the operation will fail.  If not specified, a port will be allocated if this Service requires one.  If this field is specified when creating a Service which does not need it, creation will fail. This field will be wiped when updating a Service to no longer need it (e.g. changing type from NodePort to ClusterIP). More info: https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport",
                      "format": "int32",
                      "type": "integer"
                    },
                    "port": {
                      "description": "The port that will be exposed by this service.",
                      "format": "int32",
                      "type": "integer"
                    },
                    "protocol": {
                      "description": "The IP protocol for this port. Supports \"TCP\", \"UDP\", and \"SCTP\". Default is TCP.",
                      "type": "string"
                    },
                    "targetPort": {
                      "description": "IntOrString is a type that can hold an int32 or a string.  When used in JSON or YAML marshalling and unmarshalling, it produces or consumes the inner type.  This allows you to have, for example, a JSON field that can accept a name or number.",
                      "format": "int-or-string",
                      "oneOf": [
                        {
                          "type": "string"
                        },
                        {
                          "type": "integer"
                        }
                      ]
                    }
                  }
                },
                "x-kubernetes-list-map-keys": [
                  "port",
                  "protocol"
                ],
                "x-kubernetes-list-type": "map",
                "x-kubernetes-patch-merge-key": "port",
                "x-kubernetes-patch-strategy": "merge"
              },
              "publishNotReadyAddresses": {
                "description": "publishNotReadyAddresses indicates that any agent which deals with endpoints for this Service should disregard any indications of ready/not-ready. The primary use case for setting this field is for a StatefulSet's Headless Service to propagate SRV DNS records for its Pods for the purpose of peer discovery. The Kubernetes controllers that generate Endpoints and EndpointSlice resources for Services interpret this to mean that all endpoints are considered \"ready\" even if the Pods themselves are not. Agents which consume only Kubernetes generated endpoints through the Endpoints or EndpointSlice resources can safely assume this behavior.",
                "type": "boolean"
              },
              "selector": {
                "description": "Route service traffic to pods with label keys and values matching this selector. If empty or not present, the service is assumed to have an external process managing its endpoints, which Kubernetes will not modify. Only applies to types ClusterIP, NodePort, and LoadBalancer. Ignored if type is ExternalName. More info: https://kubernetes.io/docs/concepts/services-networking/service/",
                "type": "object",
                "additionalProperties": {
                  "type": "string"
                }
              },
              "sessionAffinity": {
                "description": "Supports \"ClientIP\" and \"None\". Used to maintain session affinity. Enable client IP based session affinity. Must be ClientIP or None. Defaults to None. More info: https://kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies",
                "type": "string"
              },
              "sessionAffinityConfig": {
                "description": "SessionAffinityConfig represents the configurations of session affinity.",
                "type": "object",
                "properties": {
                  "clientIP": {
                    "description": "ClientIPConfig represents the configurations of Client IP based session affinity.",
                    "type": "object",
                    "properties": {
                      "timeoutSeconds": {
                        "description": "timeoutSeconds specifies the seconds of ClientIP type session sticky time. The value must be >0 && <=86400(for 1 day) if ServiceAffinity == \"ClientIP\". Default value is 10800(for 3 hours).",
                        "format": "int32",
                        "type": "integer"
                      }
                    }
                  }
                }
              },
              "topologyKeys": {
                "description": "topologyKeys is a preference-order list of topology keys which implementations of services should use to preferentially sort endpoints when accessing this Service, it can not be used at the same time as externalTrafficPolicy=Local. Topology keys must be valid label keys and at most 16 keys may be specified. Endpoints are chosen based on the first topology key with available backends. If this field is specified and all entries have no backends that match the topology of the client, the service has no backends for that client and connections should fail. The special value \"*\" may be used to mean \"any topology\". This catch-all value, if used, only makes sense as the last value in the list. If this is not specified or empty, no topology constraints will be applied. This field is alpha-level and is only honored by servers that enable the ServiceTopology feature. This field is deprecated and will be removed in a future version.",
                "type": "array",
                "items": {
                  "type": "string"
                }
              },
              "type": {
                "description": "type determines how the Service is exposed. Defaults to ClusterIP. Valid options are ExternalName, ClusterIP, NodePort, and LoadBalancer. \"ClusterIP\" allocates a cluster-internal IP address for load-balancing to endpoints. Endpoints are determined by the selector or if that is not specified, by manual construction of an Endpoints object or EndpointSlice objects. If clusterIP is \"None\", no virtual IP is allocated and the endpoints are published as a set of endpoints rather than a virtual IP. \"NodePort\" builds on ClusterIP and allocates a port on every node which routes to the same endpoints as the clusterIP. \"LoadBalancer\" builds on NodePort and creates an external load-balancer (if supported in the current cloud) which routes to the same endpoints as the clusterIP. \"ExternalName\" aliases this service to the specified externalName. Several other fields do not apply to ExternalName services. More info: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types",
                "type": "string"
              }
            }
          },
          "status": {
            "description": "ServiceStatus represents the current status of a service.",
            "type": "object",
            "properties": {
              "conditions": {
                "description": "Current service state",
                "type": "array",
                "items": {
                  "description": "Condition contains details for one aspect of the current state of this API Resource.",
                  "type": "object",
                  "required": [
                    "type",
                    "status",
                    "lastTransitionTime",
                    "reason",
                    "message"
                  ],
                  "properties": {
                    "lastTransitionTime": {
                      "description": "Time is a wrapper around time.Time which supports correct marshaling to YAML and JSON.  Wrappers are provided for many of the factory methods that the time package offers.",
                      "format": "date-time",
                      "type": "string"
                    },
                    "message": {
                      "description": "message is a human readable message indicating details about the transition. This may be an empty string.",
                      "type": "string"
                    },
                    "observedGeneration": {
                      "description": "observedGeneration represents the .metadata.generation that the condition was set based upon. For instance, if .metadata.generation is currently 12, but the .status.conditions[x].observedGeneration is 9, the condition is out of date with respect to the current state of the instance.",
                      "format": "int64",
                      "type": "integer"
                    },
                    "reason": {
                      "description": "reason contains a programmatic identifier indicating the reason for the condition's last transition. Producers of specific condition types may define expected values and meanings for this field, and whether the values are considered a guaranteed API. The value should be a CamelCase string. This field may not be empty.",
                      "type": "string"
                    },
                    "status": {
                      "description": "status of the condition, one of True, False, Unknown.",
                      "type": "string"
                    },
                    "type": {
                      "description": "type of condition in CamelCase or in foo.example.com/CamelCase.",
                      "type": "string"
                    }
                  }
                },
                "x-kubernetes-list-map-keys": [
                  "type"
                ],
                "x-kubernetes-list-type": "map",
                "x-kubernetes-patch-merge-key": "type",
                "x-kubernetes-patch-strategy": "merge"
              },
              "loadBalancer": {
                "description": "LoadBalancerStatus represents the status of a load-balancer.",
                "type": "object",
                "properties": {
                  "ingress": {
                    "description": "Ingress is a list containing ingress points for the load-balancer. Traffic intended for the service should be sent to these ingress points.",
                    "type": "array",
                    "items": {
                      "description": "LoadBalancerIngress represents the status of a load-balancer ingress point: traffic intended for the service should be sent to an ingress point.",
                      "type": "object",
                      "properties": {
                        "hostname": {
                          "description": "Hostname is set for load-balancer ingress points that are DNS based (typically AWS load-balancers)",
                          "type": "string"
                        },
                        "ip": {
                          "description": "IP is set for load-balancer ingress points that are IP based (typically GCE or OpenStack load-balancers)",
                          "type": "string"
                        },
                        "ports": {
                          "description": "Ports is a list of records of service ports If used, every port defined in the service should have an entry in it",
                          "type": "array",
                          "items": {
                            "type": "object",
                            "required": [
                              "port",
                              "protocol"
                            ],
                            "properties": {
                              "error": {
                                "description": "Error is to record the problem with the service port The format of the error shall comply with the following rules: - built-in error values shall be specified in this file and those shall use\n  CamelCase names\n- cloud provider specific error values must have names that comply with the\n  format foo.example.com/CamelCase.",
                                "type": "string"
                              },
                              "port": {
                                "description": "Port is the port number of the service port of which status is recorded here",
                                "format": "int32",
                                "type": "integer"
                              },
                              "protocol": {
                                "description": "Protocol is the protocol of the service port of which status is recorded here The supported values are: \"TCP\", \"UDP\", \"SCTP\"",
                                "type": "string"
                              }
                            }
                          },
                          "x-kubernetes-list-type": "atomic"
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        },
        "x-kubernetes-group-version-kind": [
          {
            "group": "",
            "kind": "Service",
            "version": "v1"
          }
        ]
      },
      "translations": {
        "en": {
          "name": "SERVICE NEW",
          "category": "DD",
          "spec.type": "Service type",
          "spec.ports": "Ports //TODO",
          "spec.clusterIP": "Cluster IP",
          "spec.externalIPs": "External IPs //TODO",
          "metadata.ownerReferences": "Controlled by"
        }
      }
    },
    "deployments": {
      "resource": {
        "kind": "Deployment",
        "group": "apps",
        "version": "v1",
        "scope": "namespace",
        "path": "deployments-new"
      },
      "details": {
        "header": [
          {
            "path": "$serviceWithTheSameName",
            "name": "Service with the same name",
            "valuePreprocessor": "PendingWrapper",
            "formula": "data.metadata.name & ' (of IP:' & data.spec.clusterIP & ')'"
        }],
        "body": [{
            "widget": "ResourceList",
            "path": "$pods",
            "name": "Me pods",
            "disableCreate": true
        }]
      },
      "relations": {
        "pods": {
          "kind": "Pod",
          "group": "api",
          "version": "v1",
          "ownerLabelSelectorPath": "spec.selector.matchLabels"
        },
        "serviceWithTheSameName": {
          "kind": "Service",
          "group": "api",
          "version": "v1",
          "resourceName": "same-name"
        }
      }
    }
  }
```
</details>

**Related issue(s)**

Closes #1471


core-ui/src/shared/hooks/useMounted.js
```diff
@@ -0,0 +1,13 @@
...
+// prevents `setState` on onmounted component
+export function useMounted() {
+  const mounted = useRef(false);
+
+  useEffect(() => {
+    mounted.current = true;
+    return () => (mounted.current = false);
+  }, []);
+
+  return mounted;
```
wchrapka, line 12: There;s a hook for that already, the filename is 'useComponentDidMount'

Wawrzyn321: Actually it's not the same? It doesn't return "mountedness" status (which is required here), it just runs a function once.

wchrapka: It's a fairly similar functionality. Why is it better to have it separated?

wchrapka:
Links to controlling resources don't work:
image

wchrapka (approved):

wchrapka (approved):

docs/extensibility/display-widgets.md
```diff
@@ -125,6 +125,27 @@ Table widgets display array data as rows of a table instead of free-standing com
 
 <img src="./assets/display-widgets/Table.png" alt="Example of a table widget" style="border: 1px solid #D2D5D9">
 
+### ResourceList
+
+ResourceList widgets render a list of Kubernetes resources. It should be used along with [related resources](resources.md#relations-section).
```
NHingerl, line 130: What "it"? The only singular word in the previous sentence is "a list", so is that what should be used along with related resources? 

For clarification, I suggest to replace "it" with the actual word, not sure which one it is:

```suggestion
ResourceList widgets render a list of Kubernetes resources. The resource list should be used along with [related resources](resources.md#relations-section).
```
or 

```suggestion
ResourceList widgets render a list of Kubernetes resources. The ResourceList widgets should be used along with [related resources](resources.md#relations-section).
```

```suggestion
ResourceList widgets render a list of Kubernetes resources. The Kubernetes resources should be used along with [related resources](resources.md#relations-section).
```

docs/extensibility/display-widgets.md
```diff
@@ -125,6 +125,27 @@ Table widgets display array data as rows of a table instead of free-standing com
...
+
+If such resource list was already defined in Busola, the configuration will be reused. To obtain custom columns, specify `columns` field.
```
NHingerl, line 132:
```suggestion
If such resource list was already defined in Busola, the configuration will be reused. To obtain custom columns, specify the `columns` field.
```

docs/extensibility/resources.md
```diff
@@ -176,6 +184,59 @@ renders the same set of data as:
 ]
 ```
 
+## relations section
+
+This section contains an object that maps a relation name to relation configuration object. Relation name preceded by dollar sign '\$' can be then used in `path` expression.
```
NHingerl, line 189:
```suggestion
The `relations` section contains an object that maps a relation name to a relation configuration object. The relation name preceded by a dollar sign '\$' can be then used in the `path` expression.
```

"Can" it be used, or "is" it used? Does "can" imply it's optional or there are other alternatives? Maybe we can drop the "can".

docs/extensibility/resources.md
```diff
@@ -176,6 +184,59 @@ renders the same set of data as:
...
+
+It's possible to use both relation name and a path, e.g. `{"path": $myRelatedResource.metadata.labels}` will return the `metadata.labels` of related resource.
```
NHingerl, line 191:
```suggestion
It's possible to use both relation name and a path; for example, `{"path": $myRelatedResource.metadata.labels}` returns the `metadata.labels` of the related resource.
```

docs/extensibility/resources.md
```diff
@@ -176,6 +184,59 @@ renders the same set of data as:
...
+
+### Relation configuration object fields
+
+Those fields are used to build the related resource URL and filter the received data.
+
+- **kind** - _[required]_ Kubernetes resource kind.
+- **group** - _[required]_ Kubernetes resource group.
+- **version** - _[required]_ Kubernetes resource version.
+- **namespace** - resource's Namespace name, defaults to original resource's Namespace. If set to `null` the relation with match clusterwide resources / resources in all namespaces.
```
NHingerl, line 200:
```suggestion
- **namespace** - the resource's Namespace name; it defaults to the original resource's Namespace. If set to `null`, the relation matches cluster-wide resources or resources in all Namespaces.
```

I'm guessing "with" is a typo for "will", and I'd say we can replace future tense with present tense here.

docs/extensibility/resources.md
```diff
@@ -176,6 +184,59 @@ renders the same set of data as:
...
+- **resourceName** - specific resource name, leave empty to match all resources of given type.
```
NHingerl, line 201:
```suggestion
- **resourceName** - a specific resource name; leave empty to match all resources of a given type.
```

docs/extensibility/resources.md
```diff
@@ -176,6 +184,59 @@ renders the same set of data as:
...
+- **ownerLabelSelectorPath** - path to original object's `selector` type property (e.g. `spec.selector.matchLabels` for Deployment, used to select matching Pods).
```
NHingerl, line 202:
```suggestion
- **ownerLabelSelectorPath** - the path to original object's `selector` type property; for example, `spec.selector.matchLabels` for Deployment, used to select matching Pods.
```

docs/extensibility/resources.md
```diff
@@ -176,6 +184,59 @@ renders the same set of data as:
...
+- **selector** - [JSONata](https://docs.jsonata.org/overview.html) function enabling user to write a custom matching logic. It receives a data context of:
```
NHingerl, line 203:
```suggestion
- **selector** - [JSONata](https://docs.jsonata.org/overview.html) function enabling the user to write a custom matching logic. It receives a data context of:
```

docs/extensibility/resources.md
```diff
@@ -232,3 +293,17 @@ spec:
...
+
+### Value preprocessors
+
+Value preprocessors are used as a middleware between a value and the actual renderer. They can transform a given value and pass it further or stop processing, rendering to view early.
+
+#### List of value preprocessors
+
+- **PendingWrapper** - useful when value resolves to a triple of `{loading, error, data}`:
+
+  - For `loading` equal to `true`, it will display a loading indicator.
```
NHingerl, line 305:
```suggestion
  - For `loading` equal to `true`, it displays a loading indicator.
```

docs/extensibility/resources.md
```diff
@@ -232,3 +293,17 @@ spec:
...
+  - For truthy `error`, it will display an error message.
```
NHingerl, line 306:
```suggestion
  - For truthy `error`, it displays an error message.
```

docs/extensibility/resources.md
```diff
@@ -232,3 +293,17 @@ spec:
...
+  - Otherwise, it will pass `data` to display component.
```
NHingerl, line 307:
```suggestion
  - Otherwise, it passes `data` to the display component.
```

docs/extensibility/resources.md
```diff
@@ -232,3 +293,17 @@ spec:
...
+
+  Unless custom handling of error or loading state is required, using the `PendingWrapper` is recommended, e.g. for fields that use [related resources](#relations-section).
```
NHingerl, line 309:
```suggestion
  Unless you need custom handling of error or loading state, we recommend using the `PendingWrapper`, for example, for fields that use [related resources](#relations-section).
```

docs/extensibility/resources.md
```diff
@@ -232,3 +293,17 @@ spec:
 ```
 
 If you provide both `translations` and `translations-{lang}` sections, they are merged together.
+
+### Value preprocessors
+
+Value preprocessors are used as a middleware between a value and the actual renderer. They can transform a given value and pass it further or stop processing, rendering to view early.
```
NHingerl, line 299:
```suggestion
Value preprocessors are used as a middleware between a value and the actual renderer. They can transform a given value and pass it to the widget; or stop processing and render it so you can view it immediately without passing it to the widget.
```

docs/extensibility/resources.md
NHingerl: Alright! Thank you for clarification :)

wchrapka (approved):

PR #1516, opened 2022-06-29
Resource widget

qbalukom:


akucharska (approved):

docs/extensibility/form-widgets.md
```diff
@@ -53,6 +53,41 @@ CodeEditor widgets render a versatile code editor that can be used to edit any v
 }
 ```
 
+### Resource
+
+Resource widgets render a dropdown list of specified resources and stores the selected one as a string containing it's name.
```
majakurcius, line 58:
```suggestion
Resource widgets render a dropdown list of specified resources and store the selected one as a string containing its name.
```
- `stores` -> `store`: because `resource widgets` is plural
- `it's` = it is; `its` = belonging to **it**

docs/extensibility/form-widgets.md
```diff
@@ -53,6 +53,41 @@ CodeEditor widgets render a versatile code editor that can be used to edit any v
...
+
+#### Widget-specific parameters
+
+- **kind** - _[required]_ Kubernetes kind of the resource.
+- **group** - API group used for all requests. Not provided for native kubernetes resources.
```
majakurcius, line 63:
```suggestion
- **group** - API group used for all requests. Not provided for native Kubernetes resources.
```

docs/extensibility/form-widgets.md
```diff
@@ -53,6 +53,41 @@ CodeEditor widgets render a versatile code editor that can be used to edit any v
...
+- **version** - _[required]_ API version used for all requests.
+- **scope** - either `namespace` or `cluster`. When set to cluster namespaced resources will be fetched from all namespaces. Defaults to `cluster`.
```
majakurcius, line 65:
```suggestion
- **scope** - either `namespace` or `cluster`. When set to `cluster`, namespaced resources are fetched from all Namespaces. Defaults to `cluster`.
```

docs/extensibility/form-widgets.md
```diff
@@ -53,6 +53,41 @@ CodeEditor widgets render a versatile code editor that can be used to edit any v
...
+- **namespace** - namespace to fetch resources from. Used only when scope is `namespace` and resources need to be fetched from a specific namespace. Defaults to active namespace when omitted.
```
majakurcius, line 66:
```suggestion
- **namespace** - Namespace to fetch resources from. Used only when scope is `namespace` and resources need to be fetched from a specific Namespace. Defaults to the active Namespace when omitted.
```

majakurcius, line 66: Consider making these parameters a table, like this:

```
| **Parameter** | **Description** |
|---------------|-----------------|
| **Example param** | Example description |
```

qbalukom: Worth considering, but if so it would be better to change for all widgets - so I'd rather leave it as a separate documentation improvement task (only as an investigation, as I'm not actually sure if it's a good idea).

majakurcius (approved):

akucharska (approved):

PR #1519, opened 2022-06-30
Fix monaco editor height

### LINKED ISSUE #1447: Yaml Editor is small not only in safari
**Description**

When importing yaml in Safari, the yaml editor looks smaller than Chrome

**Expected result**

The size should be the same

**Actual result**

**Steps to reproduce**

**Troubleshooting**

image

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Fixes #1447


core-ui/src/shared/components/MonacoEditorESM/Editor.scss
```diff
@@ -1,9 +1,8 @@
 .resource-form__wrapper {
   display: flex;
   flex-direction: column;
-  height: calc(100% - 21px - 36px);
+  flex-grow: 2;
```
qbalukom, line 4: why 2? i don't think it is in relation with other `flex-glow` items? or is it? `1` would be sufficient here.
```suggestion
flex-grow: 1;
```

core-ui/src/shared/components/MonacoEditorESM/Editor.scss
```diff
@@ -1,9 +1,8 @@
 .resource-form__wrapper {
```
qbalukom, line 1: whoever wrote this file didn't scope these rules for just the editor, so we're changing all wrappers in all the forms. Perhaps we could add scoping while we're here? and make sure we don't break anything else along the way?
or if not, create a separate task for this - we're having too many cases of similar situations.

qbalukom:
This breaks the wizard layouting and makes the whole windows scrollable instead of wizard content (so that wizard header and footer disappear when the window is shorter now).
image

qbalukom (changes requested):
(see the single comments)

qbalukom (approved):

PR #1533, opened 2022-07-01
EXT: custom lang for CodeViewer

### LINKED ISSUE #1531: Recover language parameter in CodeViewer widget
**Description**

Re-add the feature to choose formatting language for `CodeViewer` widget.
Keep the current language detection (is it the same detection as is/was used in config maps? maybe use a single source for language detection?)
Re-add the optional `language` parameter that overrides any detection when provided.

The original parameter got removed in https://github.com/kyma-project/busola/pull/1430 (might not have been 100% implemented).

**Note**
Include information what languages are supported - potentially simply as a link to monaco.

wchrapka:
**Description**

A user can choose a custom language for the EXT CodeViewer component.

**Related issue(s)**
Fixes #1531


core-ui/src/components/Extensibility/components/CodeViewer.js
```diff
@@ -19,26 +19,38 @@ export function CodeViewer({ value, structure, schema }) {
...
+    switch (language) {
+      case 'yaml':
+        parsedValue = jsyaml.dump(value);
+        break;
+      default:
+        //this includes JSON and other languages
+        parsedValue = stringifyIfObject(value);
+    }
+  } catch (e) {
+    window.alert(
```
dariadomagala, line 49: Do we really want to show a window alert for it? Could we just show a regular notification instead?

docs/extensibility/display-widgets.md
```diff
@@ -96,14 +96,19 @@ Columns widgets render the child widgets in two columns.
 
 ### CodeViewer
 
-CodeViewer widgets display values using a read-only code editor. The editor autodetects the language.
+CodeViewer widgets display values using a read-only code editor.
+
+#### Widget-specific parameters
+
+- **language** - used for code highlighting. Editor supports languages handled by [Monaco](https://code.visualstudio.com/docs/languages/overview) and autodetects `yaml` and `json`.
```
dariadomagala, line 103: it automatically sets the language to `yaml` also for jsons so it's misleading

wchrapka: Do you suggest no default language? YAML is superset of JSON, so it can understand all JSONs, but not the other way around.

core-ui/src/components/Extensibility/components/CodeViewer.js
wchrapka: Fixed

core-ui/src/components/Extensibility/components/CodeViewer.js
```diff
@@ -46,9 +47,10 @@ const getValueAndLang = (value, structure) => {
         parsedValue = stringifyIfObject(value);
     }
   } catch (e) {
-    window.alert(
-      `CodeViewer: received resource is in a language other than defined: ${e}`,
-    );
+    LuigiClient.uxManager().showAlert({
+      text: `CodeViewer - the received resource is in a language other than defined: ${e}`,
```
dariadomagala, line 51: 1. Please use `useNotification`. We use luigi notifications only for notifications from `core`, where we don't have the mechanism from `core-ui`.
2. Please move the text to the translations file.

dariadomagala (approved):

dariadomagala (approved):

NHingerl (approved):

PR #1543, opened 2022-07-05
expandable table

qbalukom:


core-ui/src/shared/components/GenericList/GenericList.scss
```diff
@@ -36,6 +36,7 @@
     }
   }
 
+  /*
```
akucharska: If this is not necessary then please remove it

akucharska (commented):
It would be good to address the issue with the `margin` at the `columns` component. If not now then create an issue for it.
Screen Shot 2022-07-05 at 11 07 01
Screen Shot 2022-07-05 at 11 06 48

core-ui/src/shared/components/GenericList/GenericList.scss
qbalukom: already did :)

akucharska (dismissed):

akucharska (approved):

NHingerl (approved):

PR #1568, opened 2022-07-07
Fix kubeconfigID bugs

### LINKED ISSUE #1437: FEATURE REQUEST: redirect to clusters overview after importing a multi-context kubeconfig file
**Description**

It would be great to have the option to show the cluster overview page after importing a multi-context kubeconfig file via the `kubeconfigID`.  

Maybe it is also viable to add this as a configuration option, as part of the `kubeconfigID` configuration? In that case it could be decoupled from the multi-context use-case and be a stand-alone addition to the kubeconfigID feature.  
Maybe something like this:

```
"KUBECONFIG_ID": {
  "config": {
    "showClusterOverview": "true"
  }
},
```

**Reasons**

We would like our users to be able to access the busola dashboard through one url and be able to choose which cluster to connect to from there.  Currently, the user would be connected to their current context, which would create extra friction and might confuse some users.

Think of the following user flow:

1. User visits busola.example.com/?kubeconfigID=multicontext
2. Busola loads all clusters from this kubeconfig
3. The user is redirected to the clusters overview page and can choose which cluster they want to connect to

### LINKED ISSUE #1490: App may freeze when logging throufh kubeconfigID
**Description**

When logging to a cluster by a get param `kubeconfigID`, the application can freeze after a successful login. It's 50/50 this will happen. Looks a bit like listeners are not getting connected or there's a loop. Refresh resolves the problem

**Expected result**

Application never freezes. 

**Actual result**

Application sometimes freezes.

**Steps to reproduce**

Use prod setup if possible or change KUBECONFIG_ID to a local path and create a file you'll access ?kubeconfigID=file.yaml

```
    "KUBECONFIG_ID": {
        "config": {
          "kubeconfigUrl": "assets/kubeconfig"
        }
      },
```
Visit the app with ?kubeconfigID=multi.yaml param appended to the url

**Troubleshooting**

### LINKED ISSUE #1527: Switching clusters added through kubeconfigID
**Description**

React breaks (blank page) if you add a cluster through `kubeconfigID` and then go to cluster overview.

Conntected to #1490 

image

**Expected result**
React doesn't break.

**Actual result**
React breaks.

**Steps to reproduce**
1. Add a cluster through the interface
2. Add another cluster through kubeconfigID
3. Go to cluster list
4. If error doesn't emerge, switch clusters one more time

**Troubleshooting**

Since we're going to remove luigi maybe, maybe there's a way to add clusters through kubeconfigID using React handlers. Currently Luigi handles `kubeconfigID` and React handles new registrations through UI. Maybe it's as easy as adding an optional chaining operator `something?.localCompare()`, but this part of app is probelmatic in general.

wchrapka:
**Description**

Moves the kubeconfigID login logic to the React app, don't redirect the app if there are multiple clusters.

**Related issue(s)**

Fixes #1527, fixes #1490, resolves #1437

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -229,7 +229,7 @@ Note that some features must be run before the application starts the bootstrap 
   ```
 
 * **KUBECONFIG_ID** – is used to configure the URL to which Busola sends a request to download a kubeconfig file. If you add `?kubeconfigID={your ID}` to the Busola URL, Busola tries to download the kubeconfig from `{kubeconfigUrl}/{yourID}`. If the operation succeeds, Busola adds the kubeconfing file to the cluster.
-  If you use a full address in the **kubeconfigUrl** field, Busola also reads it. Optionally, instruct Busola to show the clusters overview rather than the current context cluster with the **showClustersOverview** property.
+  If you use a full address in the **kubeconfigUrl** field, Busola also reads it. Optionally, you can instruct Busola to show the clusters overview rather than the current context cluster, using the **showClustersOverview** property.
 
   Default settings:
 


Wawrzyn321 (dismissed):

docs/features.md
```diff
@@ -229,15 +229,16 @@ Note that some features must be run before the application starts the bootstrap
   ```
 
 * **KUBECONFIG_ID** – is used to configure the URL to which Busola sends a request to download a kubeconfig file. If you add `?kubeconfigID={your ID}` to the Busola URL, Busola tries to download the kubeconfig from `{kubeconfigUrl}/{yourID}`. If the operation succeeds, Busola adds the kubeconfing file to the cluster.
-  If you use a full address in the **kubeconfigUrl** field, Busola also reads it.
+  If you use a full address in the **kubeconfigUrl** field, Busola also reads it. Optionally, instruct Busola to show the clusters overview rather than the current context cluster with the **showClustersOverview** property.
```
grego952, line 232:
```suggestion
  If you use a full address in the **kubeconfigUrl** field, Busola also reads it. Optionally, you can instruct Busola to show the clusters overview rather than the current context cluster, using the **showClustersOverview** property.
```

Wawrzyn321 (approved):

grego952 (approved):

PR #1573, opened 2022-07-08
Add first session & cluster change tracking

Wawrzyn321:
**Description**

- Also fix tiny bug on Application details
- When someone doesn't use our frontend BUT uses backend - we don't use `/backend/tracking` endpoint. Instead I've added possibility to increate logs output by setting the TRACKING flag.

**Testing**

Remember to enable TRACKING feature in `defaultConfig.json` (on frontend and probably on backend).

**Related issue(s)**

### DIFF
--- a/core-ui/src/hooks/useTracking.js
+++ b/core-ui/src/hooks/useTracking.js
@@ -1,76 +1,50 @@
 import { useEffect } from 'react';
+import LuigiClient from '@luigi-project/client';
 import { useLocation } from 'react-router-dom';
-import { baseUrl } from 'shared/hooks/BackendAPI/config';
-import { useConfig } from 'shared/contexts/ConfigContext';
-import { useFeature } from 'shared/hooks/useFeature';
-import { v4 as uuid } from 'uuid';
 
-function getLoggingId() {
-  const STORAGE_KEY = 'busola.logging-id';
-
-  let loggingId = localStorage.getItem(STORAGE_KEY);
-  if (!loggingId) {
-    loggingId = uuid();
-    localStorage.setItem(STORAGE_KEY, loggingId);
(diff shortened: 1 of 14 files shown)


core-ui/src/hooks/useTracking.js
```diff
@@ -1,53 +1,56 @@
...
+
+  useEffect(() => {
+    sendTrackingRequest({
+      event: 'SESSION_START',
+      data: {
+        hostname: window.location.hostname,
+        apiServerAddress: cluster?.cluster.server || null,
+      },
+    });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
wchrapka, line 54: I'm not sure about the intention. If this should change when a user switches a cluster, then maybe add `cluster?.cluster.server` to the dep array. If this fires only once per the app life, they maybe use the `useComponentDidMount` hook?

core/src/luigi-config/cluster-management/cluster-management.js
```diff
@@ -53,6 +54,13 @@ export function getAfterLoginLocation(clusterName, kubeconfig) {
   }`;
 }
 
+async function trackClusterChange(params) {
```
wchrapka, line 57: this will be sent even if the tracking is disabled

On a second note, React is well aware of the cluster change, why not fire it from React when a microcontext changes?

core-ui/src/hooks/useTracking.js
Wawrzyn321: This thing should fire only once, when the app is loaded - so it's fine, but I probably merge it with `trackClusterChange` from `core`.

wchrapka (approved):

PR #1584, opened 2022-07-12
Extensibility: Details view: Hide key if the value is empty from panels

### LINKED ISSUE #1539: Extensibility: Details view: Hide key if the value is empty from panels
**Description**
I want to be able to hide the full entries from panels or other components when the value is empty. So instead of showing a key with an empty value to the right side, it would not show the key at all. I've attached 2 screenshots below to explain what I mean:

**Current**
image

**Wanted**
image

**Reasons**
We want to hide fields which are empty and not relevant to the user when viewing our resources. There is also the case that we have fields which are mutually exclusive so that only one of the two will ever be there and we want to only display the relevant one.

Wawrzyn321:
**Description**

- As in issue.
- Add more tests

**Related issue(s)**

Closes #1539

### DIFF
--- a/core-ui/src/setupTests.js
+++ b/core-ui/src/setupTests.js
@@ -53,6 +53,24 @@ jest.mock('graphviz-react', () => ({
   Graphviz: () => 'Graphviz mock',
 }));
 
+jest.mock('react-i18next', () => ({
+  // this mock makes sure any components using the translate hook can use it without a warning being shown
+  useTranslation: () => {
+    return {
+      t: key => {
+        if (Array.isArray(key)) {
+          return key[0];
+        }
+        return key;
+      },
+      i18n: {
+        changeLanguage: () => new Promise(() => {}),
+        options: {},
+      },
+    };
+  },
+}));
+
 Enzyme.configure({ adapter: new Adapter() });
 
 export async function expectToSolveWithin(
--- a/core-ui/src/shared/components/ResourceDetails/ResourceDetails.js
+++ b/core-ui/src/shared/components/ResourceDetails/ResourceDetails.js
@@ -311,11 +311,23 @@ function Resource({
           />
         </PageHeader.Column>
 
-        {customColumns.map(col => (
-          <PageHeader.Column key={col.header} title={col.header}>
-            {col.value(resource)}
-          </PageHeader.Column>
-        ))}
+        {customColumns
+          .filter(col => {
+            const { visible, error } = col.visibility?.(resource) || {
+              visible: true,
+            };
+            if (error) {
+              col.value = () =>
+                t('common.messages.error', { error: error.message });
+              return true;
+            }
+            return visible;
+          })
+          .map(col => (
+            <PageHeader.Column key={col.header} title={col.header}>
+              {col.value(resource)}
+            </PageHeader.Column>
+          ))}
       </PageHeader>
       <DeleteMessageBox resource={resource} resourceUrl={resourceUrl} />
       {(customComponents || []).map(component =>
--- a/core-ui/src/shared/components/ResourceDetails/ResourceDetails.js
+++ b/core-ui/src/shared/components/ResourceDetails/ResourceDetails.js
@@ -290,6 +290,17 @@ function Resource({
     }
   };
 
+  const filterColumns = col => {
+    const { visible, error } = col.visibility?.(resource) || {
+      visible: true,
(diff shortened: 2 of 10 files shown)


docs/extensibility/resources.md
```diff
@@ -106,6 +106,10 @@ The `details` section defines the display structure for the details page. It con
 - **widget** - optional widget to render the defined entry. By default the value is displayed verbatim. For more information about the available widgets, see [Display widgets](display-widgets.md).
 - **valuePreprocessor** - name of [value preprocessor](#value-preprocessors),
 - **formula** - optional formula used to modify data referred to by the `path` property. To learn more about using formulas, see [JSONata](https://docs.jsonata.org/overview.html).
+- **visibility** - by default all fields are visible; however `visibility` property can be used to control a single item display.
+  - If set to `false` explicitly, the field won't render.
```
grego952, line 110:
```suggestion
  - If set to `false` explicitly, the field doesn't render.
```

docs/extensibility/resources.md
```diff
@@ -106,6 +106,10 @@ The `details` section defines the display structure for the details page. It con
...
+  - If set to any string, this property will be treated as jsonata formula, determining (based on current value given as `data`) if the field should be visible.
```
grego952, line 111:
```suggestion
  - If set to any string, this property is treated as jsonata formula, determining (based on current value given as `data`) if the field should be visible.
```

docs/extensibility/resources.md
```diff
@@ -106,6 +106,10 @@ The `details` section defines the display structure for the details page. It con
...
+  - If not set the field will always be rendered.
```
grego952, line 112:
```suggestion
  - If not set, the field always renders.
```

docs/extensibility/resources.md
```diff
@@ -106,6 +106,10 @@ The `details` section defines the display structure for the details page. It con
 - **widget** - optional widget to render the defined entry. By default the value is displayed verbatim. For more information about the available widgets, see [Display widgets](display-widgets.md).
 - **valuePreprocessor** - name of [value preprocessor](#value-preprocessors),
 - **formula** - optional formula used to modify data referred to by the `path` property. To learn more about using formulas, see [JSONata](https://docs.jsonata.org/overview.html).
+- **visibility** - by default all fields are visible; however `visibility` property can be used to control a single item display.
```
grego952, line 109:
```suggestion
- **visibility** - by default all fields are visible; however **visibility** property can be used to control a single item display.
```
if visibility is a parameter it should be bolded

docs/extensibility/resources.md
```diff
@@ -106,6 +106,10 @@ The `details` section defines the display structure for the details page. It con
 - **widget** - optional widget to render the defined entry. By default the value is displayed verbatim. For more information about the available widgets, see [Display widgets](display-widgets.md).
 - **valuePreprocessor** - name of [value preprocessor](#value-preprocessors),
 - **formula** - optional formula used to modify data referred to by the `path` property. To learn more about using formulas, see [JSONata](https://docs.jsonata.org/overview.html).
+- **visibility** - by default all fields are visible; however **visibility** property can be used to control a single item display.
```
wchrapka, line 109: JSONata tested successfully for Panel children, doesn't work for other components:
CodeViewer - for `"path": "spec.files",` requires `"visibility": "$exists(data.spec.files)"`
Text - for `"path": "kind"` requires `"visibility": "$exists(data.kind)"`
same for the widgets in the header
etc.

core-ui/src/setupTests.js
```diff
@@ -53,6 +53,24 @@ jest.mock('graphviz-react', () => ({
   Graphviz: () => 'Graphviz mock',
 }));
```
wchrapka, line 54: @dariadomagala is working on the tests, maybe she wants a say.

core-ui/src/shared/components/ResourceDetails/ResourceDetails.js
```diff
@@ -290,6 +290,17 @@ function Resource({
     }
   };
 
+  const filterColumns = col => {
+    const { visible, error } = col.visibility?.(resource) || {
```
wchrapka, line 294: didn't we agree we don't need it for columns?

wchrapka (approved):

grego952 (approved):

PR #1591, opened 2022-07-13
add feature defaultKubeconfig

### LINKED ISSUE #1438: FEATURE REQUEST: Set default kubeconfig to load with kubeconfigID feature
**Description**

Currently, the `kubeconfigID` feature doesn't seem to have a way of defining a default kubeconfig to load.  
This means, that no kubeconfig is loaded when accessing `busola.example.com/`.  

It would be great to have the option to define a default to load when nothing else is specified. Probably it would make sense to add it to the `kubeconfigID` feature configuration as such:

```
"KUBECONFIG_ID": {
  "config": {
    "defaultKubeconfig": "supersecretkubeconfig"
  }
},
```

**Reasons**

This is somewhat related to https://github.com/kyma-project/busola/issues/1437. We would like to provide the user with a single URL to access the dashboard. Since the `kubeconfigID` feature loads the kubeconfig and then removes the parameter from the URL, any bookmarks that are created after the initial load would not be correct and lead to confusion the next visit (or the very latest after clearing local data).

wchrapka:
**Description**

Introduces a new feature to kubeconfigID -> the defaultKubeconfig value is used when there are no clusters in busola and a user visits the home page: `/`.

**Related issue(s)**
Resolves #1438


KaiReichart:
Hey @wchrapka, thanks for the great work! I tried it already, works great!

May I add 1 request? It would be great to also have this feature enabled on the /clusters path, since this is the path users would most likely add to their bookmarks. I understand this might have make the thing a bit more complicated.  ..

akucharska (approved):

docs/features.md
```diff
@@ -229,7 +229,11 @@ Note that some features must be run before the application starts the bootstrap
   ```
 
 * **KUBECONFIG_ID** – is used to configure the URL to which Busola sends a request to download a kubeconfig file. If you add `?kubeconfigID={your ID}` to the Busola URL, Busola tries to download the kubeconfig from `{kubeconfigUrl}/{yourID}`. If the operation succeeds, Busola adds the kubeconfing file to the cluster.
-  If you use a full address in the **kubeconfigUrl** field, Busola also reads it. Optionally, you can instruct Busola to show the clusters overview rather than the current context cluster, using the **showClustersOverview** property.
+  If you use a full address in the **kubeconfigUrl** field, Busola also reads it.
+
+  - **showClustersOverview** - optional configuration to instruct Busola to show the clusters overview rather than the current context cluster, after the clusters are loaded.
```
grego952, line 234:
```suggestion
  - **showClustersOverview** - optional configuration to instruct Busola to show **Clusters Overview** rather than the current context cluster, after the clusters are loaded.
```

docs/features.md
```diff
@@ -229,7 +229,11 @@ Note that some features must be run before the application starts the bootstrap
...
+
+  - **defaultKubeconfig** - define the optional defaultKubeconfigID to load this kubeconfig when you visit Busola homepage `/` and there are no memorized clusters in the application.
```
grego952, line 236:
```suggestion
  - **defaultKubeconfig** - define the optional default **KUBECONFIG_ID** to load this kubeconfig when you visit Busola homepage `/` and there are no memorized clusters in the application.
```

grego952 (approved):

dariadomagala (approved):

PR #1602, opened 2022-07-14
EXT: Fix Gaps between columns

### LINKED ISSUE #1575: EXT: Gap between Columns widget and the others widget isn't equal
**Description**
The gap between `Columns` widget and the others isn't equal.

**Expected result**

The gap between the widgets should be the same.

**Actual result**
The gap(1) between the `header` and the `Columns` widget is different than the one between `Columns` and `Pizza's recipe secret`(2).

image

**Steps to reproduce**

**Troubleshooting**

mrCherry97:
**Description**

- Change className extensibility-columns for panel-grid,
- Remove className extensibility-columns,
- Add prop for disabling margins inside columns,
- Added case for panel,
- Removed unnecessary console.log
- ...

**Related issue(s)**
Closes #1575


core-ui/src/components/Extensibility/components/Columns.js
```diff
@@ -1,13 +1,12 @@
...
 import { Widget } from './Widget';
-import './Columns.scss';
 
 export function Columns({ structure, ...props }) {
   return (
-    <div className="extensibility-columns">
+    <div className="panel-grid">
       {structure.children.map(child => (
-        <Widget structure={child} {...props} />
+        <Widget structure={child} hasExternalMargin={false} {...props} />
```
akucharska, line 9: Maybe we can rename `hasExternalMargin` because for me this name suggests the opposite behavior.
`hasExternalMargin={false}` -> the external wrapper **don't have** margin so add one
`hasExternalMargin={true}` -> the external wrapper **have** margin so skip it

mrCherry97: I used the same as it is in `GenericList`, so we have to think about remaining everything

akucharska (dismissed):

akucharska (approved):

PR #1604, opened 2022-07-14
fix codeviewer yaml with empty string

### LINKED ISSUE #1589: EXT: Code viewer breaks UI when is empty
**Description**
Looking on pizza example when user provide only name of pizza without `Order details` it provides to page with extensibility error.

**Expected result**
CodeViewer isn't displayed and doesn't throw error.

**Actual result**
image

**Steps to reproduce**
1. Go to `Pizza Orders`
2. Create Pizza Order
3. Provide name without Order details
4. Click Create 
5. Check details view

**Troubleshooting**

wchrapka:
**Description**

Fix a codeviewer bug, emerging when the language is set to yaml and the value is undefined.

**Related issue(s)**
Fixes #1589


core-ui/src/components/Extensibility/components/CodeViewer.js
```diff
@@ -16,33 +16,36 @@ export function CodeViewer({ value, structure, schema }) {
...
 
-    try {
-      switch (language) {
-        case 'yaml':
-          parsedValue = jsyaml.dump(value);
-          break;
-        default:
-          //this includes JSON and other languages
-          parsedValue = stringifyIfObject(value);
+    if (value) {
```
dariadomagala, line 19: `0` / `false` are proper values to display, this conditions should be only for empty values

dariadomagala (changes requested):

dariadomagala (approved):

PR #1618, opened 2022-07-18
Replace graph legend handling

Lyczeq:
**Description**

- ...
- ...
- ...

**Related issue(s)**


Lyczeq:
After spending some time on that topic, it is not worth the effort to replace the current solution with the `useCssVariable` hook at a push. It would require changing a lot of logic which is not valuable if current mechanisms work well.

PR #1625, opened 2022-07-19
changed path of defaultKubeconfigId feature

KaiReichart:
**Description**

- Change the path of the defaultKubeconfig to `/clusters`

In a previous [MR](https://github.com/kyma-project/busola/pull/1591), @wchrapka implemented the feature for setting a default kubeconfigID on the root path of the application. While this is an improvement, it only somewhat fixes the underlying issue.  

With this implementation, the end user would still set their bookmarks to the `/clusters` path, since this is the URL that will be redirected to from the root path. This means, that (depending on the configured storage option) the user will at some point be met with an empty cluster list again. 

It was mentioned, that an invalid kubeconfig will lead to an infinite loop, which (in my testing) turns out to not actually be completely accurate.

This popup is presented twice:
Screenshot 2022-07-19 at 13 31 04

After accepting these popups, the empty cluster list is shown:

Screenshot 2022-07-19 at 13 31 09

In my opinion, this is acceptable behavior for an invalid kubeconfig, especially since this feature has to be set up explicitly. 

**Related issue(s)**

https://github.com/kyma-project/busola/issues/1438
https://github.com/kyma-project/busola/pull/1591

### DIFF
--- a/core/src/luigi-config/kubeconfig-id/loadKubeconfigById.js
+++ b/core/src/luigi-config/kubeconfig-id/loadKubeconfigById.js
@@ -26,7 +26,7 @@ export async function loadKubeconfigById(kubeconfigId) {
     return null;
   }
 
-  const isHome = /^\/?$/.test(window.location.pathname);
+  const isHome = /^\/clusters?$/.test(window.location.pathname);
   const areClusters = Object.getOwnPropertyNames(getClusters()).length;
   const defaultKubeconfig = kubeconfigIdFeature.config.defaultKubeconfig;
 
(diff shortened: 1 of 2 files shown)


valentinvieriu:
@KaiReichart As we discussed on the Slack Channel, we probably don't want to send users in a redirect loop if somehow the default kubeconfig file is broken, so we should allow users to add a cluster.
We cannot trigger the kubeconfig fetch on /clusters
As we discussed, we can add a special button that will fetch that file.

KaiReichart:
Closing this in favor of https://github.com/kyma-project/busola/pull/1659/

PR #1634, opened 2022-07-20
EXT take schema from OpenAPI

### LINKED ISSUE #1596: SPIKE Extensibility: Evaluate The posibility of not hardcoding the Schema in our configuration
**Description**

Adding the schema to our configuration adds complexity and also increases the size of the configmap. Research if there is a way to to use the schema from the Kubernetes API endpoint and not store it in the configmap.

**Acceptance Criteria**

-  A user with Namespace access only should be able to utilise the extension. No special privileges should e needed

wchrapka:
**Description**

- Take EXT schemas from the `/openapi/v2` kuberenetes endpoint. 
- Fetch the openapi when the cluster changes

**Related issue(s)**

Resolves #1596


core-ui/src/components/App/App.js
```diff
@@ -45,61 +50,63 @@ export default function App() {
   }, []);
 
   return (
```
wchrapka, line 52: Added only the AppContext.Provider to the JSX

core-ui/src/components/App/AppContext.js
```diff
@@ -0,0 +1,8 @@
+import { createContext } from 'react';
+
+export const AppContext = createContext({
```
qbalukom, line 3: wouldn't it be cleaner to create a `SchemaContext`?
So we don't end up with a single monolithic context for everything?

core-ui/src/components/App/resourceSchemas/resourceSchemaWorkerApi.js
```diff
@@ -0,0 +1,45 @@
+export let schemasWorker = null;
```
qbalukom, line 1: i'm not sure about exporting `schemasWorker` directly. Since we're doing an API here, I think it would be more consistent to export `terminate` and `isAvailable` functions instead (names can be different, of course)

core-ui/src/components/App/AppContext.js
wchrapka: When Luigi is removed, we'll add its microcontext replacement here, no point to divide it to multiple contexts

core-ui/src/components/App/resourceSchemas/resourceSchemaWorkerApi.js
wchrapka: done

core-ui/src/components/App/AppContext.js
qbalukom: I guess we can wait until then with this topic

qbalukom (approved):

grego952 (approved):

PR #1644, opened 2022-07-22
Fix simple list names overwrites

### LINKED ISSUE #1555: Extensibility: Form View: Define form field names
**Description**
I want to be able to define a pretty text name for different fields the user should supply. For instance I want it to be `Type` instead of `spec.type` on the left side of the form field.

**Reasons**
In the details view the fields are already having the pretty name so I believe it would be more user friendly if when creating or editing a resource they see the same pretty names instead of the json path.

**Attachments**
image

dariadomagala:
**Description**

- ...
- ...
- ...

**Related issue(s)**

Fixes #1555


Wawrzyn321 (changes requested):
I deployed the _pizzas_ on a fresh cluster... I suppose it should show valid names?
image

Wawrzyn321 (approved):

PR #1649, opened 2022-07-25
Fix namespace and resource name conflict for extensibility 

### LINKED ISSUE #1647: Ext: Extensibility views might depend on a namespace
**Description**

Extensibility views somehow depend on a namespace and don't work properly. 
I've had `pizzas` and `pizzas-orders` examples on a cluster, and added `services` example to my localhost configuration. Everything worked fine on all namespaces until I checked it on `pizzas` namespace. Even tho the path was correct (`/namespaces/pizzas/services-new`), it was trying to show me the `pizzas` resource (but failed since it was using groups and versions from the services example, but name from pizzas example).

How the call should look like:
`/backend/api/v1/namespaces/pizzas/services`
How it really looks like:
`/backend/api/v1/namespaces/pizzas/pizzas`
How working pizzas call look like:
`/backend/apis/busola.example.com/v1/namespaces/pizzas/pizzas`

**Expected result**

Services example should work on all namespaces

**Actual result**

Services example doesn't work on a `pizzas` namespace

**Steps to reproduce**

**Troubleshooting**

dariadomagala:
**Description**

To test it go to `/cluster/shoot--hasselhoff--kmain/namespaces/pizzas/workloadgroups` for example.

- Fix function that chose which CR is the current one
- ...
- ...

**Related issue(s)**

Resolves #1647


core-ui/src/components/Extensibility/useGetCRbyPath.js
```diff
@@ -1,15 +1,19 @@
 import { useMicrofrontendContext } from 'shared/contexts/MicrofrontendContext';
 
 export const useGetCRbyPath = () => {
-  const { customResources = [] } = useMicrofrontendContext();
-  const resource = customResources.find(el => {
+  const { customResources = [], namespaceId } = useMicrofrontendContext();
+
+  const resource = (customResources || []).find(el => {
```
wchrapka, line 6: `|| []` is already at line 4
```suggestion
  const resource = (customResources).find(el => {
```

wchrapka (approved):

PR #1651, opened 2022-07-25
Fix example EXT bug

wchrapka:
**Description**

Adds missing `"` to the EXT pizza example


Lyczeq:
That bug has already been fixed in the following [PR](https://github.com/kyma-project/busola/pull/1652) along with one other bug.

PR #1659, opened 2022-07-27
Add button to load default kubeconfig

KaiReichart:
**Description**

Adding a Button to the Cluster list if 

- the defaultKubeconfigId feature is enabled
- the Clusterlist is empty

in order to load the default kubeconfig into the dashboard

As discussed with @valentinvieriu, this is favorable to always loading the default kubeconfig on the `/clusters` path proposed in https://github.com/kyma-project/busola/pull/1625

Screenshot 2022-07-27 at 11 06 39


core/src/luigi-config/communication.js
```diff
@@ -101,6 +102,9 @@ export const communication = {
         await setCluster(params?.kubeconfig?.['current-context']);
       }
     },
+    'busola.loadDefaultKubeconfigId': async () => {
+      await loadDefaultKubeconfigId();
```
Wawrzyn321, line 106: It is not an async function, we can get rid of `async`, `await` and `async` in `core/src/luigi-config/kubeconfig-id/loadKubeconfigById.js`.

core-ui/src/components/Clusters/views/ClusterList.js
```diff
@@ -163,6 +164,19 @@ function ClusterList() {
     />
   );
 
+  const loadDefaultClusterButton = (
+    <>
+      {features?.KUBECONFIG_ID?.config?.defaultKubeconfig && (
```
Wawrzyn321, line 169: We need to also check if `KUBECONFIG_ID` is enabled, because 
```json
"KUBECONFIG_ID": {
  "isEnabled": false,
  "config": {
    "defaultKubeconfig": "tets.yaml"
  }
},
```
will render the button, while the feature is disabled and clicking the button won't load the kubeconfig.

core/src/luigi-config/communication.js
KaiReichart: done

core-ui/src/components/Clusters/views/ClusterList.js
KaiReichart: done

Wawrzyn321 (approved):

grego952 (approved):

PR #1665, opened 2022-07-28
Err boundary for EXT

### LINKED ISSUE #1412: EXT: Validate config against a schema and show errors on ErrorBoundary page
**Description**

When the configuration is wrong and the code breaks, we catch the error and show it as an error - and that's fine.
image

Would be better to list the invalid properties beneath, like "this property should be an object, not an array" or other message (the actual error?) in case the schema is fine.

wchrapka:
**Description**

Add a separate error boundary to the EXT views to show a user the errors in his configuration.

**Related issue(s)**

Resolves #1412


core-ui/src/components/Extensibility/errors.md
```diff
@@ -0,0 +1,34 @@
+### Extensibility Error Handling
+
+There's a special error handler, that can be used when validating input from the EXT Config Maps.
+
+Users make errors in configuration, we can display the error's message and even the responsible part of a Config Map with the `throwConfigError` handler:
+
+```js
+if (!structure || typeof structure !== 'object') {
```
dariadomagala, line 8: I don't think we need to fully explain what will be displayed, which cases are handled here, show our code, etc. I don't even think we need separate documentation for that (since the user cannot do anything about it, it's not being configured or anything). We can have a short mention in the main documentation, that would be enough IMHO.

core-ui/src/components/Extensibility/components/Widget.js
```diff
@@ -72,6 +73,14 @@ export function Widget({ structure, value, inlineRenderer, ...props }) {
   } = useRelationsContext();
 
   let childValue;
+
+  if (!structure || typeof structure !== 'object') {
+    throwConfigError(t('extensibility.not-an-object'), structure);
+  }
+  if (!structure.path && !structure.children && !Array.isArray(structure)) {
```
dariadomagala, line 80: can we add a more general handler for the path - it not only cannot be empty, but it also cannot be a boolean for example.

core-ui/src/components/Extensibility/errors.md
wchrapka: It wasn't meant for the users, but for the developers who work on the software. A beginning to a guide 'how to work with extensibility'.

dariadomagala: Then @grego952 could you suggest what would be the best place for such docs? Should we keep it in the same place as docs for the users or should we move it somewhere else?

grego952: Correct me if I'm wrong but I think that most Busola users are developers. If it's only a short mention, I'd keep it in the main documentation.

grego952: Or if it's only for our team, maybe we can put it in our internal repo?

dariadomagala (dismissed):

core-ui/src/components/Extensibility/errors.md
```diff
@@ -0,0 +1,34 @@
+### Extensibility Error Handling
+
+There's a special error handler, that can be used when validating input from the EXT Config Maps.
+
+Users make errors in configuration, we can display the error's message and even the responsible part of a Config Map with the `throwConfigError` handler:
```
grego952, line 5:
```suggestion
When users make errors in configuration, you can display the error's message and even the responsible part of a Config Map with the `throwConfigError` handler:
```

grego952 (approved):

dariadomagala (approved):

grego952 (approved):

dariadomagala (approved):

PR #1669, opened 2022-07-28
Tooltip for allowing modifying protected resources

### LINKED ISSUE #1189: User should be instructed what to do to modify the protected resource
**Description**
When a resource is protected, a user cannot modify or delete it. Even though it's not recommended it may be sometimes needed. Busola allows users to do that but it's not mentioned anywhere.
For now the tooltip shows only the information about the original use, e.g.
Zrzut ekranu 2022-04-1 o 11 05 57
It would be good to update the protected resource's tooltip with short instruction what to do to modify or delete the protected resource.

**Area**

**Reasons**
User should be informed that it is possible to modify or delete the protected resource.

**Assignees**

@grego952 

**Attachments**

mrCherry97:
**Description**

- Add tooltips for protected resources actions,
   - Delete
   - Edit
   - Restart
- Add tooltips on the list and details,
- Add margin to `View Yaml` button

**Related issue(s)**
Closes #1189


core/src/i18n/en.yaml
```diff
@@ -600,6 +600,8 @@ common:
     key: Key must start and end with an alphanumeric character, can contain '-', '_' or '.'.
     key-value: Key and value must start and end with an alphanumeric character, can contain '-', '_' or '.'.
     line: Ln
+    protected-resources-info: To allow for modifying protected resources go to Preferences > Clusters.
+    protected-resources-view-yaml: View YAML. To allow for protected resources go to Preferences > Clusters.
```
grego952, line 604:
```suggestion
    protected-resources-view-yaml: View YAML. To allow for modifying protected resources go to Preferences > Clusters.
```

grego952 (dismissed):

core-ui/src/shared/components/ResourceDetails/ResourceDetails.js
```diff
@@ -183,12 +184,17 @@ function Resource({
   const editAction = () => {
     if (protectedResource) {
       return (
-        <Button
-          className="fd-margin-end--tiny"
-          onClick={() => openYaml(resource)}
+        <Tooltip
```
dariadomagala, line 187: Maybe we could add the same information in the editor, somewhere next to the `read only` information?
Screen Shot 2022-07-28 at 14 22 32
(btw, the text is barely readable, if we agree to change the text, please also change its color)

core-ui/src/shared/components/ResourcesList/ResourcesList.js
```diff
@@ -237,6 +237,7 @@ export function ResourceListRenderer({
       nameSelector(resource) + '.yaml',
       handleSaveClick(resource),
       isProtected(resource),
+      isProtected(resource),
```
dariadomagala, line 240: why twice?

mrCherry97: Because we are setting `readOnly` and `isProtected` parameters, it's the same in this case.

dariadomagala (approved):

grego952 (approved):

PR #1695, opened 2022-08-03
Fix name validation

### LINKED ISSUE #1687: Name of configmap does not accept .
**Description**

Seems that validation is not allowing . in the name

**Expected result**

The name should accept . in the name

**Actual result**

**Steps to reproduce**

**Troubleshooting**

image

Wawrzyn321:
**Description**

- Adjust regex to allow '.'

**Related issue(s)**

Closes #1687


mrCherry97 (approved):

mrCherry97:
In my opinion, we have to make validation with `.` as default and for resources that allow names without it custom validation for only `-` allowing, but as we discuss on a daily, we are just changing regex for now.

Wawrzyn321:
This is how it was done before, but I rolled it back after a daily discussion.

mrCherry97 (approved):

grego952 (approved):

wchrapka (approved):

PR #1705, opened 2022-08-03
EXT: Convert "resource" section to "general"

### LINKED ISSUE #1697: Extensibility: Convert "resource" section to "general"
- Rename `resource` section to `general`.
- Update the structure to be more organized
- Introduce name/group/description to the resource (this should still be passed through i18n in case somebody wants to translate them)
- Rename `path` field to `urlPath` or `urlSegment`

Currently suggested structure (but not necessarily a final decision):
```
{
  resource: { kind, version, group },
  name,
  category,
  description,
  urlPath,
  featureFlags: { ... }
}
```
(`featureFlags` is a future idea that is not implemented yet)

Lyczeq:
**Description**

- Adjusted examples and docs
- Resource group is no longer required. Previously, when I wanted to use it with `Service` it was necessary to set a group as an empty string. Now, if the resource belongs to the "core" (also called legacy) k8s group, the user doesn't have to define the group in the config.
- Adjusted the code to use information from the `general` section instead of `resource`

**Related issue(s)**
Resolves #1697

### DIFF
--- a/core-ui/src/components/BusolaExtensions/BusolaExtensionCreate.js
+++ b/core-ui/src/components/BusolaExtensions/BusolaExtensionCreate.js
@@ -78,13 +78,13 @@ export function BusolaExtensionCreate({ formElementRef, onChange }) {
           <>
             <ResourceForm.FormField
               required
-              propertyPath="$.translations.en.name"
+              propertyPath="$.general.name"
               label={t('common.labels.name')}
               input={Inputs.Text}
             />
             <ResourceForm.FormField
               required
-              propertyPath="$.translations.en.category"
+              propertyPath="$.general.category"
               label={t('common.labels.category')}
               input={Inputs.Text}
             />
--- a/core-ui/src/components/Extensibility/helpers.js
+++ b/core-ui/src/components/Extensibility/helpers.js
@@ -149,3 +149,8 @@ export const throwConfigError = (message, code) => {
   e.name = 'Extensibility Config Error';
   throw e;
 };
+
+export const createOpenApiSchemaId = resource => {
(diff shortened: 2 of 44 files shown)


core-ui/src/components/BusolaExtensions/BusolaExtensionCreate.js
```diff
@@ -78,13 +78,13 @@ export function BusolaExtensionCreate({ formElementRef, onChange }) {
           <>
             <ResourceForm.FormField
               required
-              propertyPath="$.translations.en.name"
+              propertyPath="$.general.name"
```
Wawrzyn321, line 81: Why was it changed? We should keep this stuff in `translations`.

core-ui/src/components/Extensibility/helpers.js
```diff
@@ -61,11 +61,11 @@ export const useGetTranslation = path => {
...
+  const { version, group } = api;
+  const apiVersion = group ? `${group}/${version}` : version;
+
   const template = {
-    apiVersion:
-      api.group === 'core' || api.group === ''
-        ? api.version
-        : `${api.group}/${api.version}`,
+    apiVersion,
     kind: pluralize(api.kind, 1),
```
Wawrzyn321, line 69: Not your code, but `kind` is always singular.
```suggestion
    kind: api.kind
```

Wawrzyn321: (And probably delete the `pluralize` import.)

grego952 (dismissed):

grego952 (approved):

Wawrzyn321 (approved):

PR #1760, opened 2022-08-18
Fixed Certificates and the ResourceLink widget

Lyczeq:
**Description**

- Adjusted examples from the resources directory to the new format
- Fixed a ResourceLink, previously it was necessary to use linkText, now it takes a name as a default

**Related issue(s)**

### DIFF
--- a/core-ui/src/components/Extensibility/components/ResourceLink.js
+++ b/core-ui/src/components/Extensibility/components/ResourceLink.js
@@ -48,6 +48,7 @@ export function ResourceLink({ value, structure, originalResource }) {
   const linkContent = tExt(structure.linkText, {
     data: value,
     root: originalResource,
+    defaultValue: value,
   });
 
   return (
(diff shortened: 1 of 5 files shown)


examples/resources/configuration/certificates.yaml
```diff
@@ -25,70 +24,64 @@ data:
...
-              "type": "integer",
-              "name": "revisionHistoryLimit"
+              "source": "spec.revisionHistoryLimit",
+              "name": "Revision History Limit"
             },
             {
-              "path": "spec.secretName",
-              "type": "string",
+              "source": "spec.secretName",
               "name": "secretName"
```
dariadomagala, line 53:
```suggestion
              "name": "Secret name"
```
since we're adjusting everything else...

core-ui/src/components/Extensibility/components/ResourceLink.js
```diff
@@ -48,6 +48,7 @@ export function ResourceLink({ value, structure, originalResource }) {
   const linkContent = tExt(structure.linkText, {
     data: value,
     root: originalResource,
+    defaultValue: value.name,
```
dariadomagala, line 51:
```suggestion
    defaultValue: value?.name,
```

dariadomagala (approved):

PR #1790, opened 2022-08-24
Extensibility: Custom form fields

qbalukom:
- improve form field reordering
- add variables (non-resource fields) to forms


wchrapka:
I wasn't able to make this work as well:

```
    "form": [
      {
        "var": "useDescription",
        "type": "boolean"
      },
      {
        "simple": true,
        "path": "spec.host",
        "visibility": "$useDescription"
      }
]

**Comment by qbalukom** [comment]
> It doesn't work for me for that JSON, any idea why?:
> 
> ```
>  "form": [
>       {
>         "simple": true,
>         "path": "spec.host"
>       },
>       {
>         "widget": "FormGroup",
>         "simple": true,
>         "path": "spec.trafficPolicy",
>         "children": [
>           {
>             "widget": "FormGroup",
>             "path": "loadBalancer",
>             "children": [
>               {
>                 "path": "simple"
>               },
> 
>               {
>                 "widget": "FormGroup",
>                 "path": "consistentHash",
>                 "children": [
>                   {
>                     "var": "consistentHashSelector",
>                     "type": "string",
>                     "enum": ["choose One", "httpHeaderName"]
>                   },
>                   {
>                     "path": "httpHeaderName",
>                     "visibility": "$vars.consistentHashSelector = 'httpHeaderName'"
>                   }
>                 ]
>               },
> ```

there is no array anywhere, all parents are `FormGroup`s. So no `$vars` will be generated. so you should use `$consistentHashSelector`. (although now i see it might be a good idea to have root vars in `$vars` when not in an array... just in case)

**Comment by qbalukom** [comment]
> I wasn't able to make this work as well:
> 
> ```
>     "form": [
>       {
>         "var": "useDescription",
>         "type": "boolean"
>       },
>       {
>         "simple": true,
>         "path": "spec.host",
>         "visibility": "$useDescription"
>       }
> ]
> ```

The only issue i see here is that var field doesn't have `"simple": true,` - maybe that's the issue?

**Review by wchrapka — APPROVED** [review]

**Comment by wchrapka** [comment]
> > It doesn't work for me for that JSON, any idea why?:
> > ```
> >  "form": [
> >       {
> >         "simple": true,
> >         "path": "spec.host"
> >       },
> >       {
> >         "widget": "FormGroup",
> >         "simple": true,
> >         "path": "spec.trafficPolicy",
> >         "children": [
> >           {
> >             "widget": "FormGroup",
> >             "path": "loadBalancer",
> >             "children": [
> >               {
> >                 "path": "simple"
> >               },
> > 
> >               {
> >                 "widget": "FormGroup",
> >                 "path": "consistentHash",
> >                 "children": [
> >                   {
> >                     "var": "consistentHashSelector",
> >                     "type": "string",
> >                     "enum": ["choose One", "httpHeaderName"]
> >                   },
> >                   {
> >                     "path": "httpHeaderName",
> >                     "visibility": "$vars.consistentHashSelector = 'httpHeaderName'"
> >                   }
> >                 ]
> >               },
> > ```
> 
> there is no array anywhere, all parents are `FormGroup`s. So no `$vars` will be generated. so you should use `$consistentHashSelector`. (although now i see it might be a good idea to have root vars in `$vars` when not in an array... just in case)

It would help me greatly in the serialization of my schema, if the syntax was the same for arrays and regular fields.

File: docs/extensibility/resources.md [hunk 1]
```diff
@@ -97,6 +101,48 @@ If you target elements of an array rather that the array itself, you can use `it
...
+
+```json
+[
+  { "var": "useDescription", "type": "boolean" },
+  { "path": "spec.description": "visibility": "$useDescription" },
+  { "path": "spec.items", "widget": "GenericList": "children": [
+    { "path": "[]", "children": [
+      { "path": "name" },
+      { "var": "itemMode", "type": "string", "enum": ["simple", "verbose"] },
+      { "path": "price", "visibility": "$itemMode = 'verbose'" },
```
wchrapka, line 138: This notation doesn't seem to work when we have 2+ arrays with the same selector i.e. we have the same JSON in :

1. trafficPolicy (works)
2. port -> trafficPolicy (if trafficPolicy from 1st point is not selected, switches value for point 2. and 3., otherwise doesn't work)
3. subset -> trafficPolicy (if trafficPolicy from 1st point is not selected, switches value for point 2. and 3., otherwise doesn't work)
4. subset -> port -> works as long 1 or 2 or 3 are not selected
```
       {
                        "var": "consistentHashSelector",
                        "type": "string",
                        "enum": ["simple", "httpHeaderName"]
                      },
                      {
                        "path": "httpHeaderName",
                        "visibility": "$consistentHashSelector = 'httpHeaderName'"
                      },
```

qbalukom: variable names are not scoped. they have to be unique across the config. i didn't mention it in the docs?

wchrapka (dismissed):

docs/extensibility/resources.md
```diff
@@ -97,6 +101,48 @@ If you target elements of an array rather that the array itself, you can use `it
 ]
 ```
 
+### Variable fields
+
+Additionally it's possible to define variable fields. In this case **path** is ommited and instead a **var** argument is used to specify variable name to assign to. Variable names have to be unique across the extension. Such a value is not added to the resultant yaml, but instead stored in memory and provided to any [JSONata](https://docs.jsonata.org/overview.html) handlers as variables (e.g. `$foo`). Variables will be provided for the current context - so if a variable is defined inside an array, the value is for that specific item. To access raw values, the predefined `$vars` variable has to be used.
+
+When using a variable inside an array it has to be wrapped inside a `[]` element (see example).
```
grego952, line 108:
```suggestion
When using a variable inside an array it has to be wrapped inside a `[]` element (see [example](#example)).
```

docs/extensibility/resources.md
```diff
@@ -97,6 +101,48 @@ If you target elements of an array rather that the array itself, you can use `it
...
+
+### Item parameters
+
+- **var** - _[required]_ variable name.
+- **type** - _[required]_ type of field, as defined by JSON Schema.
+
+All other fields can be used analogous to regular form items (except for the **path** and **children** parameters).
```
grego952, line 115:
```suggestion
All other fields can be used analogously to regular form items (except for the **path** and **children** parameters).
```

docs/extensibility/resources.md
```diff
@@ -97,6 +101,48 @@ If you target elements of an array rather that the array itself, you can use `it
...
+
+### Predefined variables
+
+All jsonata expressions have a few variables that are predefined instead of being read from variable fields.
```
grego952, line 119:
```suggestion
All JSONata expressions have a few variables that are predefined instead of being read from variable fields.
```

docs/extensibility/resources.md
```diff
@@ -97,6 +101,48 @@ If you target elements of an array rather that the array itself, you can use `it
...
+
+- **\$item** - when in an array this contains the current item.
+- **\$index** - contains the index of the current item within the array.
+- **\$indexes** - contains a list of all the indexex for nested arrays.
```
grego952, line 123:
```suggestion
- **\$indexes** - contains a list of all the indexes for nested arrays.
```

docs/extensibility/resources.md
```diff
@@ -97,6 +101,48 @@ If you target elements of an array rather that the array itself, you can use `it
...
+- **\$vars** - contains a map of all the raw variable values.
+
+### Example
+
+In the example the visibility for item price and color are analogous - the former one uses scoped variables for the current item, the latter extracts the value from an array variable using provided index - this is mostly useful for complex scenarios only.
```
grego952, line 128:
```suggestion
In the example, the visibility for item price and color are analogous - the former uses scoped variables for the current item, and the latter extracts the value from an array variable using provided index - this is mostly useful for complex scenarios only.
```

grego952 (approved):

wchrapka (approved):

PR #1806, opened 2022-08-31
Proxy: change 503 to 502

Wawrzyn321:
**Testing**

Use the following code in our App.js - it makes continuous calls that return 503/502:

```
import { useSingleGet } from 'shared/hooks/BackendAPI/useGet';

export default function App() {
const fetch = useSingleGet();

  setInterval(async () => {
    try {
      console.log(
        await fetch(
          '/api/v1/namespaces/kyma-system/services/monitoring-prometheus:web/proxy/api/v1/status/runtimeinfo',
        ),
      );
    } catch (e) {
      console.log(e);
    }
  }, 2000);

  return null;
```

### DIFF
--- a/backend/common.js
+++ b/backend/common.js
@@ -105,7 +105,7 @@ export const makeHandleRequest = () => {
 
       // change all 503 into 502
       const statusCode =
-        k8sResponse.statusCode !== 503 ? k8sResponse.statusCode : 502;
+        k8sResponse.statusCode === 503 ? 502 : k8sResponse.statusCode ;
 
       res.writeHead(statusCode, {
         'Content-Type': k8sResponse.headers['Content-Type'] || 'text/json',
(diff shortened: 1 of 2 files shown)


backend/common.js
```diff
@@ -103,7 +103,11 @@ export const makeHandleRequest = () => {
           'Response headers are potentially dangerous',
         );
 
-      res.writeHead(k8sResponse.statusCode, {
+      // change all 503 into 502
+      const statusCode =
+        k8sResponse.statusCode !== 503 ? k8sResponse.statusCode : 502;
```
wchrapka, line 108: IMO it would be more readable without the negation 
https://moderatemisbehaviour.github.io/clean-code-smells-and-heuristics/general/g29-avoid-negative-conditionals.html
```suggestion
        k8sResponse.statusCode === 503 ? 502 : k8sResponse.statusCode ;
```

Wawrzyn321: My pleasure.

wchrapka (approved):

PR #1821, opened 2022-09-07
Allow users to create function without dependencies

pPrecel:
**Description**

- as in the title

**Related issue(s)**

Fixes https://github.com/kyma-project/kyma/issues/15339

### DIFF
--- a/core-ui/src/components/Functions/FunctionDetails/Tabs/Code/CodeAndDependencies/CodeAndDependencies.js
+++ b/core-ui/src/components/Functions/FunctionDetails/Tabs/Code/CodeAndDependencies/CodeAndDependencies.js
@@ -72,6 +72,7 @@ export default function CodeAndDependencies({ func }) {
     }
 
     setDisabledCause(DISABLED_CAUSES.VALID);
+    setDependencies(deps);
   }
 
   function handleSave() {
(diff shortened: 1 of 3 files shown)


kwiatekus (approved):
tested on local busola instance.
Works as expected.
Empty dependencies are allowed ( defaulted to `{}` by serverless backend )

core-ui/src/components/Functions/FunctionDetails/Tabs/Code/CodeAndDependencies/CodeAndDependencies.js
```diff
@@ -72,6 +72,7 @@ export default function CodeAndDependencies({ func }) {
     }
 
     setDisabledCause(DISABLED_CAUSES.VALID);
+    setDependencies(deps);
```
Wawrzyn321, line 75: When is it needed? I removed it and it works just fine.

pPrecel: It's needed when dependencies input from UI contains only spaces, `\n` or `\t`. If you would pass only these symbols in the dependencies on the UI, then the function would report a build error.

Wawrzyn321 (approved):

PR #1831, opened 2022-09-12
EXT: Move Istio Service Entries to an extenstion

### LINKED ISSUE #1734: Move ISTIO Service Entries from the Busola codebase to an extension.
**Description**

Move ISTIO Service Entries from the Busola codebase to an extension.

**Acceptance Criteria**

- [x] The application behaves and looks similar - small changes are accepted 
- [ ] Test are adjusted
- [x] We try to reuse the existing widgets.If new widgets are needed, they need to be created and documented
- [x] If edit fields are YAML only, we introduce form base editor

Lyczeq:
**Description**

- Added the list section
- Added the form section
- Added the details section
- Added translations and placeholders

**Related issue(s)**
Resolves #1734


examples/resources/istio/service-entries.yaml
```diff
@@ -0,0 +1,240 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: serviceentries
+  namespace: lyczeq
```
mrCherry97, line 5: I think namespace should be `kube-public`

examples/resources/istio/service-entries.yaml
```diff
@@ -0,0 +1,240 @@
...
+          {
+            "name": "summary",
+            "widget": "Panel",
+            "source": "spec",
+            "visibility": "$boolean($exists(data.hosts) or $exists(data.addresses) or $exists(data.subjectAltNames))",
+            "children": [
+              { "name": "spec.hosts", "source": "$item.hosts", "widget": "JoinedArray", "visibility": "$exists(*)"},
+              { "name": "spec.addresses", "source": "$item.addresses", "widget": "JoinedArray", "visibility": "$exists(*)"},
+              { "name": "spec.exportTo", "source": "$item.exportTo", "widget": "Labels", "placeholder": "Exported to all Namespaces"},
+              { "name": "spec.subjectAltNames", "source": "$item.subjectAltNames", "widget": "JoinedArray", "visibility": "$exists(*)"},
```
mrCherry97, line 51: Previously we displayed here `Labels` and the title was configuration, if we want to follow our previous implementation it should be changed, if you have reason why it's like this it can stay, not strong opinion
image

mrCherry97 (changes requested):
Let's pull from main, and check tooltips

examples/resources/istio/service-entries.yaml
Lyczeq: Yes, definitely :D

examples/resources/istio/service-entries.yaml
Lyczeq: I will change the title to the `Configuration` but I'd stay with the chosen widgets. We use normal strings for hosts in VirtualServices, so it would keep consistency. Same as addresses, in my opinion it's better to have them as strings not labels/tokens. I've chosen Labels widget for `exportTo` value because the elements can sometimes look like `*` or `.`, so it would be more readable if they are presented as labels/tokens.

grego952 (approved):

mrCherry97 (approved):

PR #1850, opened 2022-09-15
dynamic values

qbalukom:


core-ui/src/components/Extensibility/components-form/ResourceRefRenderer.js
```diff
@@ -47,7 +47,7 @@ export function ResourceRefRender({
 
   return (
     <ExternalResourceRef
-      title={t('extensibility.widgets.resource-ref.header')}
+      title={tFromStoreKeys(storeKeys, schema)}
```
Wawrzyn321, line 50: Please remove it from translations.

Wawrzyn321 (dismissed):

docs/extensibility/form-section.md
```diff
@@ -59,6 +59,8 @@ When using a variable inside an array it has to be wrapped inside a `[]` element
 
 - **var** - _[required]_ variable name.
 - **type** - _[required]_ type of field, as defined by JSON Schema.
+- **defaultValue** - default value to use for the variable when opening the form.
```
grego952, line 62:
```suggestion
- **defaultValue** - default value used for the variable when opening the form.
```

grego952 (approved):

Wawrzyn321 (changes requested):
It looks like dynamic/default values don't work for items nested inside arrays:
```json
 "children": [
    "path": "[]",
      "children": [
		{
			"var": ...
```

You can check it on `/cluster/shoot--hasselhoff--kmain/namespaces/default/qqqs` - if you use YAML:
```yaml
apiVersion: busola.example.com/v1
kind: Qqq
spec:
  list-items:
    - description: 'a'
      name: 'b'
```

and get back to "Simple" the `description` is removed, while it should be left alone, according to `dynamicValue: "$boolean($item.description)"`.

Wawrzyn321 (approved):

PR #1861, opened 2022-09-20
EXT form: remove doubled form state

### LINKED ISSUE #1810: Performance: Redesign ResourceForm
`ResourceForm` as it is was built to manage underlying data. However with ui-schema being used for building forms that data-management layer is not necessary. This is most probably causing performance issues in extensibility forms.
Investigate options of separating data management from the visuals and make it possible to swap between various ui-schemes (in this case internal and ui-schema).
Hint: Consider using a context provider.

wchrapka:
**Description**

Removes double rendering of ext form, data is kept in an immutable record (ui-schema format) and converted only at need (instead of at every keystroke). 

**Related issue(s)**
Resolves #1810


wchrapka:
<html>
<body>


/test pre-busola-integration-test-k3s-cluster
--



</body>
</html>

qbalukom (approved):

PR #1883, opened 2022-09-26
Change code blocks from the Example sections to YAML

### LINKED ISSUE #1873: Change all the Extensibility examples into yaml format
**Description**

We want to be consistent in our documentation. We decided to promote the usage of YAML for the extensions. We need to adjust all our documentation and examples to use yaml.

grego952:
**Description**

- Change code blocks from the Example sections to YAML
- Minor doc improvements

**Related issue(s)**
Resolves #1873

### DIFF
--- a/docs/extensibility/display-section.md
+++ b/docs/extensibility/display-section.md
@@ -577,9 +577,7 @@ Table widgets display array data as rows of a table instead of free-standing com
 ```yaml
 - source: spec.toppings
   widget: Table
-  collapsibleTitle:
-    "'Topping #' & $string($index + 1) & (' ' & $join($keys($item),
-    ' '))"
+  collapsibleTitle: "'Topping #' & $string($index + 1) & (' ' & $join($keys($item), ' '))"
   collapsible:
     - source: quantity
   children:
--- a/docs/extensibility/display-section.md
+++ b/docs/extensibility/display-section.md
@@ -39,32 +39,22 @@ Each object adds a new column to your table.
 
 ### Example
 
-```json
-[
-  {
-    "source": "spec.url",
-    "sort": {
-      "default": true,
-      "compareFunction": "$compareStrings($first, $second)"
-    }
-  },
-  { "source": "spec.priority", "widget": "Badge" },
-  { "source": "$join(spec.toppings.name, ', ')" },
-  {
-    "name": "quantityIsMore",
-    "source": "$filter(spec.toppings, function ($v, $i, $a) { $v.quantity > $average($a.quantity) })"
-  },
-  { "source": "$join(spec.volumes.name, ', ')" },
-  {
-    "source": "$filter(spec.volumes, function ($v, $i, $a) {'configMap' in $keys($v)})" // List the array of Volume objects that have a config map
-  },
-  {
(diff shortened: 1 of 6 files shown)


docs/extensibility/form-section.md
```diff
@@ -311,30 +284,26 @@ ResourceRef widgets render two dropdowns to select the associated resources' nam
...
+      enum: '$keys($secret.data)'
+- path: spec.my-gateways
+  widget: ResourceRef
+  resource:
+    kind: Gateway
+    group: networking.istio.io
+    version: v1alpha3
+  toInternal:
+    "( $values := $split($, '/'); { 'namespace': $values[0], 'name': $values[1]
+    } )"
```
qbalukom, line 305: I think it would be cleaner to use the `|` notation here
```suggestion
  toInternal: |
    (
      $values := $split($, '/');
      { 'namespace': $values[0], 'name': $values[1] }
    )
```

docs/extensibility/form-section.md
```diff
@@ -283,14 +252,18 @@ KeyValuePair widgets render an `object` value as a list of fields. One is used f
...
-  "keyEnum": ["prefix", "regex", "exact"]
-}
+```yaml
+path: spec.my-data
+widget: KeyValuePair
+description: Key and value must start and end with an alphanumeric character.
+tooltip:
+  Labels are intended to be used to specify identifying attributes of objects
+  that are meaningful and relevant to users, but do not directly imply semantics to
+  the core system.
```
qbalukom, line 262: This is not a correct way to specify multi-line strings in yaml. either `>` or `|` has to be used
```suggestion
tooltip: >
  Labels are intended to be used to specify identifying attributes of objects
  that are meaningful and relevant to users, but do not directly imply semantics to
  the core system.
```

qbalukom (changes requested):

docs/extensibility/form-section.md
qbalukom: Also I wonder about widget examples, as we've got a bit of an inconsistency here. Some examples are given as an array:
```
- source: spec.foo
  name: Foo
- source: spec.bar
  name: Bar
```
but most singhle-entry examples are given just as an object:
```
source: spec.foo
name: Foo
```
however not all.
Maybe for the sake of consistency we should use the same format, suggesting it's an element of an array:
```
- source: spec.foo
  name: Foo
```
even if it's just a single entry?

docs/extensibility/display-section.md
```diff
@@ -671,20 +574,21 @@ Table widgets display array data as rows of a table instead of free-standing com
...
-      "source": "$item.price",
-      "sort": { "default": true, "compareFunction": "$second -$first" }
-    }
-  ]
-}
+```yaml
+- source: spec.toppings
+  widget: Table
+  collapsibleTitle:
+    "'Topping #' & $string($index + 1) & (' ' & $join($keys($item),
```
qbalukom, line 581: this is a strange line break...

qbalukom (changes requested):

qbalukom (approved):

NHingerl (approved):

PR #1893, opened 2022-09-28
Ext init phase

### LINKED ISSUE #1818: EXTENSIBILITY: Extensions have an init phase to detect if user has access to the resource underneath
**Description**

Extensions even if present can be disabled if the user has no access to them.

**Reasons**

**Attachments**

wchrapka:
**Description**
Hides resources that don't exists or user doesn't have persmission for 

**Related issue(s)**

Resolves #1818


core/src/assets/customResources/customResources.json
```diff
@@ -1,81 +1 @@
...
-        "spec.exportTo": "Exported To Namespaces",
-        "spec.host": "Host",
-        "spec.trafficPolicy": "Traffic Policy",
-        "spec.subsets": "Subsets",
-        "spec.workloadSelector": "Workload Selector"
-      }
-    }
-  }
-}
+{}
```
dariadomagala, line 1: I believe we should move this resource to a config map in examples

dariadomagala (dismissed):

Wawrzyn321 (approved):

PR #1905, opened 2022-10-04
Introduce Reset form button

### LINKED ISSUE #1879: UX: reset button in footer
Reset option should be present in forms even if no presets are set.
Reset button should be located between the "create/update" button and the "cancel" button.
When clicking the button it should reset to the empty template when creating a new resource, or the original value when editing.
No "clear" value should be present in the presets dropdown anymore.

Lyczeq:
**Description**

- Remove Clear templates from our core-ui components
- Remove Clear templates from extensibility
- Fix some inputs to allow reseting form
- Add Button to reset a form(resource)
- Remove `Disable Sidecar` feature from CronJobs. The feature was previously removed from Jobs, so I think it should be removed from CronJobs too.

**Related issue(s)**
Resolves #1879


dariadomagala (changes requested):
I believe that reset button in namespaces should also reset limits and quotas settings

core-ui/src/resources/Namespaces/NamespaceCreate.js
```diff
@@ -90,6 +91,32 @@ export function NamespaceCreate({
...
+          ? cloneDeep(initialNamespace)
+          : createNamespaceTemplate(),
+      );
+
+      if (!initialNamespace) {
+        setWithLimits(false);
+        setLimits(createLimitRangeTemplate({}));
+        setWithMemory(false);
+
+        setMemory(createResourceQuotaTemplate({}));
```
dariadomagala, line 107: I believe the 3 functions below are not needed

dariadomagala (approved):

grego952 (approved):

PR #1920, opened 2022-10-10
Default form fields in schema

Wawrzyn321:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/core-ui/src/components/Extensibility/ExtensibilityCreate.js
+++ b/core-ui/src/components/Extensibility/ExtensibilityCreate.js
@@ -1,5 +1,5 @@
 import React, { useCallback, useMemo, useState } from 'react';
-import Immutable from 'immutable';
+import Immutable, { List } from 'immutable';
 import pluralize from 'pluralize';
 import { useTranslation } from 'react-i18next';
 import * as jp from 'jsonpath';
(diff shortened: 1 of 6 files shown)


docs/extensibility/form-section.md
```diff
@@ -275,6 +293,7 @@ KeyValuePair widgets render an `object` value as a list of fields. One is used f
   - **valueEnum[]** - an array of options to generate a value input field with a dropdown.
 - **inputInfo** - a string below the input field that shows how to fill in the input.
 - **description** - a string displayed in a tooltip when you hover over a question mark icon, next to the input's label. The default value is taken from the CustomResourceDefintion (CRD).
+- **defaultOpen** - a boolean that specifies if the widget should be expanded by default. Defaults to `false`.
```
mrCherry97, line 296: Here have set the default `true` so it should be changed in the code or here

core-ui/src/components/Extensibility/ExtensibilityCreate.js
```diff
@@ -90,8 +90,55 @@ export function ExtensibilityCreateCore({
...
+    };
+
+    return [
+      nameField,
+      labelsField,
+      annotationsField,
+      ...schemaRules.filter(r => !PREDEFINED_PATHS.includes(r.path)),
+    ];
+  };
+
```
mrCherry97, line 139: image
Maybe we can move it to a separate file because this file is so so so big.

mrCherry97 (changes requested):
/woof

mrCherry97 (dismissed):
Maybe a good option will be to set it open when value is provided?
Scenario:
1. Open Create, fill in some fields that are without defaultExpanded flag, go to YAML and go back to Simple
2. Go to details, click Edit -> fields with values should be open.

We can think about it, make an issue and decide later.

Wawrzyn321:
I'd go for new issue then.

docs/extensibility/form-section.md
```diff
@@ -100,6 +100,23 @@ In the example, the visibility for item price and color are analogous - the form
           visibility: '$useDescription'
 ```
 
+## Default fields
+
+Each form is created with following fields:
```
grego952, line 105:
```suggestion
Each form is created with the following fields:
```

grego952 (approved):

mrCherry97 (approved):

Issue #1939, opened 2022-10-14
Rewrite Luigi side menu


wchrapka:
**Description**

- [x] create onClick fn for links in the navigation https://github.com/kyma-project/busola/pull/1981
- [x] map static navigation nodes to resources/***/index.js  https://github.com/kyma-project/busola/pull/1937
- [x] move footer https://github.com/kyma-project/busola/pull/1969
- [x] create fn to assemble the navigation structure based on re-maped nodes 
- [x] write unit test for the logic
- [x] style and html the navigation JS object with fundamental
- [x] handle the unusual nodes (CR, Extensibility, Halm, External) https://github.com/kyma-project/busola/pull/1978
- [x] create a config map toggle
- [x] handle the external nodes https://github.com/kyma-project/busola/pull/1993

Uncovered issues:

- [x] selectors reload too often as a config feature `prometheus` loads few times https://github.com/kyma-project/busola/pull/1987

dariadomagala:
Things still to do as a follow-up:

- [ ] External nodes - add an icon next to the node name showing that it's external. (it's implemented in fundamental, but the types don't match)
- [ ] Hide all popovers after clicking on them (for example namespaces dropdown stays open after changing the namespace)
- [ ] If changing a cluster fails, the navigation doesn't happen, but the name in the header shows the new cluster name (maybe it will be fixed after the removal of Luigi?)
- [ ] Make the left side nav smaller on mobile/small screens
- [ ] Write tests

Issue #1924, opened 2022-10-11
Remove Luigi as dependency


valentinvieriu:
**Description**

As we moved away from using Microfrontends as extensibility, we no longer need the advantages provided by Luigi Framework

**Reasons**

Streamline our application, and speed up bootstrap time. 

**Attachments**

- [x] https://github.com/kyma-project/busola/issues/1925
- [x] https://github.com/kyma-project/busola/issues/1926
- [ ] https://github.com/kyma-project/busola/issues/1927
- [x] https://github.com/kyma-project/busola/issues/1939
- [x] https://github.com/kyma-project/busola/issues/2071
- [x] https://github.com/kyma-project/busola/issues/2077
- [x] https://github.com/kyma-project/busola/issues/2078
- [x] https://github.com/kyma-project/busola/issues/2079

Lyczeq:
In the following [issue](https://github.com/kyma-project/busola/pull/1931) we managed to move `disableResourceProtection`, `dontConfirmDelete` and `pageSize` states to Recoil but not Theme, Language and `showHiddenNamespaces`.  The last three share state with Luigi, so if we want to deliver changes iteratively, it would be better to leave them for now and move them later.

Lyczeq:
In the following [issue](https://github.com/kyma-project/busola/pull/1969) I've added a `version.yaml`, where the busola version is kept. We should adjust our Makefile when deleting Luigi, because currently the new busola version is set in `core/src/assets/version.yaml`.

Lyczeq:
1a. Create store for Luigi context
&
1b. Remove cache
2. Recreate Navigation Luigi Logic in React (leave it in Luigi as it does more things there), hide the Luigi nav.
We have to move: Ext calls, `addExternalNodes`, `selfsubjectrules`, node list, destinationRules
1c. Modal with pagination and settings, feature toggles stay in Luigi
3. settings - "feature toggles" moved to React (addon to 1c)
---done
4. top nav - createNamespacesList, handlers in React to change the namespace
5.
clusterManagemnt, clusterStorage
login
Cluster handshake
kubeconfigID, behaviours like saving the last visited view, setting the current cluster when the name is present in URL
constants
feature-checks
feature-discovery (REMOVE PRIMARY, SECONDARY)
auth
6. move tracking
7. reimplenet toast (find a library)
8. remove iframe

dariadomagala:
Luigi removal tasks (that I can think of):

- [ ] rewrite all Luigi navigates to react-router
[...]

PR #1969, opened 2022-10-21
Rewrite Luigi's navigation footer

Lyczeq:
**Description**

- Added a Footer component with the same logic as it is in Luigi
- Added a feature flag to switch between Luigi and React navigation

**Related issue(s)**
See #1939


core-ui/src/sidebar/Footer/useGetBusolaVersion.ts
```diff
@@ -0,0 +1,63 @@
+import { useState, useEffect } from 'react';
+import { useTranslation, TFunction } from 'react-i18next';
+import jsyaml from 'js-yaml';
+
+const BUSOLA_GITHUB_REPOSITORY_LINK = 'https://github.com/kyma-project/busola';
+const BUSOLA_GITHUB_PULLS_LINK = 'https://github.com/kyma-project/busola';
+const BUSOLA_GITHUB_COMMMITS_LINK = 'https://github.com/kyma-project/busola';
```
wchrapka, line 7: triple MMM

core-ui/src/sidebar/Footer/Footer.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { useGetLegalLinks } from './useGetLegalLinks';
+
+export function Footer() {
+  const { t } = useGetTranslation();
+  const { githubLink, busolaVersion } = useGetBusolaVersionDetails();
+  const legalLinks = useGetLegalLinks();
+
+  return (
+    <footer>
+      <div>
```
wchrapka, line 13: this div is probably not needed

core-ui/src/sidebar/Footer/useGetLegalLinks.ts
```diff
@@ -0,0 +1,31 @@
...
+import { useMicrofrontendContext } from 'shared/contexts/MicrofrontendContext';
+
+type LegalLink = {
+  label: string;
+  link: string;
+};
+
+export const useGetLegalLinks = (): LegalLink[] => {
+  const { t, i18n } = useTranslation();
+  const { features } = useMicrofrontendContext() as any;
```
wchrapka, line 11: So, as the microfrontendContext is going to be dropped, the new components should subscribe to the recoil store. There is already a recoil atom for features `configFeaturesAtom` . Actually maybe I'll rewrite the features state today

core-ui/src/sidebar/Sidebar.js
```diff
@@ -2,11 +2,14 @@ import React from 'react';
...
 import { useMicrofrontendContext } from 'shared/contexts/MicrofrontendContext';
 import { ErrorBoundary } from 'shared/components/ErrorBoundary/ErrorBoundary';
+import { Footer } from './Footer/Footer';
 
 export const Sidebar = () => {
-  const { kubeconfig } = useMicrofrontendContext();
+  const { features } = useMicrofrontendContext();
+
+  // turn on the feature locally to work on the new Navigation
+  if (!features?.REACT_NAVIGATION?.isEnabled) return null;
```
wchrapka, line 11: this is temporary so this is okay.

core-ui/public/assets/version.yaml
```diff
@@ -0,0 +1 @@
+version: dev
```
wchrapka, line 1: what does that do? I think this works like the file is mounted by a config map on prod? But probably not yet. You can take the info if this is development or not without a requst by checking the global `process.env.NODE_ENV`. 

I'm not sure about this approach but talked with Piotr and agreed that as long as we are mounting the files in Luigi, we should take them from Luigi? Maybe this mount can be already moved to React's assets

core-ui/src/sidebar/Footer/useGetBusolaVersion.ts
```diff
@@ -0,0 +1,63 @@
...
+      const busolaVersion = await getBusolaVersion(t);
+      const githubLink = createGithubLink(busolaVersion);
+
+      setVersionDetails({
+        busolaVersion,
+        githubLink,
+      });
+    };
+
+    getVersion();
```
wchrapka, line 58:
```suggestion
    void getVersion();
```

core-ui/src/sidebar/Footer/useGetBusolaVersion.ts
```diff
@@ -0,0 +1,63 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
wchrapka, line 59: `t` has a constant identity so you don't need to disable eslint here.

resources/web/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
         - name: busola
-          image: eu.gcr.io/kyma-project/busola-web:PR-1974
+          image: eu.gcr.io/kyma-project/busola-web:PR-1969
```
wchrapka, line 19: It's a not a big task, why not style it all together?

wchrapka (changes requested):

core-ui/public/assets/version.yaml
Lyczeq: This is a temporary solution until we remove Luigi. More info [here](https://github.com/kyma-project/busola/issues/1924#issuecomment-1286885130).

wchrapka (approved):

PR #1978, opened 2022-10-24
Handle navigation custom nodes

wchrapka:
**Description**

 - Handles navigation custom nodes - adds them as 'additional' nodes. The alternative approach would have to have something like this in the resource configs: 

```
export const requiredFeatures = [
  configFeaturesNames.EXTENSIBILITY,
 { type: "ConfigMap", apiVersion: "" }  // check later if this is present
];
```
This might be an idea for the future if the number or dependent features would grow.

- Refactors navigation selectors

**Related issue(s)**

See https://github.com/kyma-project/busola/issues/1939


core-ui/src/components/BusolaExtensions/extensionsNode.ts
```diff
@@ -0,0 +1,10 @@
+export const extensionsNavNode = {
```
Wawrzyn321, line 1:
```suggestion
import { NavNode } from "state/types";

export const extensionsNavNode: NavNode = {
```

core-ui/src/components/CustomResources/customResourcesNode.ts
```diff
@@ -0,0 +1,12 @@
+import { Scope } from '../../state/types';
+
+export const createCustomResourcesNavNode = (scope: Scope) => ({
```
Wawrzyn321, line 3: Would you kindly add typing also here and in other similar nodes?

core-ui/src/state/navigation/sidebarNavigationNodesSelector.ts
```diff
@@ -0,0 +1,35 @@
...
+  defaultValue as defaultNamespaceName,
+} from '../activeNamespaceIdAtom';
+import { hasCurrentScope } from './filters/hasCurrentScope';
+import { partial } from 'lodash';
+
+export const sidebarNavigationNodesSelector: RecoilValueReadOnly<Category[]> = selector<
+  Category[]
+>({
+  key: 'scopedNavigationSelector',
+  get: async ({ get }) => {
```
Wawrzyn321, line 17:
```suggestion
  get: ({ get }) => {
```
?

core-ui/src/state/navigation/addAdditionalNodes.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as cmMetadata from 'resources/ConfigMaps';
+
+export const addAdditionalNodes = (
+  navNodes: NavNode[],
+  scope: Scope,
+  configFeatures: ConfigFeatureList,
+) => {
+  const extNavList = [...navNodes];
+
+  addResource(createClusterNode(scope), extNavList.length, extNavList);
```
Wawrzyn321, line 22: I'm not sure about this - I'd go for 2 separate constant nodes (one for Cluster Details, one for navigating back to clusterwide) instead of a function - they don't differ a lot now, but we'll probably add a navigation handler later.

wchrapka: agreed

Wawrzyn321 (approved):

PR #1979, opened 2022-10-25
update jsonata in forms

qbalukom:
- allow use of data sources in forms
- fix enum/visibility in forms
- update form variables to match display


core-ui/src/components/Extensibility/hooks/useJsonata.js
```diff
@@ -5,37 +5,44 @@ import { last, mapValues } from 'lodash';
 import { jsonataWrapper } from '../helpers/jsonataWrapper';
 import { DataSourcesContext } from '../contexts/DataSources';
 
-export function useJsonata(
```
Wawrzyn321, line 8: Could we add types here? I can see that we return a function that returns an array of 2 or sometimes 3 elements and also has `async` property, it would be easier to use if we had this stated explicitly.

core-ui/src/components/Extensibility/contexts/DataSources.js
```diff
@@ -112,14 +113,18 @@ export function DataSourcesContextProvider({ children, dataSources }) {
...
         intervals.current.push(
           setInterval(
             () => fetchResource(dataSource, dataSourceName, resource),
             REFETCH_INTERVAL,
           ),
         );
+
+        return firstFetch;
+      } else {
+        return Promise.resolve();
```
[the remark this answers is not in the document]
qbalukom: i think you're right. it's a bit of an oversight from an earlier version, and i didn't check for potential multiple access

core-ui/src/components/Extensibility/contexts/DataSources.tsx
```diff
@@ -0,0 +1,209 @@
...
+import { useObjectState } from 'shared/useObjectState';
+import * as jp from 'jsonpath';
+import { jsonataWrapper } from '../helpers/jsonataWrapper';
+
+export interface Resource {
+  metadata: {
+    name: string;
+    namespace: string;
+    labels: { [key: string]: string };
+    annotations: { [key: string]: string };
```
Wawrzyn321, line 13:
```suggestion
    labels: Record<string, string>;
    annotations: Record<string, string>;
```
Maybe that's a little bit nicer?

core-ui/src/shared/useObjectState.ts
```diff
@@ -0,0 +1,20 @@
...
+export type ObjectStateType<T> = [
+  T,
+  (key: string, value: any) => void,
+  Dispatch<SetStateAction<T>>,
+];
+
+export function useObjectState<T>(): ObjectStateType<T> {
+  const [objectCache, setObjectCache] = useState<T>({} as T);
+
+  const updateObjectCache = (key: string, value: any) => {
```
Wawrzyn321, line 12:
```suggestion
  const updateObjectCache = (key: string, value: T) => {
```

core-ui/src/components/Extensibility/contexts/DataSources.tsx
Wawrzyn321: And probably in other places.

core-ui/src/shared/useObjectState.ts
qbalukom: nope. because `T` is the main object, while `updateObjectCache` updates only a part of it. not sure if typescript has "one element of an object" type (afaik there is a type for subset, but that's not what's happening here)

Wawrzyn321 (approved):

PR #1982, opened 2022-10-25
Removing TW as mandatory code owners for MD files

valentinvieriu:
We will add them on a need basis, to avoid spamming them with not needed notofications

**Description**

- ...
- ...
- ...

**Related issue(s)**


NHingerl:
Thank you for your willingness to reduce our workload; but I don't think we should make TW review optional; especially because Busola is a customer-facing GUI.

PR #1987, opened 2022-10-26
Luigi separate secondary config

wchrapka:
**Description**

- separate dependencies rerendering atoms unnecessarily
- remove atom default values where they're not needed

**Related issue(s)**

See #1939


core-ui/src/state/authDataAtom.ts
```diff
@@ -4,12 +4,8 @@ export type AuthDataState =
   | { 'client-certificate-data': string; 'client-key-data': string }
   | {
       token: string;
-    }
-  | null;
```
Wawrzyn321, line 8: I'd allow null here - authData can be null when we are at clusters list.

core-ui/src/state/clusterAtom.ts
```diff
@@ -6,11 +6,8 @@ export type ClusterState = {
     'certificate-authority-data': string;
   };
   name: string;
-} | null;
-
-const defaultValue = null;
+};
```
Wawrzyn321, line 9: Same here - cluster may not be set.

core-ui/src/state/clusterConfigAtom.ts
```diff
@@ -3,13 +3,10 @@ import { atom, RecoilState } from 'recoil';
 export type ClusterConfigState = {
   requiresCA: boolean;
   storage: 'localStorage' | 'sessionStorage' | string;
-} | null;
-
-const defaultValue: ClusterConfigState = null;
+};
```
Wawrzyn321, line 6: -||-

Wawrzyn321 (approved):

PR #1998, opened 2022-11-02
Add overwrite flag

akucharska:
**Description**

- Add overwrite flag that can disable overwriting values when the field is hidden

**Related issue(s)**


docs/extensibility/form-section.md
```diff
@@ -45,6 +46,16 @@ If you target elements of an array rather than the array itself, you can use the
       name: Item name
     - path: service.url
     - path: service.port
+    - path: service.number
```
qbalukom, line 49: you added a mutually exclusive `visibility` and yet you didn't include `overwrite` here. Which should be used.
we could use an example with single `visibility` and no `overwrite` in addition to it as well

core-ui/src/components/Extensibility/plugins/VisibilityHandler.js
```diff
@@ -20,6 +20,10 @@ export function VisibilityHandler({
 
   // rule.visibility won't work for "var", we must use schema.get('visibility')
   const visibilityFormula = schema.get('visibility');
+  const overwrite =
+    typeof schema.get('overwrite') !== 'undefined'
+      ? schema.get('overwrite')
+      : true;
```
qbalukom, line 26: wouldn't `??` be enough here?
```suggestion
  const overwrite = schema.get('overwrite') ?? true;
```

qbalukom (approved):

docs/extensibility/form-section.md
```diff
@@ -33,6 +33,7 @@ If you target elements of an array rather than the array itself, you can use the
 - **simple** - parameter used to display the simple form. It is `false` by default.
 - **advanced** - parameter used to display the advanced form. It is `true` by default.
 - **visibility** - a [JSONata](jsonata.md) expression controlling the visibility of the element.
+- **overwrite** - parameter used to disable the overwriting (clearing) of hidden fields. Used together with **visibility**, defaults to `true`. **NOTE:** it is recommended to set **overwrite** to `false` when defining fields with the same `path` and different **visibility** conditions.
```
grego952, line 36:
```suggestion
- **overwrite** - parameter used to disable the overwriting (clearing) of hidden fields. Used together with **visibility**, defaults to `true`. 
**NOTE:** it is recommended to set **overwrite** to `false` when defining fields with the same `path` and different **visibility** conditions.
```

grego952 (approved):

qbalukom (approved):

grego952 (approved):

PR #2011, opened 2022-11-07
Readonly name

akucharska:
**Description**

- Set main name widget in read-only mode when resource is edited
- Fix error with wrong import

**Related issue(s)**


core-ui/src/components/Extensibility/components-form/NameRenderer.js
```diff
@@ -16,11 +16,13 @@ export function NameRenderer({
 }) {
   const { t: tExt } = useGetTranslation();
   const extraPaths = schema.get('extraPaths') || [];
+  const editMode = schema.get('editMode') || false;
 
   return (
     <K8sNameField
       value={value}
       kind={resource.kind}
+      readOnly={editMode}
```
mrCherry97, line 25: We don't need `editMode` in schema, if I use `widget: Name` in schema for input, it is disabling input in Create too. 

image

We need that only in `ExtensibilityCreate` and `ResourceDetails`, like we had it in "old" implementation.

mrCherry97 (dismissed):
If we want to allow user to decide what input can be edited, we need to make it in different way in my opinion.

dariadomagala (changes requested):
All examples are in light mode, we should stick to one approach

grego952 (approved):

PR #2028, opened 2022-11-15
Replace Serverless with the Extension

### LINKED ISSUE #1953: Replace Serverles with the extension
**Description**

We replace our current implementation with the extensibility option.

**Acceptance Criteria**

- [ ] Validate with Product Owner if the current implementation is correct
- [x] Replace the current implementation with extensibility
- [ ] Deploy to stage and Test all the edge cases on stage
- [ ] Deploy to production
- [x] Remove old codebase - this includes all not needed views
- [x] Adjust Tests to use extensions
- [ ] Adjust Kyma documentation

### LINKED ISSUE #2031: EXT: Presets in Functions don't work
**Description**
The Functions form presets created by the extensibility don't work. It successfully pastes the runtime information in the yaml but the source code of the function is missing. It is probably caused by the `visibility` flag that uses `sourceType` var. 

**Expected result**
Source code of the function's preset should be changed succesfully.

**Actual result**
When clicking on one of the presets (Node or Python) the runtime is changed, but source code is missing.

**Steps to reproduce**
- Go to the function's create form that uses extensibility. Go to `YAML`.
- Choose the preset you like. 
- New runtime value should appear, but the source object is empty.

**Troubleshooting**
It is probably caused by the `visibility` flag that uses `sourceType` var. 

Presets property:
```yaml
  presets: |-
    - name: Node.js Function
      default: true
      value:
        spec:
          runtime: nodejs14
          source:
            inline:
              source: |-
                module.exports = {
                  main: async function (event, context) {
                    const message = `Hello World`
                      + ` from the Kyma Function ${context["function-name"]}`
                      + ` running on ${context.runtime}!`;
                    console.log(message);
                    return message;
                  }
                }
    - name: Python Function
      value:
        spec:
          runtime: python39
          source:
            inline:
              source: |-
                def main(event, context):
                    message = "Hello World from the Kyma Function "+context['function-name']+" running on "+context['runtime']+ "!";
                    print(message)
                    return message
```

The piece of the code responsible for CodeEditor that causes the problem:
```yaml
    - path: spec.source.inline
      visibility: $sourceType = 'Inline' # the cause
      children:
        - path: source
          widget: CodeEditor
          language: '$contains($root.spec.runtime, "node") ? "javascript" : "python"'
        - path: dependencies
          widget: CodeEditor
          language: '$contains($root.spec.runtime, "node") ? "javascript" : "plaintext"'
```

Lyczeq:
**Description**

- Remove Functions codebase
- Remove english and polish translations
- Adjust functions, apirules and subscriptions tests ( functions have two ResourceList widgets of those two resources)
- Modified copyToMonaco Cypress Command to be able to choose Monaco (Function view has two Monaco editors for Source and Dependencies)
- Remove Add New Workload Button from Namespace Overview

**Related issue(s)**
Resolves #1953
Fixes #2031


tests/tests/namespace/run-prepare-functions.spec.js
```diff
@@ -7,26 +7,56 @@ const API_RULE_AND_FUNCTION_NAME = 'in-cluster-eventing-publisher';
...
+    cy.createFunction(
+      API_RULE_AND_FUNCTION_NAME,
+      'fixtures/in-cluster-eventing-publisher.js',
+      'fixtures/in-cluster-eventing-publisher-dependencies.json',
+    );
+  });
+
+  // edit test case is the last one because of the following error:
+  // "the object has been modified; please apply your changes to the latest
+  // version abnd try again"
```
IwonaLanger, line 42:
```suggestion
  // version and try again"
```

tests/tests/namespace/run-prepare-functions.spec.js
```diff
@@ -7,26 +7,56 @@ const API_RULE_AND_FUNCTION_NAME = 'in-cluster-eventing-publisher';
...
+  // we need to wait until the Function isn't modified in the meantime
```
IwonaLanger, line 43: I'm not sure what you mean here. Maybe this:  in the meantime we need to wait until the Function IS modified

Lyczeq: I will change it to: "we must wait for the function to remain unchanged". What do you think?

IwonaLanger (approved):

dariadomagala (changes requested):
1. details don't show the code
2. switching from node to python doesn't change the editor value
3. apirules should use the new version already (v1beta1)

dariadomagala (dismissed):

IwonaLanger (approved):

dariadomagala (dismissed):

dariadomagala (dismissed):

dariadomagala (approved):

grego952 (approved):

Issue #1964, opened 2022-10-19
Remove current ISTIO implementation being replaced by extensions


valentinvieriu:
**Description**

We replace our current implementation with the extensibility option.

**Acceptance Criteria**

- [ ] Remove old codebase

PR #2045, opened 2022-11-21
remove destination rules from core-ui

mrCherry97:
**Description**

- remove old codebase,
- adjust tests,
- fix DR config map (mostly jsonata changes),
- remove DR `mjs` files

**Related issue(s)**
part of #1964


IwonaLanger (approved):

Lyczeq (commented):
In `core-ui/src/hooks/useMessageList.js` I found an usage of `DestinationRules` and some other Istio resources. This is kept in `RESOURCE_PATH` map that is used in `formatInvolvedObject` component(?). Since those resources aren't in busola anymore we should find different way on how to handle navigation in Events view. I would propose to use a hook/state that keeps the information about the Nodes and then filter whether the involvedObject is there and then navigate to that.

Lyczeq:
> In `core-ui/src/hooks/useMessageList.js` I found an usage of `DestinationRules` and some other Istio resources. This is kept in `RESOURCE_PATH` map that is used in `formatInvolvedObject` component(?). Since those resources aren't in busola anymore we should find different way on how to handle navigation in Events view. I would propose to use a hook/state that keeps the information about the Nodes and then filter whether the involvedObject is there and then navigate to that.

We've decided to adjust that when removing Luigi.

PR #2082, opened 2022-11-30
Apirules never ending story

akucharska:
**Description**

- Remove create API Rule and Subscription from function and service details
- fix tests
- default accessStrategy when adding a new rule
- port autocompletion - service on rule level

**Related issue(s)**


mrCherry97 (approved):
/woof

PR #2083, opened 2022-11-30
Luigiless links - sidenav

qbalukom:


dariadomagala (changes requested):
back to cluster details doesn't work, the `isNodeSelected` function doesn't work

PR #2120, opened 2022-12-07
Luigiless remove navigateFn/fixedUrl

qbalukom:


core-ui/src/resources/StorageClasses/PersistentVolumeClaimsList.js
```diff
@@ -1,20 +1,11 @@
 import React from 'react';
-import LuigiClient from '@luigi-project/client';
 
+import { useUrl } from 'hooks/useUrl';
```
akucharska, line 3: Do we use it anywhere?

PR #2176, opened 2022-12-15
fix dynamic values even more

qbalukom:
1. do not reset values so visibility is not changed
2. do not fire triggers when switching from yaml to form


akucharska (approved):
/meow

PR #2178, opened 2022-12-15
Fix double navigation

akucharska:
**Description**

- Fix double navigation for apirules
- fix oauth2clients navigation label

**Related issue(s)**


tests/support/views/api-rules-view.js
```diff
@@ -4,7 +4,6 @@ Cypress.Commands.add(
     // Remove eq(1) after fixing doubled catagories
```
mrCherry97, line 4: This comment is unnecessary right now

mrCherry97 (changes requested):
/woof

mrCherry97 (approved):
/woof

PR #2184, opened 2022-12-16
fix: displaying version, leftNav styling, adjusting to smaller device, topNav styling

mrCherry97:
**Description**

- fetching `version.yaml`
- remove unnecessary file in assets
- fix CSS for leftNav,
- fix topNav,
- add useSidebarCondensed for smaller device.

**Related issue(s)**

### DIFF
--- a/core-ui/src/sidebar/useSidebarCondensed.tsx
+++ b/core-ui/src/sidebar/useSidebarCondensed.tsx
@@ -1,0 +1,35 @@
+import { useState, useEffect } from 'react';
+import { useSetRecoilState } from 'recoil';
+import { isSidebarCondensedState } from 'state/preferences/isSidebarCondensedAtom';
+
+function getWindowDimensions() {
+  const { innerWidth: width } = window;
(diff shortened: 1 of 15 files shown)


core-ui/src/sidebar/useSidebarCondensed.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return windowDimensions;
+}
+
+export default function useSidebarCondensed() {
+  const width = useWindowDimensions();
+
+  const setSidebarCondensed = useSetRecoilState(isSidebarCondensedState);
+
+  if (width <= 900) {
```
qbalukom, line 32: not sure about mixing block styles. I'd suggest keeping consistent.
Either:
```
if (foo) {
  option1();
} else {
  option2();
}
```
or
```
if (foo)
  option1();
else
  option2();
```
or
```
if (foo) option1();
else option2();
```
but not a mix

qbalukom (approved):

PR #2203, opened 2022-12-19
Gardener login

Wawrzyn321:
**Description**

- Add Gardener login feature. To be tested with your Gardener kubeconfigs.

* We need to have login with Gardener as a feature
* This feature is toggled off by default
* The feature accepts as configuration the kubeconfig file that is needed to login the user - this kubeconfig file uses OIDC
* the kubeconfigfile does not contain namespace, so we load all projects that the user has access to
* we don't have yet an intermediary step, where the user has the chance of selecting the available clusters- if there is time for this, we ld be nice to have this

**Related issue(s)**


qbalukom (dismissed):

docs/features.md
```diff
@@ -108,7 +108,17 @@ EXTENSIBILITY:
             link: https://github.com/kyma-project/busola
   ```
 
-* **GZIP** – is used to indicate whether a response from the backend server should be compressed or not.
+- **GARDENER_LOGIN** - is used to enable or disable the option of logging in with the Gardener kubeconfig. If enabled, you must set **kubeconfig** parameter to a valid kubeconfig object.
```
grego952, line 111:
```suggestion
- **GARDENER_LOGIN** - is used to enable or disable the option of logging in with the Gardener kubeconfig. If enabled, you must set the **kubeconfig** parameter to a valid kubeconfig object.
```

grego952 (approved):

mrCherry97 (approved):

PR #2231, opened 2022-12-28
Fix getting schemas after cluster change

akucharska:
**Description**

- Fix getting schemas after cluster change

**Related issue(s)**


core-ui/src/state/authDataAtom.ts
```diff
@@ -108,6 +109,7 @@ export function useAuthHandler() {
   const cluster = useRecoilValue(clusterState);
   const setAuth = useSetRecoilState(authDataState);
   const navigate = useNavigate();
+  const setLastFetched = useSetRecoilState(openapiLastFetchedState);
 
   useEffect(() => {
     console.log(
```
mrCherry97, line 115: can we removed that clg or we need to open task for that?

mrCherry97 (approved):

mrCherry97 (approved):

PR #2284, opened 2023-01-16
add "decode" option to panel

qbalukom:


src/components/Extensibility/components/Panel.js
```diff
@@ -44,6 +57,21 @@ export function Panel({
               />
             ))
           : null}
+        {decodable && (
+          <LayoutPanel.Actions>
+            <Button
+              option="transparent"
+              glyph={isDecoded ? 'hide' : 'show'}
+              //disabled={!secret?.data}
```
mrCherry97, line 65: probably it should be uncommented

mrCherry97 (changes requested):
Code looks fine, I will test it a little bit and make re-review

src/components/Extensibility/components/Panel.js
qbalukom: probably should be removed

mrCherry97 (dismissed):

docs/extensibility/display-section.md
```diff
@@ -524,6 +524,7 @@ Panel widgets render an object as a separate panel with its own title (based on
 - **header** - an optional array that allows you to, for example, display labels in the panel header.
 - **disablePadding** - an optional boolean which disables the padding inside the panel body.
 - **description** - displays a custom description on the resource list page. It can contain links. If the **translations** section has a translation entry with the ID that is the same as the **description** string, the translation is used.
+- **decodable** - decodes the values of all the children using base64 - has to be used together with the **source** parameter.
```
grego952, line 527:
```suggestion
- **decodable** - decodes the values of all the children using base64 - must be used together with the **source** parameter.
```

public/schemas/schema-details.yaml
```diff
@@ -259,6 +259,9 @@ $widgets:
           type: boolean
         description:
           type: string
+        decodable:
+          type: boolean
+          description: Allow decoding of child values using base64
```
grego952, line 264:
```suggestion
          description: Allows decoding of child values using base64
```

mrCherry97 (approved):

PR #2287, opened 2023-01-17
fix: navigate to namespace with limited access

Lyczeq:
**Description**

- fix navigating to namespace with limited access
- fix in displaying `IncorrectPath` component for a second when refreshing extensibility views
- add test

How to test?
- Create a namespace only user with `cluster-admin` role
- Upload a new kubeconfig
- Go to e.g. Deployments view
- On main, Incorrect Path is shown but shouldn't

**Related issue(s)**


qbalukom (changes requested):

src/components/App/NamespaceRoutes.tsx
```diff
@@ -4,24 +4,28 @@ import { useTranslation } from 'react-i18next';
...
-  const namespaces = useRecoilValue(namespacesState);
 
-  if (!namespaces.includes(namespaceId as string))
+  const { error } = useGet(`/api/v1/namespaces/${namespaceId}`, {
+    skip: false,
+    pollingInterval: 0,
+    onDataReceived: () => {},
+  });
+
+  if (error)
```
qbalukom, line 28: if the contents of an if don't fit in a single line i would suggest going with 
```suggestion
  if (error) {
```
instead

qbalukom (approved):

PR #2304, opened 2023-01-24
More triggers

qbalukom:
add an "init" trigger on:
- opening the form
- choosing a preset
- resetting the form


mrCherry97 (changes requested):
looks good, first choosing preset is working properly, but changing between presets is not updating spec

mrCherry97 (approved):

akucharska (dismissed):
Setting port in API Rules does not work in this branch

mrCherry97 (approved):

PR #2306, opened 2023-01-24
fix: namespace dropdown switch context

### LINKED ISSUE #2295: Switching Namespace changes context
Switching namespace should not change the context. If I'm on Functions on namespace x and switch to namespace y the function context should stay. It should only revert to the namespace overview, only if the route does not exist

Lyczeq:
**Description**

- fix bug of switching namespace context
- add tests for Namespace Dropdown

**Related issue(s)**
Closes #2295


src/header/NamespaceDropdown/NamespaceDropdown.tsx
```diff
@@ -1,18 +1,49 @@
...
+  const switchNamespace = (namespaceName: string) => {
+    if (!pathname.includes('/namespaces/'))
+      return clusterUrl(`namespaces/${namespaceName}`);
+
+    const pathElements = pathname
+      .substring(pathname.indexOf('namespaces/'))
+      .split('/');
+
+    const resourceType = pathElements[2];
+    const resourceName = pathElements[3];
```
qbalukom, line 25: i don't like this manual splitting - i think it would be better to use `useMatch` - similar to how it's used inside `useUrl`. Something similar to this:
```suggestion
    const {resourceType, resourceName } = useMatch({ path: '/cluster/:cluster/namespaces/:namespace/:resourceType/:resourceName', end: false })?.params
```

src/header/NamespaceDropdown/NamespaceDropdown.tsx
```diff
@@ -1,18 +1,49 @@
...
+
+    if (!resourceName) return pathname.replace(namespace, namespaceName);
```
qbalukom, line 27: is the `resourceName` handler even needed here?
it seems to work just fine with just the `namespaceUrl` below

src/state/types.ts
```diff
@@ -68,14 +68,22 @@ export type ExtResource = {
...
-  clusterUrl: (path: string) => string;
-  namespaceUrl: (path: string) => string;
-  scopedUrl: (path: string) => string;
-  resourceListUrl: (path: string) => string;
-  resourceUrl: (path: string) => string;
+  clusterUrl: UrlFunction;
+  namespaceUrl: UrlFunction;
+  scopedUrl: UrlFunction;
+  resourceListUrl: UrlFunction;
+  resourceUrl: UrlFunction;
```
qbalukom, line 86: technically `resourceListUrl` and `resourceUrl` accept the resource as the first parameter (currently just defined as `any` - but should have proper `Resource` type in the future), not string.

qbalukom (changes requested):

src/state/types.ts
Lyczeq: Technically, we have `K8sResource` type so I will change it to this one

src/header/NamespaceDropdown/NamespaceDropdown.tsx
```diff
@@ -1,18 +1,46 @@
...
+      path: '/cluster/:cluster/namespaces/:namespace/:resourceType',
+      end: false,
+    })?.params ?? {};
+
+  const switchNamespace = (namespaceName: string) => {
+    if (!namespace) return clusterUrl(`namespaces/${namespaceName}`);
+
+    return namespaceUrl(resourceType, {
+      namespace: namespaceName,
+    });
```
qbalukom, line 26: do we really need that `if` here?
won't just
```suggestion
    return namespaceUrl(resourceType, {
      namespace: namespaceName,
    });
```
work?
unless that adds `undefined` to the url, but that should be handled in `useUrl`
(use `path ?? ''` instead of just `path` in the url string)

src/header/NamespaceDropdown/NamespaceDropdown.tsx
```diff
@@ -1,18 +1,46 @@
...
+  const namespaces = useRecoilValue(namespacesState);
+  const { clusterUrl, namespace, namespaceUrl } = useUrl();
+
+  const { resourceType = '' } =
+    useMatch({
+      path: '/cluster/:cluster/namespaces/:namespace/:resourceType',
+      end: false,
+    })?.params ?? {};
+
+  const switchNamespace = (namespaceName: string) => {
```
qbalukom, line 21: `switchNamespace` is a rather strange name for a function that simply returns a url instead of doing any switching. Although with the comment below this function might not be needed at all

qbalukom (changes requested):

qbalukom (approved):

PR #2335, opened 2023-02-03
fix: badge highlights jsonata function usage 

### LINKED ISSUE #2051: Functions follow-up
**Description**

We need to improve the functions view with the extensibility configuration.

- [x] ❗  Change the source code after changing the language (not on edit).
- [x] Remove default runtime language
- [x] Add the `code` editor also to the simple view in Edit mode.
- [x] Remove the Runtime Image Override field, it's not important.
- [x] New version of Functions has the `profile` option when choosing the limits and requests. We should show the `profile` dropdowns in the simple view, and current config options in the advanced view.
- [x] Add auth details for the git repos config on the details view.
- [x] Add information about running replicas to the header on the details view. (for example 0/2 as running/desired)
- [x] Hide scaling options if doesn't exist, it's not always present in the newest version.
- [ ] Add the link to logs.
- [x] Move functions higher in the left nav (not sure if now, maybe after Luigi removal)

**Reasons**

Improve users experience.

**Attachments**

Lyczeq:
**Description**

- change the old way of jsonata usage to the new one that returns array of value and an error

**Related issue(s)**
Closes #2051

### DIFF
--- a/src/components/Extensibility/components/Badge.js
+++ b/src/components/Extensibility/components/Badge.js
@@ -32,8 +32,8 @@ export function Badge({
       if (Array.isArray(rule)) {
         return rule.includes(value);
       } else {
-        const [doesMatch, error] = jsonata(rule);
-        return error ? false : doesMatch;
+        const [doesMatch, matchError] = jsonata(rule);
+        return matchError ? false : doesMatch;
       }
     });
     if (match) {
--- a/src/components/Extensibility/components/Badge.js
+++ b/src/components/Extensibility/components/Badge.js
@@ -32,12 +32,8 @@ export function Badge({
       if (Array.isArray(rule)) {
         return rule.includes(value);
       } else {
-        try {
-          return jsonata(rule);
-        } catch (e) {
-          console.warn(`invalid rule: ${rule}`, e);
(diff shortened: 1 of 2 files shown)


src/components/Extensibility/components/Badge.js
```diff
@@ -32,12 +32,8 @@ export function Badge({
       if (Array.isArray(rule)) {
         return rule.includes(value);
       } else {
-        try {
-          return jsonata(rule);
-        } catch (e) {
-          console.warn(`invalid rule: ${rule}`, e);
-          return null;
-        }
+        const [doesMatch, error] = jsonata(rule);
```
mrCherry97, line 35: rename `error` for some `doesMatchError` or smth like this

src/components/Extensibility/components/Badge.js
```diff
@@ -32,12 +32,8 @@ export function Badge({
...
+        return error ? false : doesMatch;
```
mrCherry97, line 36: we are returning, but not displaying info about it, please add displaying error

mrCherry97 (changes requested):

PR #2354, opened 2023-02-17
Multi namespace

qbalukom:
- add "all namespaces" option  in namespace dropdown
- display resources from all namespaces when in "all namespaces" mode
- show namespace resources from all namespaces in command palette
- show "all namespaces" option when typing "ns all", "ns -a" or "ns *" in command palette

### DIFF
--- a/public/i18n/en.yaml
+++ b/public/i18n/en.yaml
@@ -240,7 +240,7 @@ command-palette:
     open-preferences: Open Preferences
     resource-aliases: Built-in Resource Aliases
     short-help: Type to search, e.g. "ns default" or "pod nginx"
-    all-namespaces: List resources from all namespaces
+    all-namespaces: List resources from all Namespaces
   item-actions:
     navigate: Navigate
     open: Open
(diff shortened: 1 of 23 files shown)


mrCherry97 (changes requested):
As we have a talk, we need to catch something like this error 
image

mrCherry97 (changes requested):
typing `cmk +k` and `ns all` navigates all the time to namespace overview, when I am on for example Pods view and type in command pallet `ns all` I need to have two options: switch (redirect to pods view for all namespaces) and navigate (redirect to namespace overview)

mrCherry97 (approved):

mrCherry97 (approved):

public/i18n/en.yaml
```diff
@@ -240,6 +240,7 @@ command-palette:
     open-preferences: Open Preferences
     resource-aliases: Built-in Resource Aliases
     short-help: Type to search, e.g. "ns default" or "pod nginx"
+    all-namespaces: List resources from all namespaces
```
NHingerl, line 243:
```suggestion
    all-namespaces: List resources from all Namespaces
```

Ignore this if we don't capitalize "Namespace" in Busola.

mrCherry97 (dismissed):

NHingerl (dismissed):

mrCherry97 (dismissed):

NHingerl (approved):

NHingerl (approved):

mrCherry97 (approved):

PR #2379, opened 2023-03-08
Add TAG_DEFAULT_TAG build-arg to Dockerfiles

Ressetkk:
In order to migrate to image-builder and remove privileged mode from build jobs `sed`'ing needs to be moved into the Dockerfile with usage of `--build-arg`.

Additionally update getGithubLink logic to parse new tag format. This should be safe and always return at least a 1 object array.

https://github.com/kyma-project/test-infra/issues/5891

### DIFF
--- a/src/sidebar/Footer/useGetBusolaVersion.ts
+++ b/src/sidebar/Footer/useGetBusolaVersion.ts
@@ -16,8 +16,9 @@ function createGithubLink(version: string): string {
     if (version.toString().startsWith('PR-')) {
       return `${BUSOLA_GITHUB_LINKS.PULLS}/${version.slice(3)}`;
     }
-
-    return `${BUSOLA_GITHUB_LINKS.COMMITS}/${version}`;
+    // if the string doesn't match regex, returns full string anyway
+    const commit = version.replace(/v[0-9]-/, '');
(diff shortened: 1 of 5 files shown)


Ressetkk:
PR is ready for review, needs to be aligned with https://github.com/kyma-project/test-infra/pull/7137


However, this implementation should work fine with current implementation.Replace here should replace the first part of the tag, leaving only commit SHA.

Though it's not ideal and probably changing the logic by explicitly defining commit id as env variable would be better, than parsing it from the tag...

/hold

Ressetkk:
PR is ready for review
/unhold

src/sidebar/Footer/useGetBusolaVersion.ts
```diff
@@ -16,8 +16,9 @@ function createGithubLink(version: string): string {
     if (version.toString().startsWith('PR-')) {
       return `${BUSOLA_GITHUB_LINKS.PULLS}/${version.slice(3)}`;
     }
-
-    return `${BUSOLA_GITHUB_LINKS.COMMITS}/${version}`;
+    // if the string doesn't match regex, returns full string anyway
+    const commit = version.replace(/v[0-9]-/, '');
```
valentinvieriu, line 20: Can you explain when we have this situation that the hash contains vxx- ?

valentinvieriu (approved):

PR #2384, opened 2023-03-10
 debugonata

qbalukom:


src/components/Extensibility/hooks/useDebugger.scss
```diff
@@ -0,0 +1,46 @@
+.debugonata {
+  width: auto;
```
mrCherry97, line 2: I will suggest to not make this auto, for better UX something like 70% could be better

src/components/Extensibility/hooks/useDebugger.scss
```diff
@@ -0,0 +1,46 @@
...
+    dt {
+      font-weight: bold;
+    }
+  }
+
+  .vars,
+  .data-sources {
+    padding-left: 1rem;
+
+    dd {
```
mrCherry97, line 30: maybe good option will be to make background for YAML code as sap-neutral or something like this, because right now it's unreadable

src/components/Extensibility/hooks/useDebugger.scss
```diff
@@ -0,0 +1,46 @@
+.debugonata {
+  width: auto;
+  max-height: 50vh !important;
+
+  h1 {
+    font-weight: bold;
+    border-bottom: 1px solid var(--sapTextColor);
+    padding: 0.5rem;
+  }
```
mrCherry97, line 9: all texts are very small and unreadable for me, could we adjust font size to the rest of busola?

src/components/Extensibility/hooks/useDebugger.tsx
```diff
@@ -0,0 +1,197 @@
...
+              .filter(([name, data]: [string, any]) => data.query)
+              .map(([name, data]: [string, any]) => (
+                <DebugonataQuery name={name} data={data} />
+              ))}
+          </div>
+        ))}
+      </ul>
+    </>
+  );
+}
```
mrCherry97, line 143: could you move some functions to separated files to have better file organisation? there is a lot of mess right now and files are too big with too much functionalities inside

src/components/Preferences/Preferences.tsx
```diff
@@ -104,6 +125,16 @@ export function Preferences() {
             <ResourcesValidationSettings />
           </div>
         </VerticalTabs.Content>
+        <VerticalTabs.Content id={3}>
+          <div>
+            {isExtensibilityEnabled && (
```
mrCherry97, line 130: probably this `isExtensibilityEnabled` should be higher before `<VerticalTabs.Content id={3}>`

src/components/Preferences/Preferences.tsx
```diff
@@ -104,6 +125,16 @@ export function Preferences() {
...
+              <Toggle
```
mrCherry97, line 131: Toggle name could provide a little bit misunderstanding, could you change this for something like `DevelopmentSettings` or `DebugonataSettings` or something like this as we have for namespaceSetting etc

mrCherry97 (changes requested):
Please check my comments and on gateways after opening debugger whole view brakes

PR #2440, opened 2023-04-17
Fix bug not required parent

chriskari:
**Description**

- added required flag to FormGroup
- added required flag to ResourceRenderer
- implemented custom function for form validation instead of built-in checkValidity-function, which produced wrong results

**Related issue(s)**


resources/web/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
         - name: busola
-          image: europe-docker.pkg.dev/kyma-project/dev/busola-web:PR-2454
+          image: eu.gcr.io/kyma-project/busola-web:PR-2440
```
mrCherry97, line 19:
```suggestion
          image: europe-docker.pkg.dev/kyma-project/dev/busola-web:PR-2440
```

src/components/Extensibility/components-form/CollapsibleRenderer.js
```diff
@@ -25,6 +25,7 @@ export function CollapsibleRenderer({
       title={tFromStoreKeys(storeKeys, schema)}
       defaultOpen={defaultOpen}
       nestingLevel={nestingLevel}
+      required={String(schema.get('required')) === 'true'}
```
mrCherry97, line 28: get this flag outside of this, like we have for `defaultOpen` for example, and probably you don't need thus `String()`

mrCherry97 (changes requested):

src/components/Extensibility/components-form/CollapsibleRenderer.js
```diff
@@ -19,12 +19,14 @@ export function CollapsibleRenderer({
   const gridTemplateColumns = `repeat(${columns}, 1fr)`;
 
   const defaultOpen = schema.get('defaultExpanded');
+  const required = schema.get('required') === true;
```
mrCherry97, line 22:
```suggestion
  const required = schema.get('required');
```

mrCherry97 (changes requested):

PR #2498, opened 2023-06-14
Use webcomponents in side navigation

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


tests/integration/support/generateRandomString.js
```diff
@@ -0,0 +1,9 @@
+export function generateRandomString(length) {
+  let result = '';
+  const characters = 'abcdefghijklmnopqrstuvwxyz';
+  const charactersLength = characters.length;
+  for (let i = 0; i < length; i++) {
+    result += characters.charAt(Math.floor(Math.random() * charactersLength));
+  }
+  return result;
+}
```
mrCherry97: this function is not used anywhere, we need that?

mrCherry97 (changes requested):
when I go to for example cluster role binding item in sidenav is marked
image

but when I click for example on storage to close this category, crb should be still marked, but it's not
image

src/sidebar/CategoryItem.tsx
```diff
@@ -1,31 +1,36 @@
...
-  const handleAddExpandedCategory = () => {
-    if (expanded) {
+  const handleAddExpandedCategory = (e: Ui5CustomEvent) => {
+    e.preventDefault();
+
+    console.log(
+      'handleAddExpandedCategory',
+      categoryName,
+      categoryName === e.target.parentElement?.getAttribute('text'),
+    );
```
mrCherry97, line 31: Unnecesary console.log

mrCherry97 (changes requested):
navigation for small device is not working, when I click for example on extensions then I'm not redirected anywhere
image


Sidenav for small device is too small or there is unnecessary margin somewhere
image
old view, how it should look:
image

Akatuoro:
> when I go to for example cluster role binding item in sidenav is marked
> but when I click for example on storage to close this category, crb should be still marked, but it's not

This is happening because of the default behaviour in the `SideNavigation` component.
On click (‘selection’), it sets every element’s `selected` property to false except the newly selected one:

https://github.com/SAP/ui5-webcomponents/blob/main/packages/fiori/src/SideNavigation.ts#L216C3-L216C3

When clicking on a category item, this gets overridden afterwards by `selected={false}`:

https://github.com/kyma-project/busola/pull/2498/files#diff-84924c1d7c5cbabdf412d1a40eebc3dfb1e736aaefb122a1a264d167e61e65c8R48

The solution is to register an event handler and prevent the default:
`<SideNavigation collapsed={isSidebarCondensed} onSelectionChange={(e) => e.preventDefault()}>`

mrCherry97 (approved):

PR #2514, opened 2023-07-10
Dynamic loading of web-components

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


mrCherry97 (changes requested):
Custom components are not working with vars (only with path).

We need to have docs for that.

The formatting of components is not working correctly, every component should be rendered in his line and I need to be able to provide the name. After that Name should be rendered and the component needs to be rendered on the right. For example to `Chechbox` I provided the name which is not rendered anywhere and my second custom components should be in next line.
image

Functionality is "ready" only for form components, the question is if we want the same for normal details components.

Is it possible to pass only values to the component and not render a few times?
Use case:
I want to use the UI5 web component called `MultiComboBox`, but when I will make it for module names, then it is rendered as many times as many inputs as I have.

src/styles/index.scss
```diff
@@ -221,45 +224,47 @@ input[type='checkbox'] {
...
+  src: url('~@sap-theming/theming-base-content/content/Base/baseLib/baseTheme/fonts/SAP-icons.woff')
     format('woff');
   font-weight: normal;
   font-style: normal;
 }
 
 @font-face {
   font-family: 'BusinessSuiteInAppSymbols';
-  src: url('~@sap-theming/theming-base-content/content/Base/baseLib/sap_fiori_3/fonts/BusinessSuiteInAppSymbols.woff')
+  src: url('~@sap-theming/theming-base-content/content/Base/baseLib/baseTheme/fonts/BusinessSuiteInAppSymbols.woff')
```
mrCherry97, line 259: Also this is a problem, because we need to migrate to UI5 everything.

mrCherry97 (changes requested):
Changing theme after calling a custom webcomponent, but we had a discuss about it.

PR #2530, opened 2023-07-21
Upgrade wizard with checkboxss

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


Akatuoro (commented):
There could be a few index discrepancies:

The link uses `resource.spec.modules[index]` while the Checkbox setter uses `resource.spec.modules.findIndex(module => module.name === name)`

If we assume that each module has a unique name, I think it would be good to fix the module index and the module object for consistency. E.g.:
```js
const moduleIndex = resource?.spec?.modules?
    resource.spec.modules.findIndex(module => module.name === name) :
    index;
const module = resource?.spec?.modules?.[moduleIndex];
```


If we assume that `moduleTemplate` has a unique name, then we could also fixate that. But for the `channelTests`, it seems that there could be multiple module templates with the same name, but different channels.

If so, then it'd be good to fixate the specific module template that matches the currently selected channel, e.g.:
```js
const channelModuleTemplate = parsedOptions?.moduleTemplates?.find(moduleTemplate => {
    if (moduleTemplate?.metadata?.labels['operator.kyma-project.io/module-name'] === name && moduleTemplate?.spec?.channel === channel) return moduleTemplate;
    return null;
});
```

The `isBeta` variable is currently based on any `moduleTemplate` with the module name, not considering the selected channel. From my understanding, this should be changed to consider both the name and channel.

src/components/Extensibility/components-form/Modules/Modules.js
```diff
@@ -0,0 +1,226 @@
...
+          });
+
+        return channelTest.push({
+          text: moduleTemplate.spec.channel,
+          additionalText: '',
+        });
+      } else {
+        return '';
+      }
+    });
```
Akatuoro, line 108: Could be simplified to a filter and map
```suggestion
    const channelTest = parsedOptions?.moduleTemplates?.filter(moduleTemplate => moduleTemplate?.metadata?.labels['operator.kyma-project.io/module-name'] === name).map(moduleTemplate => {
        const additionalText = moduleTemplate?.spec?.descriptor?.component?.version?
            `(version: ${moduleTemplate?.spec?.descriptor?.component?.version})` :
            '';
        return channelTest.push({
          text: moduleTemplate.spec.channel,
          additionalText,
        });
    });
```

src/components/Extensibility/components-form/Modules/Modules.js
```diff
@@ -0,0 +1,226 @@
...
+        moduleTemplate?.metadata?.labels['operator.kyma-project.io/beta'] ===
+          'true',
+    );
+
+    return (
+      <>
+        <div className="flexbox fd-margin-bottom--sm">
+          <div className="fd-color--neutral">
+            {index === 0 ? `${sectionName}:` : ''}
+          </div>
```
Akatuoro, line 139: sectionName is already in the collapsible section part below

src/components/Extensibility/components-form/Modules/Modules.js
```diff
@@ -0,0 +1,226 @@
...
+                      return module.name === name;
+                    })
+                  : index,
+              );
+            }}
+          >
+            {name}
+          </Checkbox>
+          <div>
+            <p className="fd-color--neutral">Module channel overwrite:</p>
```
Akatuoro, line 161: translation missing

src/components/Extensibility/components-form/Modules/Modules.js
```diff
@@ -0,0 +1,226 @@
...
+            <ComboboxInput
+              compact
+              disabled={!isChecked}
+              placeholder={'Uses Default Kyma Channel'}
```
Akatuoro, line 165: translation missing

src/components/Extensibility/components-form/Modules/Modules.js
```diff
@@ -0,0 +1,226 @@
...
+          </div>
+
+          {!parsedOptions?.displayDocs && link ? (
+            <Link
+              href={link}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="fd-align-vertical-center"
+            >
+              DOCUMENTATION
```
Akatuoro, line 204: translation missing

mrCherry97 (dismissed):

mrCherry97 (approved):

mrCherry97 (approved):

nataliasitko (approved):

PR #2435, opened 2023-04-13
Customizable security validation

Akatuoro:
**Description**

Changes proposed in this pull request:

- Enable providing rules for resource validation
  - The default rules are loaded via `/validation-resources/rule-set.yaml` after being bundled by webpack
  - Rule sets can be added as `ConfigMap` with the label `busola.io/resource-validation=rule-set`
- Enable policy selection - a policy is a collection of rules that should be applied
  - On cluster level via feature flag
  - Locally via the user preferences
- Update rules from datree and add a script to update them
- Add rules and policies for Kubernetes Baseline and Restricted Pod Security Standards

**Related issue(s)**

This PR partly implements #2376.

PR #2531, opened 2023-07-21
Cluster Validation Scan

### LINKED ISSUE #2376: Security and compliance validation
**Description**

Add tooling to visualize insecure or non-compliant configurations and workloads. The validation needs to be possible for the whole cluster and can also be shown as warning when creating resources. Since security and compliance needs differ for applications, the rules need to be configurable.

**Reasons**

Keeping the Kyma runtime cluster secure and compliant is an important but sometimes not easy activity. Even if Kyma is shipping the cluster secure by default, the responsibility to keep it secure is on the user side. On application side, there may also be non-compliant or insecure configuration changes, e.g. when new features are introduced.
Validation tooling can help visualize and give a report on whether the cluster is secure and whether it complies with specific standards.

**Tasks**

- [ ] This is a core functionality that needs to have in mind performance and try not to choke the API server. 
  - [ ] We evaluate the use of javascript workers, to allow background work
  - [ ] We evaluate the use of a queue, that allows the control of requests. The concurrency will be configurable allowing for at least 2 options: aggressive/active scan or background scan
  - [ ] It's acceptable that between scans the resource can change and we don't capture that. There is no need for a watch the resource approach
  - [ ] Evaluate the possibility of using an in-memory DB that can be used as a cache and also as a source for the workers.
  - [ ] Various workers ( that can run different rulesets ) can run at different intervals.

Akatuoro:
**Description**

- Add a cluster validation section in the cluster overview page
  - It is hidden behind a `CLUSTER_VALIDATION` feature flag
- The section enables a scan of the cluster
  - The scan validates the cluster resources against the rules of #2435
  - The user can configure which namespaces to scan and which policies to check
    - System namespaces like kube-system and kyma-system are excluded by default.
    - By default, the policies chosen by the user and cluster for resource validation are used
  - The user can configure the number of parallel requests to the api, 5 by default
- The scan result is shown as a tree where the user can drill down to the specific warnings
- This feature uses ui5 react webcomponents, so it includes the ui5 react webcomponent libraries and sets the necessary ThemeProvider
- The pqueue library is included for queuing requests and work for the worker thread

**Related issue(s)**

Resolves #2376


mrCherry97 (changes requested):
Scan Progress shouldn't be displayed if a user is not scanning cluster.

When I provide `Parallel Requests` <= 0 then UI breaks.

We have a talk to add button for choose all namespaces and uncheck all.

It will be nice to have some margin/padding here:
image

Info about feature flag should be added in docs and you can add it to defaultConfig with enabled false for now.

Why you are implementing TypeScript and JavaScript mixed? `ClusterValidation.js` rest of new files are `ts`.

Akatuoro:
> * [x]  Scan Progress shouldn't be displayed if a user is not scanning cluster.

Adjusted so that both Scan Progress and Scan Result is not shown if not available.

> * [x]  When I provide `Parallel Requests` <= 0 then UI breaks.

Fixed

> * [ ]  We have a talk to add button for choose all namespaces and uncheck all.

will work on that tomorrow

> * [x]  It will be nice to have some margin/padding here:
>   image

done

> * [x]  Info about feature flag should be added in docs and you can add it to defaultConfig with enabled false for now.

done

> * [x]  Why you are implementing TypeScript and JavaScript mixed? `ClusterValidation.js` rest of new files are `ts`.

Good point, I've changed `ClusterValidation.js` to `tsx` and `ResourceValidation.worker.js` to `ts`.
I had it in `js` for faster iteration with components that are not well typed. Also needed to update the fundamental-react types, but now it's working well.

There is still 1 file in `js`:
`ClusterValidationConfiguration.js` - this is explicitly in `js` because the `ResourceForm` does not have proper types. So it is very verbose to assign all `ResourceForm` properties in `ts`.

mrCherry97 (changes requested):
When I am a user with reduced permissions only for one namespace I can't list namespaces, but ClusterValidation feature is displaying all of them.
image

image

User should be able to see only this namespace where he has access.

mrCherry97:
And also parallel <0 has no sense

Akatuoro:
> When I am a user with reduced permissions only for one namespace I can't list namespaces, but ClusterValidation feature is displaying all of them. image
> 
> User should be able to see only this namespace where he has access.

Explanation: The restricted user has no access to /api/v1/namespaces , so the response status code is 403. This is counted as an error, so it goes into [processError](https://github.com/kyma-project/busola/blob/main/src/shared/hooks/BackendAPI/useGet.js#L77-L82) instead of [processDataFn](https://github.com/kyma-project/busola/blob/main/src/shared/hooks/BackendAPI/useGet.js#L74-L76)

I've included a fix for `useAvailableNamespaces`.

Akatuoro:
> And also parallel <0 has no sense

I've added a resource form validation for this. It is still possible to submit with <= 0, in which case the scan uses 1 connection.
Screenshot 2023-07-28 at 15 53 55

mrCherry97 (changes requested):
We need to have some tests for that functionality.

Akatuoro:
> We need to have some tests for that functionality.

I've added tests for the scan

src/components/Clusters/views/ClusterOverview/ClusterValidation/ClusterValidation.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+      {!scanReady && (
+        <LayoutPanel.Filters className="fd-has-padding-none">
+          <Loader type="Indeterminate" />
+        </LayoutPanel.Filters>
+      )}
+
+      <LayoutPanel.Body>
+        <Section titleText={t('cluster-validation.scan.scope')}>
+          <FlexBox>
```
akucharska, line 181: Consider adding padding to the flexbox
```suggestion
                    <FlexBox className="fd-has-padding-small">
```

akucharska (commented):
It looks really good.
My only suggestion is adding some additional styling for better looks

mrCherry97 (approved):

docs/resource-validation/README.md
```diff
@@ -55,3 +55,9 @@ policies:
 Policies can also reference rules in different rule sets as long as they are loaded in Busola.
 
 To see the full specifications, check the [example rule-set](../../examples/resource-validation/rule-set.yaml).
+
+## Scan the cluster
+
+With the `CLUSTER_VALIDATION` [feature flag](../features.md) enabled, you can use these rules to scan existing resources in your cluster.
```
IwonaLanger, line 61: I'd use the same formatting for CLUSTER_VALIDATION as in `features.md`, that is, bold font. It's not a flag, right?

IwonaLanger (approved):

PR #2546, opened 2023-08-11
Download button

Akatuoro:
**Description**

- Add download button in user preferences for downloading policies
  - Enabled policies and rules are downloaded in [datree's Custom Rules](https://hub.datree.io/custom-rules/custom-rules-overview) format, so that they can either be re-uploaded in busola on another cluster or be used for validation using the datree cli
- Add support for [datree's Custom Rules](https://hub.datree.io/custom-rules/custom-rules-overview) format when loading rule sets
- Fix duplicate enum values in Pod Security Standard rules

**Related issue(s)**

Enhancement for #2376


docs/resource-validation/README.md
```diff
@@ -56,6 +56,43 @@ Policies can also reference rules in different rule sets as long as they are loa
 
 To see the full specifications, check the [example rule-set](../../examples/resource-validation/rule-set.yaml).
 
+### Alternative format: Custom Rules
+
+Rules and policies can also be given using [datree's Custom Rules](https://hub.datree.io/custom-rules/custom-rules-overview) format as part of a rule set.
```
IwonaLanger, line 61: The link doesn't work (returns 404).

Akatuoro: @IwonaLanger, thanks for checking this. I also did a check last week and can confirm that the page was indeed a 404 page temporarily. Now it is up again...

As it is an external link, this is not under our control. Should we keep it like it is? Alternatively, we could remove the link completely or only link to the entry page hub.datree.io. But I think a deep link would be more helpful.

IwonaLanger: @Akatuoro Let's keep the link. If it was just a temporary failure, there's no need to remove it.

IwonaLanger (approved):

PR #2560, opened 2023-08-17
UI5-breadcrumbs

chriskari:
**Description**

- replaced breadcrumbs

### DIFF
--- a/src/shared/components/PageHeader/PageHeader.js
+++ b/src/shared/components/PageHeader/PageHeader.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { LayoutPanel, Breadcrumb } from 'fundamental-react';
-import { Link } from 'react-router-dom';
+import { Breadcrumbs, BreadcrumbsItem } from '@ui5/webcomponents-react';
+import { LayoutPanel } from 'fundamental-react';
 
 import './PageHeader.scss';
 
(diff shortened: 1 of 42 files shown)


tests/integration/tests/cluster/test-command-palette.spec.js
```diff
@@ -70,7 +70,10 @@ context('Test Command Palette navigation', () => {
       .first()
       .click();
 
-    cy.contains('Cluster Details - Nodes').should('be.visible');
+    //cy.contains('Cluster Details - Nodes').should('be.visible');
```
mrCherry97, line 73: why it's commented?

tests/integration/support/navigate-back-to.js
```diff
@@ -0,0 +1,12 @@
+Cypress.Commands.add('navigateBackTo', (resourceUrl, resourceName) => {
+  cy.get('ui5-breadcrumbs')
+    .find(`ui5-link[href*=${resourceUrl}]`, {
+      includeShadowDom: true,
```
mrCherry97, line 4: try to add `includeShadowDom: true` into `cypress.config.js` I made that in ShellBar PR and it works, so probably all of this includes could be potentially removed.

src/shared/components/PageHeader/PageHeader.js
```diff
@@ -31,15 +31,19 @@ export const PageHeader = ({
...
+            <Breadcrumbs design="NoCurrentPage">
               {breadcrumbItems.map(item => {
                 return (
-                  <Breadcrumb.Item aria-label="breadcrumb-item" key={item.name}>
-                    <Link to={item.url}>{item.name}</Link>
-                  </Breadcrumb.Item>
+                  <BreadcrumbsItem
+                    aria-label="breadcrumb-item"
+                    key={item.name}
+                    href={item.url}
```
mrCherry97, line 40: coulnd we use `Link` as previously or some onclick with `navigate` function? because currently clicking on breadcrumb is reloading whole page and this has bad feeling in my point of view, if not then we need to stay with this, but please try to make it

mrCherry97 (changes requested):

mrCherry97 (approved):

mrCherry97 (approved):

PR #2564, opened 2023-08-24
ui5-wizards

mrCherry97:
**Description**

- migrate add cluster wizard to UI5,
- migrate extensibility wizard to UI5,
- add WizardButtons component to be reused in every WizardStep.

**Related issue(s)**


src/shared/components/WizardButtons/WizardButtons.js
```diff
@@ -0,0 +1,53 @@
...
+            ? customFinish
+            : t('common.buttons.submit')
+          : t('clusters.buttons.next-step')}
+      </Button>
+      <Button design="Transparent" onClick={onCancel} aria-label="cancel">
+        {t('common.buttons.cancel')}
+      </Button>
+    </div>
+  );
+}
```
chriskari, line 53: maybe add some css so that the buttons are positioned like here:
https://kyma-project.github.io/price-calculator/

(previous step on the left, next step on the right)

akucharska (commented):
What is this white shadow over the buttons:
Screenshot 2023-08-30 at 08 43 17

Can we do sth with this space bellow the buttons?
Screenshot 2023-08-30 at 08 43 24

src/shared/components/WizardButtons/WizardButtons.js
mrCherry97: we will revisit CSSes after UX changes anyway.

PR #2571, opened 2023-09-12
Ui5 tabs

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


mrCherry97 (changes requested):
there is a problem with height and overflow (maybe it needs to be set as hidden) for example check functions view:
image
also we need to get rid of this background below Tab:
image

mrCherry97 (approved):

PR #2573, opened 2023-09-13
Ui5 icons

mrCherry97:
**Description**

- move Icons to UI5,
- add CSSes for Icon size (values taken from fundamental)

**Related issue(s)**


akucharska (commented):
We still use something like:
```
import { IconGlyph } from 'fundamental-react/lib/Icon/Icon';

export type Category = {
  key: PredefinedCategories;
  icon: IconGlyph;
  label: string;
  items: NavNode[];
  topLevelNode?: boolean;
};
```
in a couple of places. From what I can see webcomponents don't export similar type with available Icon names so we probably need to change it to string

akucharska (approved):

PR #2585, opened 2023-09-21
Ui5 checkbox

chriskari:
**Description**

- replaced checkboxes with ui5

**Related issue(s)**


mrCherry97 (changes requested):
if it's possible please wait to merge this after ui5-dialogs, because there is a giant refactor and there will be a lot of conflicts, so it will be better to have this PR later

mrCherry97 (changes requested):
check if this is neccesary in `index.scss`
```
// override fd-styles for Safari
input[type='checkbox'] {
  -webkit-appearance: checkbox !important;
}
```

mrCherry97 (approved):

PR #2614, opened 2023-10-19
move infoLabel and menus

mrCherry97:
**Description**

- move infoLabel and menus to ui5,
- use combo box instead of menu for search,
- use a button with a menu instead of a popover,
- fix for Modules wizard.

**Related issue(s)**


src/components/CustomResources/CustomResourcesOfType.js
```diff
@@ -58,12 +58,6 @@ export default function CustomResourcesOfType({ crdName }) {
           </Link>
         </DynamicPageComponent.Column>
       </DynamicPageComponent>
-      <CustomResources
```
akucharska, line 61: Why this is removed?

mrCherry97: because I forgot about it with `DynamicPageHeader` refactor and with this CustomResources are rendered two times

akucharska (approved):

PR #2628, opened 2023-10-27
Fix custom form-validation

chriskari:
**Description**

- ...
- ...
- ...

**Related issue(s)**


mrCherry97 (changes requested):
I shouldn't be able to click create when `required` value is not filled in
image

Works for extensibility, but it's not working for normal views.

PR #2639, opened 2023-11-07
Fix forms

mrCherry97:
**Description**

- ad functionality for fd-col from fundamental,
- rename fd-col to bsl-col
- remove unused styles

**Related issue(s)**

### DIFF
--- a/src/shared/ResourceForm/fields/MultiInput.js
+++ b/src/shared/ResourceForm/fields/MultiInput.js
@@ -100,8 +100,8 @@ export function MultiInput({
 
   const listClasses = classnames({
     'text-array-input__list': true,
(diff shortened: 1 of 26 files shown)


src/shared/ResourceForm/fields/MultiInput.js
```diff
@@ -179,7 +179,7 @@ export function MultiInput({
           {internalValue.map((entry, index) => (
             <li key={index} className="text-array-entry">
               <FlexBox alignItems="Baseline" style={{ gap: '10px' }}>
```
akucharska, line 181:
```suggestion
              <FlexBox alignItems="Baseline">
```

akucharska (approved):

PR #2712, opened 2023-12-28
Command Palette redesign

OliwiaGowor:
**Description**

- adjusted translations, corresponding to the new design
- changed Command Palette look
- changed search behavior
- adjusted e2e tests

**Related issue(s)**


public/i18n/en.yaml
```diff
@@ -175,10 +177,10 @@ command-palette:
     open-upload-yaml: Open Upload YAML
     open-preferences: Open Preferences
     resource-aliases: Built-in Resource Aliases
-    short-help: Type to search, e.g. "ns default" or "pod nginx"
+    short-help: Type <0>?</0> for help
     all-namespaces: List resources from all Namespaces
   item-actions:
-    navigate: Navigate
+    navigate: <0>Enter</0> to navigate <0>Tab</0> to autocomplete
```
NHingerl, line 183:
```suggestion
    navigate: Use <0>Enter</0> to navigate, <0>Tab</0> to autocomplete
```

public/i18n/en.yaml
```diff
@@ -175,10 +177,10 @@ command-palette:
     open-upload-yaml: Open Upload YAML
     open-preferences: Open Preferences
     resource-aliases: Built-in Resource Aliases
-    short-help: Type to search, e.g. "ns default" or "pod nginx"
+    short-help: Type <0>?</0> for help
     all-namespaces: List resources from all Namespaces
```
NHingerl, line 181: Why is line 180 in sentence case but line 178, for example, in title case? 
It might be correct, depending on where it's used in the UI (compare the [guidelines](https://help.sap.com/docs/SNG_FOR_UA/e33c591ae4494a659a3f5f983c9d1161/79ce282052d14cc38e6a203a1fc277c4.html?state=DRAFT#english%3A-capitalization-on-the-user-interface)), but I'm not sure.

Also, we [don't capitalize "namespace"](https://github.com/kyma-project/community/blob/main/docs/guidelines/content-guidelines/02-style-and-terminology.md#camelcase). (unless, ofc, if the whole string is in title case). 

```suggestion
    all-namespaces: List resources from all namespaces
```

or 

```suggestion
    all-namespaces: List Resources From All Namespaces
```

NHingerl (approved):

NHingerl (dismissed):

chriskari (approved):

mrCherry97 (changes requested):
There is one "additional empty list item" after namespaces, it shouldn't be there
image

Also there is an information on second option that I can click tab to autocomplete thich causes misunderstanding, because as a user, I might to think that I can navigate more, after second tab click there is `namespaces/default/` and I can write more, but nothing changes and after clicking tab one more time I see only `namespaces/`
step by step screenshots:
image

image

image

NHingerl (approved):

mrCherry97 (approved):

PR #2756, opened 2024-01-31
feat: adjust add cluster wizard to new design

OliwiaGowor:
**Description**

- changed Add Cluster Wizard design
- added form input in Add Cluster Wizard
- added Review step in Wizard
- adjusted tests

**Related issue(s)**


mrCherry97 (changes requested):
I found out some bug and problems, overall it looks very good, and better then it was previously, but it needs some "polish", here is a list of style problems or some bugs:
- [ ] YAML Editor should be bigger,
- [ ] If nothing is provided to YAML Editor, there should be a placeholder `Paste .kubeconfig file`
- [ ] Editing `Name` in cluster is editing only `current-context` field, and this provides errors, it should also edit a `contexts.[].name`, please check how busola react on multiple context kubeconfig
- [ ] YAML Dialog should have `Reset` button and there shouldn't be search Icon
image
- [ ] Text inside dropdown should be changed to `Drop a .kubeconfig file or click to upload` (potential effect: YAML Upload inside the cluster is using component for drag and drop, check if the name will change there also)
- [ ] BUG: when I made a mistake and paste something, remove and I want to pick a file from my computer, file is provided, but YAML is not showing anything (when I click Form and then YAML, ten kubeconfig is shown properly):
image
- [x] button text should be changed `UI Form` -> `Form`
- [ ] when I have multiple context kubeconfig align of the input is weird:
image
- [ ] Escape button is not working (it should close the wizard, check if UI5 didn't mess something, previously it worked)
- [ ] Authetication token and OIDC needs to be full-width (check figma)
image
image
- [ ] When I provide authentication token using input from previous screenshot it's not shown on the final step
image
and connected to this problem busola is reacting wierd, because token is not provided, but I'm "connected" to the cluster
image
after editing cluster using form, everything is working properly, so potentially it's a problem with `Authentication Step`

If you will have any questions or problems to reproduce some bug, please catch me up, I'll try to help you.

OliwiaGowor:
The form view look will be revisited after [in page edit](https://github.com/kyma-project/busola/pull/2757) is done because it will introduce new form design for all of them

mrCherry97 (changes requested):
- [x] YAML Editor is too big, I can't see warning or parse errors, I need to scroll down to check it
image
image

- [x] token is provided, so this warning that authentication is not provided shouldn't be displayed probably, but please double-check that
image

mrCherry97 (approved):

PR #2761, opened 2024-02-12
feat: implementation of Redesigned Status Card

OliwiaGowor:
**Description**

- created ResourceStatusCard
- added Status Card to Deployments

**Related issue(s)**


src/components/Clusters/views/ClusterOverview/ClusterOverview.js
```diff
@@ -64,6 +65,15 @@ export function ClusterOverview() {
...
             />
+            <Title
+              level="H3"
+              style={{
+                ...spacing.sapUiMediumMarginBegin,
+                ...spacing.sapUiMediumMarginTopBottom,
+              }}
+            >
+              {t('cluster-overview.headers.cluster-details')}
+            </Title>
```
mrCherry97, line 76: Why is it moved from `title` prop?

mrCherry97 (changes requested):
- [ ] opening one of the collapsed section is closing second one, user needs to be able to open two of them
image

- [x] status `False` shouldn't be green
image

- [ ] if available replicas is not existing in status, then we need to display unavailableReplicas or `0`, question to PO or UXes
```status:
  observedGeneration: 1
  replicas: 1
  updatedReplicas: 1
  unavailableReplicas: 1
  conditions:
    - type: Available
      status: 'False'
```
image

- [x] in Figma design there is `Last scale time` displayed
YAML example:
```
      lastUpdateTime: '2024-02-19T13:31:44Z'
      lastTransitionTime: '2024-02-19T13:31:44Z'
```

- [ ] also in figma there is more information about this deployment replicas
image

src/components/Clusters/views/ClusterOverview/ClusterOverview.js
OliwiaGowor: the title had to be moved outside of ResourceDetailsCard component in order to position it parallel to ResourceStatusCard, and because ClusterOverview isn't created with ResourceDetails component the title had to be put there separately

resources/web/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
         - name: busola
-          image: europe-docker.pkg.dev/kyma-project/prod/busola:latest
+          image: europe-docker.pkg.dev/kyma-project/dev/busola-web:PR-2761
```
mrCherry97, line 19:
```suggestion
          image: europe-docker.pkg.dev/kyma-project/prod/busola:latest```

mrCherry97 (changes requested):

mrCherry97 (approved):

PR #2803, opened 2024-03-01
fix: view is jumping with column layout on some views

OliwiaGowor:
**Description**

- fixed columns jumping in all Extensibility and Custom Resources views
- adjusted tests

**Related issue(s)**


mrCherry97 (changes requested):
after clicking refresh link stays, column is not opened, resource is marked as selected, and I can't open any resource, I need to click on another resource is sidenav and go back, after that I can open a resource ddetails.

mrCherry97 (changes requested):
custom resources -> pizza orders -> diavola -> refresh page
image

mrCherry97 (approved):

PR #2812, opened 2024-03-06
feat: set edit view type based on preferences

OliwiaGowor:
**Description**

- added feature to set preferred resource edit view in Preferences
- in case of "Default" setting, added feature that sets resource edit view dynamically, based on last choice

**Related issue(s)**


akucharska (changes requested):
For some reason when you go to Events detail page and switch to yaml tab the view crashes

akucharska (approved):

PR #2899, opened 2024-05-02
fix: remove redirect in Custom Resources when resource has extension defined

OliwiaGowor:
**Description**

- remove redirect in Custom Resources when resource has extension defined

**Related issue(s)**

### DIFF
--- a/src/components/CustomResources/CustomResources.js
+++ b/src/components/CustomResources/CustomResources.js
@@ -1,4 +1,3 @@
-import React from 'react';
 import * as jp from 'jsonpath';
 import pluralize from 'pluralize';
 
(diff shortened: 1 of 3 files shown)


src/components/CustomResources/CustomResources.js
```diff
@@ -127,7 +93,7 @@ export function CustomResources({
     customColumnLayout,
     layoutNumber: 'MidColumn',
     parentCrdName: crd.metadata.name,
-    handleRedirect,
+    //handleRedirect,
```
akucharska, line 96: Remove comented param

akucharska (changes requested):
When we create resource (eg pizza) from custom resource list we are redirected to pizza details. 
Screenshot 2024-05-06 at 10 44 14

Shouldn't we redirect back to custom resource details?
Screenshot 2024-05-06 at 10 44 31

akucharska (approved):

PR #2903, opened 2024-05-07
feat: brought back slash-shortcut

chriskari:
**Description**

- ...
- ...
- ...

**Related issue(s)**


mrCherry97 (approved):

mrCherry97 (changes requested):
you need to disable slashshortcut when column layout is opened, because create forms are broken and you can't write for example bitnami/nginx  in form

mrCherry97 (approved):

PR #2931, opened 2024-05-21
feat: adjust pods

akucharska:
**Description**

- ...
- ...
- ...

**Related issue(s)**


src/resources/Pods/ContainersData.js
```diff
@@ -23,52 +25,70 @@ export default function ContainersData({ type, containers, statuses }) {
...
-      />
-      {container.image && (
-        <LayoutPanelRow name={t('pods.labels.image')} value={container.image} />
-      )}
-      {container.imagePullPolicy && (
+  const ContainerComponent = ({ container, status }) => {
+    const state =
+      status?.state?.running ||
+      status?.state?.waiting ||
+      status?.state?.terminated;
```
mrCherry97, line 32: is there a possibility, that state is not presented in YAML? if yes it will be nice to have last case set as `UNKNOWN` with type `none`

mrCherry97 (changes requested):
UI brokes after opening this resource: `namespaces/default/pods/hello-world-trigger-nodejs18-build-jb4m2-rl6mm`
image

src/resources/Pods/ContainersData.js
akucharska: This is only used to get `state?.startedAt`. If it's null we return '-'

```
  if (!timestamp) {
    return EMPTY_TEXT_PLACEHOLDER;
  }
```

mrCherry97 (approved):

PR #2994, opened 2024-06-24
fix: Remove Monitoring and Health title from details except for Namespace details

akucharska:
…Details

**Description**

- Remove Monitoring and Health title from details except for Namespace details

**Related issue(s)**


mrCherry97 (changes requested):
rename `showCustomHealthCardsTitle` -> `showHealthCardsTitle` there is nothing custom In my opinion 

in ResourceHealthCrad prop should have the same name as in the rest, not just `showTitle`

mrCherry97 (approved):

PR #3020, opened 2024-06-27
test: e2e extensibility

dbadura:
**Description**

- Add extensibility test
- Fix extensibility create dialog

**Related issue(s)**

### DIFF
--- a/examples/extensions/configuration/namespace.yaml
+++ b/examples/extensions/configuration/namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: potato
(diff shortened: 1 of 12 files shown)


mrCherry97 (changes requested):
Looks very good, but I'll change that to a cluster scoped extension, to not need to create namespace etc. We are testing namespace extensions, but not at cluster scope, so it will be nice usage for this, and also test will be a little bit shorter with less files needed.

mrCherry97 (dismissed):

examples/extensions/configuration/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: potato
```
mrCherry97, line 4: please remove it

mrCherry97 (changes requested):

mrCherry97 (approved):

PR #3026, opened 2024-07-01
feat: allow to change channel for the module

OliwiaGowor:
**Description**

- added option to change channel for the module
- fixed columns in form
**Related issue(s)**


mrCherry97 (changes requested):
I have btp-operator module installed in channel fast, so when I click `add` and `advance` I need to see this version which is installed, currently regular is displayed first
image

and also all the time when I click advance first displayed in dropdown should be the channel which kyma resource is.

mrCherry97 (approved):

mmitoraj (approved):

PR #3053, opened 2024-07-10
fix(oidc): client secret may contain equal signs

v0lkc:
**Description**

This PR addresses an isssue with the current handling of the `client_secret` in busola. It does not support equal signs (`=`) within the value, leading to the failure of the oidc flow. 

**Related issue(s)**

### DIFF
--- a/src/components/Clusters/components/oidc-params.ts
+++ b/src/components/Clusters/components/oidc-params.ts
@@ -12,7 +12,22 @@ export function parseOIDCparams({ exec: commandData }: KubeconfigOIDCAuth) {
   let output: any = {};
 
   commandData.args.forEach(arg => {
-    const regex = new RegExp('^(?<key>[^=\\s]+)(?:=(?<value>.*$))');
+    /**
+     * The regular expression defined below is tailored to parse command line arguments, capturing both arguments that
+     * include values and those without values. The regex facilitates the extraction of both the keys and optionally
+     * associated values in a reliable manner.
+     *
+     * Examples of command line arguments that this regex will match:
+     * - With value: `--param-with-value=abc`
+     * - Without value: `--param-without-value`
+     *
+     * The regular expression uses named groups (`key` for the argument and `value` for its associated value, if any)
+     * to provide clear and convenient access to parts of the matched text.
+     *
+     * For an interactive example that demonstrates this regex pattern with various inputs and explains each part of the
+     * expression in detail, visit the provided link: https://regex101.com/r/3hc8qX/2
+     */
+    const regex = new RegExp('^(?<key>[^=]+)(?:=(?<value>.*$))');
     const match = arg.match(regex);
 
     if (!match) {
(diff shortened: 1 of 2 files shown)


src/components/Clusters/components/oidc-params.ts
```diff
@@ -12,7 +12,16 @@ export function parseOIDCparams({ exec: commandData }: KubeconfigOIDCAuth) {
   let output: any = {};
 
   commandData.args.forEach(arg => {
-    const [argKey, argValue] = arg.split('=');
+    const regex = new RegExp('^(?<key>[^=\\s]+)(?:=(?<value>.*$))');
```
dbadura, line 15: Could we replace the regex with something simpler?

v0lkc: Do you want me to make the regex simpler or do you want me to **not** use regex at all?

v0lkc: Honestly speaking, it is already simplified and leaves out some aspects of the real world you'd normally have to cover. Don't forget there's normally a shell environment in between and thus you could also expect to receive param values like this:

```bash
--param-without-value
--param-with-value=123

# everything below here is not supported yet
--param-with-value value-separated-by-space-instead-of-comma
--param-with-value="123"
--param-with-value='123'
--param-with-value=$(cat ...)
--param-with-value=`cat ...`
```

dbadura: I though about removing the regex at all.

v0lkc: You mean something like this? 

```ts
const parts = arg.split('=');
const argKey = parts[0];
const argValue = parts.slice(1).join('=');
```

From the bottom of my heart, please don't go this way.

v0lkc: @dbadura @mrCherry97 Can we move forward here? For the reasons mentioned above, I can't recommend to go with a different approach than regex.

mrCherry97: @v0lkc I'm not a fan of the regexes like this, because it will be painful to debug in the future then there will be some improvements or bugs. In this case, I don't see any better solution, so please provide some comment which is describing this regex

mrCherry97 (changes requested):

mrCherry97 (approved):

v0lkc:
```bash
       Spec                                              Tests  Passing  Failing  Pending  Skipped  
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✖  z-run-after.spec.js                      00:27        2        1        1        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✖  1 of 25 failed (4%)                      14:10       92       91        1        -        -  
```

Is there anything I need to do to get this check green?

mrCherry97:
sadly it's flaky, cypress sometimes does not detect UI5 components, I have rerun the test, after passing, PR should be merged

PR #3072, opened 2024-07-17
feat: separate workflow for unit tests

mrCherry97:
**Description**

- separate unit tests from build
- remove running unit tests from Dockerfiles

**Related issue(s)**

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests

### DIFF
--- a/.github/workflows/pull-unit-tests.yml
+++ b/.github/workflows/pull-unit-tests.yml
@@ -1,0 +1,49 @@
+name: PR Unit Tests
+
+on: 
+  pull_request:
+    types: [opened, edited, synchronize, reopened, ready_for_review]
+    paths:
+      - ".github/workflows/pull-unit-tests.yml"
+      - "backend/**"
+      - "public/**"
+      - "src/**"
+      - "tests/**"
+      - "package.json"
+      - "Makefile"
+      - "Dockerfile.local"
+
+jobs:
+  run-namespace-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: gardenlinux/workflow-telemetry-action@v2
+        with:
+          comment_on_pr: false
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-unit-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - name: setup_busola
+        shell: bash
+        run: |
+          set -e
+          npm ci
```
dbadura, line 33: We can use alias: `clean-install` to be easier to understand what this step is doing.

dbadura (changes requested):

.github/workflows/pull-unit-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+    types: [opened, edited, synchronize, reopened, ready_for_review]
+    paths:
+      - ".github/workflows/pull-unit-tests.yml"
+      - "backend/**"
+      - "public/**"
+      - "src/**"
+      - "tests/**"
+      - "package.json"
+      - "Makefile"
+      - "Dockerfile.local"
```
dbadura, line 14: Do we need to run test on Dockerfile changes?

.github/workflows/pull-unit-tests.yml
mrCherry97: we are using `npm ci` everywhere and it's clear, so I'll go with that to keep the convention on which we agreed, if we want to change that it should be changed in all of the workflows.

dbadura: We can extrend `ci` to `clean-install`, looks ok

OliwiaGowor (approved):

Issue #3055, opened 2024-07-10
ClusterPreview shows only one `scope`, even if there are multiple configured in the kubeconfig


v0lkc:
**Description**

The review page of the cluster wizard only displays one oidc scope, even if multiple scopes are configured in the kubeconfig. The configuration may look like this:

```
users:
- name: xsuaa
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - oidc-login
      - get-token
      - --oidc-use-access-token
      - --oidc-issuer-url=...
      - --oidc-client-id=...
      - --oidc-client-secret=...
      - --oidc-extra-scope=scope_a
      - --oidc-extra-scope=scope_b
      - --oidc-extra-scope=scope_c
```

It would only display `scope_a`

**Expected result**

`scope_a`, `scope_b` and `scope_c` should be displayed.

**Actual result**

Only `scope_a` is shown.

**Steps to reproduce**

**Troubleshooting**

PR #3209, opened 2024-08-20
fix: cluster-wizard UI & oidc-scopes

chriskari:
**Description**

- fixed UI issues of cluster wizard
- adjusted forms in cluster wizard to match UX specification
- adjusted code to handle multiple scopes specified in the kubeconfig
- changed scope input field to scopes list input
- multiple scopes can be displayed in review step
- adjusted unit test for oidc parsing

**Related issue(s)**
#3055 #3288

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests

### DIFF
--- a/src/components/Clusters/components/AddClusterWizard.js
+++ b/src/components/Clusters/components/AddClusterWizard.js
@@ -49,11 +49,13 @@ export function AddClusterWizard({ kubeconfig, setKubeconfig, config }) {
   const setIsFormOpen = useSetRecoilState(isFormOpenState);
 
   useEffect(() => {
-    const contentContainer = document
-      .getElementsByTagName('ui5-wizard')[0]
-      ?.shadowRoot?.querySelectorAll('.ui5-wiz-content-item-wrapper')[
-      selected - 1
-    ];
+    const wizard = document.getElementsByTagName('ui5-wizard')[0];
+    const wizardContent = wizard?.shadowRoot?.querySelector('.ui5-wiz-content');
+
+    const contentContainer = wizardContent?.querySelectorAll(
+      '.ui5-wiz-content-item-wrapper',
+    )[selected - 1];
+
     if (contentContainer) {
       contentContainer.style['background-color'] = 'transparent';
       contentContainer.style['padding'] = '0';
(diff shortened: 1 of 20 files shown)


src/components/Clusters/components/AddClusterWizard.js
```diff
@@ -95,6 +95,11 @@ export function AddClusterWizard({ kubeconfig, setKubeconfig, config }) {
 
   const onComplete = () => {
     try {
+      kubeconfig?.users.forEach(user => {
+        if (!user?.user?.exec?.args?.includes('--oidc-extra-scope=openid')) {
+          user?.user?.exec?.args?.push('--oidc-extra-scope=openid');
+        }
+      });
```
v0lkc, line 102: Is there a specific reason why you don't want to go this way?

https://github.com/kyma-project/busola/pull/3289/files

Busola is trying to mimic the behavior of the `kubelogin` command, which does not require `openid` to be passed as `--oidc-extra-scope`. Imho you should implement it the same way.

chriskari: Hi Christian, thanks for your review and contribution.
Yes you are correct, I was unaware of your PR, next time you could attach it to the issue to make it more apparent :D
I have a follow-up question to you: should all the defined `--oidc-extra-scope` be passed along with the openid scope in the `scope` argument of the `UserManager` constructor?

chriskari: meaning if you have defined `--oidc-extra-scope=scope_a`, `--oidc-extra-scope=scope_a` should the scope passed to the oidc UserManager be `scope: "openid scope_a scope_a"`?

v0lkc: Ok, I thought it would be enough to link the fix to the issue. Next time I'll put it right in the beginning of the issue description. You're right, the scopes have to be passed like this.

chriskari: thanks, I will incorporate your changes in my PR 👍

akucharska (changes requested):
1. Cancel button is in strange place
2. Can we get rid of double scroll?

Screenshot 2024-08-26 at 15 54 01

chriskari:
I actually adjusted all of it, this is what it looks like for me locally:
image

Issue #3456, opened 2024-11-07
[Spike] Load dynamically custom web component from URL hosted in the kubernetes cluster


pbochynski:
**Description**
With [this issue](https://github.com/kyma-project/busola/issues/3122) it was proven that the custom module UI can be rendered in the iframe. The goal of this spike is to load custom web component from the resources loaded from cluster. 
Acceptance criteria:
- [ ] create custom web component that is loaded dynamically
- [ ] use ui5 components inside the custom component to verify if we can reuse components bundled with busola

mrCherry97:
PoC is ready in connected PR, follow up task is created.

PR #3471, opened 2024-11-12
Custom web component based extension

pbochynski:
**Description**

- custom web component loaded as busola extension
- 

**Related issue(s)**
- #3456


mrCherry97 (commented):
In my opinion looks good, there is some console.logs which I don't know if they are necessary, and `<div></div> -> <div />` but it's more PoC/MVP, so it's fine, I'll consider changing
`extCustomComponentsEnabled: boolean | undefined, -> extCustomComponentsEnabled: boolean,` and calling `useFeature()` should set `isExtensibilityCustomComponentsEnabled` on false if it's undefined, but we should also change it for every feature flags.

dbadura:
Looks good for me

OliwiaGowor:
Looks good to me

akucharska (commented):
LGTM

PR #3535, opened 2024-12-13
feat: Snow Falling Animation

### LINKED ISSUE #3517: Snow Falling Animation
**Description**
We want to add a visual effect to Busola where snowflakes fall gently across the screen. This will create a festive atmosphere and enhance user experience during the winter.

**Acceptance Criteria**
- [ ] **Snowfall Effect Implementation**: A snowfall effect should be visible on the screen, with snowflakes falling gently from top to bottom.
- [ ] **Responsiveness**: The snowfall effect should be responsive and work well on various screen sizes and devices.
- [ ] **Performance**: Ensure the snowfall animation does not significantly impact the application's performance.
- [ ] **Configuration**: Provide options to easily enable or disable the snowfall effect via application settings or configuration.
- [ ] **Testing**: The snowfall effect should be tested thoroughly across different browsers and devices to ensure consistent behavior.
- [ ] **User Control**: Add a toggle switch or button in the application settings that allows users to turn the snowfall effect on or off.
- [ ] **Documentation**: Document the implementation process and how to configure the snowfall effect within the application.

### Additional Notes
- Consider using libraries like `react-snowfall` or similar for ease of implementation.
- Make sure the animation is subtle and not intrusive to the user's interaction with the app.

**Attachments**

KonradPietocha:
**Description**

- added react-snowfall library
- added button to turn the animation off and on
- using existing feature flag SNOW to enable this feature
- changed on using const instead of string for feature type

**Related issue(s)**
https://github.com/kyma-project/busola/issues/3517
resolves #3517 

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests


src/header/Header.tsx
```diff
@@ -32,16 +33,19 @@ import './Header.scss';
 import { isResourceEditedState } from 'state/resourceEditedAtom';
 import { isFormOpenState } from 'state/formOpenAtom';
 import { handleActionIfFormOpen } from 'shared/components/UnsavedMessageBox/helpers';
+import { configFeaturesNames } from 'state/types';
 
 export function Header() {
   useAvailableNamespaces();
   const [isMenuOpen, setIsMenuOpen] = useState(false);
+  const [isSnowOpen, setIsSnowOpen] = useState(false);
```
mrCherry97, line 41: this should be stored in localStorage, because if we have a user who doesn't like this feature, and will click on top to disable snow, then after refresh, he needs to click the snow icon one more time if he doesn't want to see this, we need to have this stored somewhere in local memory.

src/header/Header.tsx
```diff
@@ -168,6 +179,14 @@ export function Header() {
         }
         onProfileClick={() => setIsMenuOpen(true)}
       >
+        {isSnowEnabled && (
+          <ShellBarItem
+            onClick={() => setIsSnowOpen(!isSnowOpen)}
+            icon="heating-cooling"
+            text={'Snow'}
+            title={'Snow'}
```
mrCherry97, line 187: this should be taken from translations (check `useTranslation` usage)

mrCherry97 (changes requested):
In my opinion we need to decrease number of the snows and change the color on the light theme to be more clear to see. (or we can change image for snow to something like it is on https://minutes.tools.sap/#/ )

Also when I'm enabling snow using button, everything looks good, and snow is falling from top, when I'm connecting the cluster or clicking on something which is causing rerender, when I have a bomb from left top corner.

https://github.com/user-attachments/assets/5f604272-b4ff-430c-a6af-dfdbcb0dba1d

mrCherry97 (changes requested):
featureflag on STAGE should be also set on enable, and on prod disabled

mrCherry97 (changes requested):
On first render SNOW is disabled, it should be enabled and set in localStorage to enable

src/header/Header.tsx
```diff
@@ -77,6 +98,16 @@ export function Header() {
     return spaces;
   };
 
+  const handleSnowButtonClick = () => {
+    if (isSnowOpen) {
+      setIsSnowOpen(false);
+      localStorage.setItem(SNOW_STORAGE_KEY, JSON.stringify(false));
```
mrCherry97, line 104: Can't it be set as boolean?

mrCherry97 (changes requested):

src/header/Header.tsx
KonradPietocha: localStorage/sessionStorage only likes string variables: "Storage only supports storing and retrieving strings. If you want to save other data types, you have to convert them to strings. For plain objects and arrays, you can use [JSON.stringify()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)." https://developer.mozilla.org/en-US/docs/Web/API/Storage/setItem

mrCherry97 (approved):

PR #3589, opened 2025-01-14
fix: Inconsistent handling browser history routes

### LINKED ISSUE #3510: Inconsistent handling of browser history and routes
**Description**
Navigation in busola is not deterministic:
1. Path-based routing shows error messages before the correct view is displayed. You can reproduce it on any view that is registered by extension. Example: go to Service Management -> Service Instances. Now go to the address bar of the browser and hit enter (return). The page is reloaded but before the correct view is displayed a user can see the message:
   Screenshot 2024-11-29 at 15 08 14
2. When the user navigates between views the browser history gets 2 entries for each page (one with Busola and other with name of the view:

Screenshot 2024-11-29 at 15 17 49

The history should contain only Secrets and Config Maps entries. If I want to go back to the previous view I need to press back button twice.

3. History navigation doesn't work at all with navigation nodes introduced by extensions. If you do the same exercise as before with switching between Service Instances and Service Bindings, the browser back button has no effect and the displayed content. The active navigation node and the URL in the address bar will change, but you also need to hit back button twice to go back to the previous url.

**Expected result**
1. I can open link to the custom view in busola without seeing incorrect path message.
2. I can navigate to the previous/next view with browser history buttons.

KonradPietocha:
**Description**

- Preventing navigation to the same current path.
- Waiting to handle incorrect paths until extensibility routes are ready.

**Related issue(s)**
Closes #3510 

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests


akucharska (changes requested):
While going back in history, the page does not render correctly. the column is not closed and display _no resource_ message, namespace view is not loaded at all

https://github.com/user-attachments/assets/6af82cc1-52c8-4b7b-b63b-635b647a8851

KonradPietocha:
> While going back in history, the page does not render correctly. the column is not closed and display _no resource_ message, namespace view is not loaded at all
> 
>  Screen.Recording.2025-01-16.at.09.40.02.mov

As I see the same strange behavior is also on main branch. Looks like it's not because of my changes but I will check and try to fix

akucharska (approved):

PR #3590, opened 2025-01-15
fix: resoruces should be displayed correctly.

dbadura:
**Description**

- handle nano cpus which are sent from metrics api
- unclutter resoruces code
- use full cpu unit in data

**Related issue(s)**

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests


src/shared/helpers/resources.js
```diff
@@ -20,36 +20,42 @@ const SI_PREFIXES_BINARY = {
...
   binary = false,
   { unit = '', withoutSpace = true, fixed = 2 } = {},
 ) {
   const prefixMap = binary ? SI_PREFIXES_BINARY : SI_PREFIXES;
   const infix = withoutSpace ? '' : ' ';
 
-  const coreValue = (
-    Math.round((+amount + Number.EPSILON) * 100) / 100
-  ).toFixed(fixed);
+  if (unit !== '' && prefixMap[unit]) {
```
KonradPietocha, line 39: If we also don't want unit to be undefined, false, 0 or null we can just use "if (unit && prefixMap[unit])"

src/shared/helpers/resources.js
```diff
@@ -20,36 +20,42 @@ const SI_PREFIXES_BINARY = {
...
+    const value = (amount / prefixMap[unit]).toFixed(fixed);
+    return {
+      value: value,
+      string: `${value}${infix}${unit}`,
```
KonradPietocha, line 43: Maybe it would be better to change the names to: "value" and "valueWithUnit" or something like that?

dbadura: I think that we can remove the whole object and return just string.

KonradPietocha (approved):

akucharska (approved):

Issue #3603, opened 2025-01-17
Update UI flow to open Kyma Companion


chriskari:
**Description**
The Kyma Companion UI flow needs to be updated to align with the latest [UI/UX design specifications](https://github.com/user-attachments/files/17567819/Kyma.Companion-compressed.pdf). The updated design requires the Companion to be opened via an AI icon in the header bar. Upon opening, users should see:

1. An introductory message.
2. Initial question suggestions fetched from the POST /conversation endpoint.

This approach differs from the previous [POC implementation](https://github.com/kyma-project/busola/pull/3512), where:

1. The assistant could be opened via buttons on individual resource pages.
2. Suggestions were displayed in a popover, and the chat interface was opened only after a user selected a suggestion or submitted a custom prompt

**Tasks**: 

- refactor the previous flow as well as the `CompanionOpener` component to match the updated design

_The necessary adjustments include, but may not be limited to_

- [x] Upon clicking the AI icon fetch prompt suggestions and open the chat interface
- [x] Populate the chat with an introductory message as well as the fetched suggestions
- [x] Remove the CompanionOpener component along with unneeded functionality

PR #3673, opened 2025-02-06
feat: change UI flow to open companion

chriskari:
**Description**

- removed previous implementation (CompanionOpener)
- initial suggestions are fetched on first render of companion
- Chat is populated with introductory message and with fetched suggestions

**Related issue(s)**
#3603

**Definition of done**

- [ ] The PR's title starts with one of the following prefixes:
  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit
- [ ] Explain clearly why you created the PR and what changes it introduces
- [ ] All necessary steps are delivered, for example, tests, documentation, merging


src/components/KymaCompanion/components/KymaCompanion.tsx
```diff
@@ -1,18 +1,52 @@
...
+import { clusterState } from 'state/clusterAtom';
+import getPromptSuggestions from '../api/getPromptSuggestions';
 import './KymaCompanion.scss';
 
 export default function KymaCompanion() {
   const { t } = useTranslation();
   const [showCompanion, setShowCompanion] = useRecoilState<ShowKymaCompanion>(
     showKymaCompanionState,
   );
+  const [suggestions, setSuggestions] = useState<any[]>([]);
```
mrCherry97, line 20: shouldn't be like this?
```suggestion
  const [suggestions, setSuggestions] = useState<string[]>([]);
```

src/components/KymaCompanion/components/KymaCompanion.tsx
```diff
@@ -1,18 +1,52 @@
...
 import { Button, Card, Title } from '@ui5/webcomponents-react';
-import { useRecoilState } from 'recoil';
+import { useRecoilState, useRecoilValue, useSetRecoilState } from 'recoil';
 import {
   ShowKymaCompanion,
   showKymaCompanionState,
 } from 'components/KymaCompanion/state/showKymaCompanionAtom';
 import Chat from './Chat/Chat';
+import { useEffect, useState } from 'react';
+import { sessionIDState } from '../state/sessionIDAtom';
```
mrCherry97, line 10: Could it be like this?
```suggestion
import { sessionIDState } from 'state/sessionIDAtom';
```

src/components/KymaCompanion/components/KymaCompanion.tsx
```diff
@@ -1,18 +1,52 @@
...
+import { clusterState } from 'state/clusterAtom';
+import getPromptSuggestions from '../api/getPromptSuggestions';
```
mrCherry97, line 12: Here also
```suggestion
import getPromptSuggestions from 'api/getPromptSuggestions';
```

src/components/KymaCompanion/components/KymaCompanion.tsx
```diff
@@ -1,18 +1,52 @@
...
+        groupVersion: 'apps/v1', // TODO
+        resourceName: 'test-x', // TODO
+        sessionID, // TODO
+        clusterUrl: cluster?.currentContext.cluster.cluster.server ?? '',
+        token: '', // TODO
+        certificateAuthorityData:
+          cluster?.currentContext.cluster.cluster[
+            'certificate-authority-data'
+          ] ?? '',
+      });
```
mrCherry97, line 40: Could it be moved to a separate folder/file? We have `api` directory.

mrCherry97 (changes requested):

src/components/KymaCompanion/components/KymaCompanion.tsx
chriskari: 👍

src/components/KymaCompanion/components/KymaCompanion.tsx
chriskari: we can move the state file if you want, so instead of having it as a subdirectory of the Companion component
`src/components/KymaCompanion/state/sessionIDAtom.ts` we can move it to the global state directory
`src/state/companion/sessionIDAtom.ts`

src/components/KymaCompanion/components/KymaCompanion.tsx
chriskari: We can outsource it into a custom hook `useGetPromptSuggestions` this will make the code cleaner. But still, it doesn't belong in the api directory because there we only define the requests we make and this part right here simply calls a function from this directory.

src/components/KymaCompanion/components/KymaCompanion.tsx
mrCherry97: I don't have strong opinion, but probably this could make sense, and will be easier to maintain, to have all of the global states in one place.

mrCherry97 (dismissed):

public/i18n/en.yaml
```diff
@@ -759,8 +759,7 @@ kyma-companion:
   error:
     title: Service is interrupted
     subtitle: A temporary interruption occured. Please try again.
-  introduction1: Hello there,
-  introduction2: How can I help you?
+  introduction: Hi, I am your Kyma assistant! You can ask me any question you have and I will try to help you with your problem to the best of my abilities. Meanwhile, you can check suggested questions below, maybe you will find it useful!
```
grego952, line 762:
```suggestion
  introduction: Hi, I am your Kyma assistant! You can ask me any question, and I will try to help you to the best of my abilities. Meanwhile, you can check the suggested questions below; you may find them helpful!
```

grego952 (approved):

mrCherry97 (approved):

Issue #3672, opened 2025-02-06
Incorrect closing of three columns with the back button


KonradPietocha:
**Description**
In Custom Resources, when we open the details to a three-column view and then go back using the back button, the columns close incorrectly and the page content starts to break.

<img width="1708" alt="Image" src="https://github.com/user-attachments/assets/61f840ca-7ea6-4dac-8d24-489a43d70343" />

<img width="1680" alt="Image" src="https://github.com/user-attachments/assets/6d89453e-2e87-4851-aa6c-c50668a5f5bc" />

**Expected result**
The columns close correctly and we see the correct step-by-step views: three columns, two, one with the Custom Resources list, etc.

**Actual result**
Columns not closing properly (second column), leading to a strange view with an unloaded page

**Steps to reproduce**
1. For example, go to Namespace, Custom Resources,
2. Select one of them e.g. functions,
3. Open it, open one of the resource details until you see three columns,
4. Go back with browser back button,
5. The third column has closed,
6. Keep clicking back until we see that there is a problem with closing the second column and further problems.

PR #3753, opened 2025-03-10
fix: Navigation and column layout problems

chriskari:
**Description**

- changed `window.history.pushState` to `navigate` as this is the proper way to handle navigation with react router
- added url parameter showCreate to handle navigation to and from create views (also fixes refreshing with create open)
- outsourced initialLayout code into `usePrepareLayoutColumns` hook
- adjusted layout navigation for normal resources, helmreleases, extensibility resources, modules, extensions, custom resources

**Related issue(s)**
#3672

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


OliwiaGowor (changes requested):
- [ ] refreshing on Modules view with two columns open causes the URL parameter `TwoColumnsMidExpanded` to disappear. As a result at first _Such module doesn't exist_ is visible, then the page refreshes and module appears back

https://github.com/user-attachments/assets/bc7d8344-f2cb-4d12-88d7-0570d87d280a

- [ ] refreshing on any resource with two columns opened causes the URL parameters to disappear and redirects to only resource page

https://github.com/user-attachments/assets/a0f0e951-0c51-4ab4-87ec-bfbea1c15341

- [ ] refresing with Create open also causes the Create colum to close and only list is visible
- [ ] refreshing on Custom Resources with three columns opened, and the third being Create causes error and blocks navigation (probably to state still being set to createOpen: true)
Screenshot 2025-03-19 at 11 48 05

KonradPietocha (approved):

OliwiaGowor (approved):

Issue #3919, opened 2025-05-15
AI Assistant: Namespace information is sometimes passed as empty from dashboard to kyma-companion


mfaizanse:
**Steps to reproduce:**
1. Enable serverless module and wait for it to get ready.
2. Go to default namespace in Kyma Dashboard.
3. Go to workloads > Functions and create a new function with following yaml:
    ```
    apiVersion: serverless.kyma-project.io/v1alpha2
    kind: Function
    metadata:
      labels:
        app.kubernetes.io/name: foo
      name: foo
    spec:
      replicas: 1
      resourceConfiguration:
        function:
          profile: S
      runtime: nodejs20
      source:
        inline:
          dependencies: "hakuna matata"
          source: >-
            module.exports = {
              main: async function (event, context) {
                var queryParams = event['extensions']['request']['query'];
    
                //read query param, for example `?greeting=Hi`
                var greetingMsg = queryParams['greeting']; 
                if(!greetingMsg) {
                  greetingMsg = 'Hello world';
                } 
                const message = greetingMsg
                  + ` from the Kyma Function ${context['function-name']}`
                  + ` running on ${context.runtime}!`;
                console.log(message);
                return message;
              }
            }
    ```
4. Click on the Joule icon to open kyma-companion chat.
5. Note that the initial questions would be not generated. Because the info passed to the kyma-companion contains an empty namespace `Request data: {"resource_kind":"Function","resource_name":"foo","resource_api_version":"serverless.kyma-project.io/v1alpha2","namespace":""}`
6. Ask any question e.g. `What is the problem?`
7. The request sent by kyma Dashboard to companion included a empty namespace.

![Image](https://github.com/user-attachments/assets/fa34b7b4-7566-41ab-ac16-d9104915f7e2)

The namespace should be `default`, not empty.

mfaizanse:
The resource context information is very important for kyma-companion to work properly. In many cases, the kyma-companion is passing incorrect resource information. For example:

```
{"resource_kind":"Tracepipelines.telemetry.kyma-project.io","resource_name":"mybackend1","resource_api_version":"apiextensions.k8s.io/v1","namespace":""}
```

The resource_api_version is incorrect, it should be `telemetry.kyma-project.io/v1alpha1`

PR #3946, opened 2025-06-04
fix: Adjust namespace in columnlayout state after create

chriskari:
**Description**

- fix undefined namespace after creating resource without explicit namespace set

**Related issue(s)**
#3919

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/resources/createResourceRoutes.js
+++ b/src/resources/createResourceRoutes.js
@@ -101,6 +101,7 @@ const ColumnWrapper = ({ list, details, create, ...props }) => {
 
   const listComponent = React.cloneElement(list, {
     ...elementListProps,
+    rawResourceType: props.resourceType,
     layoutCloseCreateUrl,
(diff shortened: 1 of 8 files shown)


src/resources/createResourceRoutes.js
```diff
@@ -101,6 +101,7 @@ const ColumnWrapper = ({ list, details, create, ...props }) => {
 
   const listComponent = React.cloneElement(list, {
     ...elementListProps,
+    rawResourceType: props.resourceType,
```
mrCherry97, line 104: We have a resourceType, the same in `elementListProps`, so we really need the additional property for this?

mrCherry97 (changes requested):

mrCherry97 (approved):

PR #4007, opened 2025-07-15
chore: Update Busola image build workflow and add PR automation for dev deploy

atisap:
**Description**

This PR updates the `busola-build.yml` workflow to match the logic of `busola-web-build.yml`, ensuring that the correct image tag is extracted from the build output and used for automated PR creation to the `busola-deploy` repository.

- Adds a job to extract the build image tag from the output.
- Adds a job to open a PR for dev deploy only when a build tag (e.g., v20...) is present and no semver release tag (e.g., v0., v1., v2.) exists.
- Automates updating the `busola_image` version in `dev/main.tf` and opening a PR with the label `pr-deploy`.

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/5912

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/.github/workflows/busola-web-build.yml
+++ b/.github/workflows/busola-web-build.yml
@@ -59,21 +59,14 @@ jobs:
         run: |
           images='${{ needs.build-web-image.outputs.images }}'
           # Get the first non-latest tag (e.g. v20250715-fcab78c9)
-          tag=$(echo "$images" | jq -r '.[]' | grep -v ':latest' | grep -oE ':[^:]+' | head -n1 | sed 's/^://')
+          tag=$(echo "$images" | jq -r '.[]' | grep -v ':latest' | awk -F: '{print $NF}' | head -n1)
           echo "image_tag=$tag" >> $GITHUB_OUTPUT
 
   open-pr-for-dev-deploy:
     name: Open PR for dev deploy
     runs-on: ubuntu-latest
     needs: [build-web-image, extract-image-tag]
-    if: |
-      contains(needs.build-web-image.outputs.images, 'v') &&
-      (
-        contains(needs.build-web-image.outputs.images, 'v20') &&
-        !contains(needs.build-web-image.outputs.images, 'v0.') &&
-        !contains(needs.build-web-image.outputs.images, 'v1.') &&
-        !contains(needs.build-web-image.outputs.images, 'v2.') # add more semver prefixes as needed
-      )
+    if: contains(needs.build-busola-image.outputs.images, 'v20')
     steps:
       - name: Checkout busola-deploy repo
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/busola-web-build.yml
```diff
@@ -46,3 +46,55 @@ jobs:
...
+          # Get the first non-latest tag (e.g. v20250715-fcab78c9)
+          tag=$(echo "$images" | jq -r '.[]' | grep -v ':latest' | grep -oE ':[^:]+' | head -n1 | sed 's/^://')
+          echo "image_tag=$tag" >> $GITHUB_OUTPUT
+
+  open-pr-for-dev-deploy:
+    name: Open PR for dev deploy
+    runs-on: ubuntu-latest
+    needs: [build-web-image, extract-image-tag]
+    if: |
+      contains(needs.build-web-image.outputs.images, 'v') &&
```
dbadura, line 70: Why is done like that?

.github/workflows/busola-web-build.yml
```diff
@@ -46,3 +46,55 @@ jobs:
...
+    needs: build-web-image
+    outputs:
+      image_tag: ${{ steps.extract_tag.outputs.image_tag }}
+    steps:
+      - name: Extract tag from built images
+        id: extract_tag
+        run: |
+          images='${{ needs.build-web-image.outputs.images }}'
+          # Get the first non-latest tag (e.g. v20250715-fcab78c9)
+          tag=$(echo "$images" | jq -r '.[]' | grep -v ':latest' | grep -oE ':[^:]+' | head -n1 | sed 's/^://')
```
dbadura, line 62: Could you explain the `grep` regex?
Can it be extracted in simpler way?

.github/workflows/busola-web-build.yml
atisap: Because looking for tags like:` v20250715-fcab78c9` and try to avoid release tag like `v0.0.15` and `PR-1234`
So the tag have to contain `v`, usually contain `v20` as well but not contain `v0`-`v2`, but yeah Thank you for pointing out, that can be make it more simple.

.github/workflows/busola-web-build.yml
atisap: `grep -oE ':[^:]+'`: Extracts the part after the last colon (the tag), e.g., :`v20250715-fcab78c9`.
`head -n1`: Takes the first match.
`sed 's/^://'`: Removes the leading colon, leaving just the tag.

Maybe with awk could be more simple:
`tag=$(echo "$images" | jq -r '.[]' | grep -v ':latest' | awk -F: '{print $NF}' | head -n1)`

PR #4179, opened 2025-08-29
fix: Details link to community module

akucharska:
**Description**

- Fix details link
- Fix installation
- 
**Related issue(s)**

**Definition of done**

- [ ] The PR's title starts with one of the following prefixes:
  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit
- [ ] Explain clearly why you created the PR and what changes it introduces
- [ ] All necessary steps are delivered, for example, tests, documentation, merging


mrCherry97 (changes requested):
Sometimes after the refresh busola is not loading the details of the community modules, sometimes they are loaded, that they are available, but it takes 10-15 seconds, which could be very confusing for the user, could we somehow speed it up, or display spinner for longer time, when we have a proper data if details are available or not

PR #4180, opened 2025-08-29
feat: Add jotai devtools

### LINKED ISSUE #4153: Implement jotai-devtools after Recoil migration
**Description**  
Now that the migration from Recoil to Jotai is complete and all atoms have `debugLabel` set, we should integrate `jotai-devtools` to improve observability and debugging capabilities during development. This will make it easier to track atom state changes and streamline debugging for developers.

**Acceptance Criteria**  
- [x] `jotai-devtools` is installed and configured in the project  
- [x] Devtools work seamlessly in development mode  
- [x] Atom states (with their `debugLabel`s) are visible and update correctly in the devtools  
- [x] Confirm no performance regressions or issues in production builds  

**Attachments**  
- [jotai-devtools GitHub](https://github.com/jotaijs/jotai-devtools)
- [Joati documentation](https://jotai.org/docs/tools/devtools)

KonradPietocha:
**Description**

- Install jotai-devtools,
- Place devtools in proper file together with jotai styles

**Related issue(s)**
Closes #4153 

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/components/JotaiDevTools/JotaiDevToolsStyling.tsx
+++ b/src/components/JotaiDevTools/JotaiDevToolsStyling.tsx
@@ -1,0 +1,11 @@
+import 'jotai-devtools/styles.css';
+
+// This is to conditionally lazy-load styles for dev tools.
+// This prevents them from loading in production and causing problems in testing.
+function JotaiDevToolsStyling({
+  children,
+}: Readonly<{ children?: React.ReactNode }>) {
+  return <>{children}</>;
+}
+
+export default JotaiDevToolsStyling;
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -25,7 +25,7 @@ import './styles/sapIllus-Layout.css';
 import './styles/index.scss';
 import './styles/fiori-helpers.scss';
 import { createRoot } from 'react-dom/client';
-import JotaiDevTools from 'components/App/JotaiDevTools';
+import JotaiDevTools from 'components/JotaiDevTools/JotaiDevTools';
 
 i18next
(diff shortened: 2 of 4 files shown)


src/index.tsx
```diff
@@ -56,13 +57,17 @@ savePreviousPath();
 
 const container = document.getElementById('root');
 const root = createRoot(container!);
+const isDevMode =
+  window.location.hostname === 'localhost' &&
+  (window.location.port === '8080' || window.location.port === '8000');
```
chriskari, line 62: Can you check if we can have a more robust check for dev mode like this?
`const isDevMode = process.env.NODE_ENV === 'development';`

I know that we have something like this in the backend, and maybe even in the frontend?

src/components/JotaiDevTools/JotaiDevToolsStyling.tsx
chriskari, line 1: The CSS import in JotaiDevToolsStyling.jsx still executes at module evaluation time, so styles will load regardless of the lazy wrapper.
This line will already prevent them being loaded in prod or on tests:
`{process.env.NODE_ENV === 'development' && <DevTools />}`

Generally I don't think we need this wrapper. I could find a similar appraoch in the docs but this seems to be specifically for usecases where you want fine-grained control over environment usage.
Since we only want to have it in local development it will be sufficient to simply add this `import 'jotai-devtools/styles.css';` in the main dev tools component and get rid of this wrapper

chriskari (approved):

PR #4272, opened 2025-10-20
fix: Arrow navigation in preferences dialog, list count reading and immediate focus shifting

### LINKED ISSUE #3248: ACC-264.1, ACC-270.3, ACC-270.13 (Level A)_ Lists_ Count is not read, user cannot navigate and immediate focus doesnot shift
v1.1 Internal Message Details
Internal Message:	
Test Plan:	ACC SAP Kyma runtime 2024
Message Priority:	High
Affected UI	Lists
Affected Check Point:	ACC-264.1,ACC-270.3,ACC-270.13
Short text:	ACC-264.1, ACC-270.3, ACC-270.13 (Level A)_ Lists_ Count is not read, user cannot navigate and immediate focus doesnot shift

1.2 Application Details
Application Name	Kyma-dashboard
Technology:	SAP UI5 
Component:	
URL:	https://dashboard.stage.kyma.cloud.sap/cluster/garden-kyma-stage--c-7074971-external/kymamodules
Username/Password:	NA
1.3 Test Environment Details
Web browser:	Google Chrome Version 127.0.6533.89
Operating System:	Microsoft Windows 11 x64
Screen Reader:	JAWS 2024 Version 2024.2406.121 ILM

1.4 JIRA/SNOW Details
Issue:	Bug
Labels:	Accessibility

2 Screen Flow
•	Launch the application using given URL. 
•	Select the ‘Link to Kyma dashboard’ from the Kyma Environment section on your SAP BTP subaccount page. 
•	Go to KubeconfigURL:https://kyma-env-broker.cp.stage.kyma.cloud.sap/kubeconfig/DFEFE4E0-E202-4455-8A61-37511B4D7553
•	Click on  the link https://dashboard.stage.kyma.cloud.sap/clusters and select Connect.
•	Upload the kubeconfing file downloaded in the first step. You can either drag and drop it or select the upload icon to choose it from the device.
•	Next Step
•	Connect to cluster, A new screen get opens
•	Open ‘Profile account select ‘Preference’ button 
•	

3. Issue Type: Manual
Observed Behavior:
On accessing the List items in the left navigation region within the dialog ‘Preference’ ..
1.Count of the List items are not being read out.
2. User cannot navigate using the Arrow keys.
3. On selecting the list items the immediate focus doesn’t shifts to the related UI.

Screenshot:
 
Expected Behavior:
It should always be possible to navigate within the control elements using navigation keys.
Here navigation within the menu items should be possible with arrow keys instead of tab keys.Initial focus should be positioned in such a way that it supports to completed all the associated task.

Providing initial focus on first elements of dialog as 'Automatic synchronization’ radio button makes an application more feasible for all screen reader and keyboard users with visual difficulties.

[ACC-264.1, ACC-270.3, ACC-270.13 (Level A)_ Lists_ Count is not read, user cannot navigate and immediate focus doesnot shift.docx](https://github.com/user-attachments/files/16687787/ACC-264.1.ACC-270.3.ACC-270.13.Level.A._.Lists_.Count.is.not.read.user.cannot.navigate.and.immediate.focus.doesnot.shift.docx)
[Issue processing-09.2021 (2).docx](https://github.com/user-attachments/files/16687789/Issue.processing-09.2021.2.docx)

awoloszyn22:
**Description**

- in the left panel in Preferences dialog arrow keys navigation was implemented
- switching tabs with arrow keys shifts immediate focus to the related UI

**Related issue(s)**
Resolves #3248 

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/components/Preferences/ThemeChooser.tsx
+++ b/src/components/Preferences/ThemeChooser.tsx
@@ -64,7 +64,6 @@ export default function ThemeChooser({
         if (key === 'ArrowDown' && tabId <= availableThemesArray.length - 2) {
           setTabId(tabId + 1);
           setUsedTheme(getThemeName(tabId + 1) || 'sap_horizon');
-          console.log(tabsListRef?.current?.children[tabId].children[0]);
           //@ts-ignore
           tabsListRef?.current?.children[tabId + 1].children[0].focus();
         } else if (key === 'ArrowUp' && tabId >= 1) {
--- a/src/components/Preferences/ThemeChooser.tsx
+++ b/src/components/Preferences/ThemeChooser.tsx
@@ -1,73 +1,48 @@
 import { useAtom } from 'jotai';
+import { RefObject } from 'react';
+import { useTranslation } from 'react-i18next';
+import { List, ListDomRef } from '@ui5/webcomponents-react';
+import { setTheme } from '@ui5/webcomponents-base/dist/config/Theme';
+
 import {
   Theme,
   isSystemThemeDark,
   themeAtom,
 } from 'state/preferences/themeAtom';
-import { setTheme } from '@ui5/webcomponents-base/dist/config/Theme';
-import { ThemePreview } from './ThemePreview/ThemePreview';
-import { useTranslation } from 'react-i18next';
-import { RefObject, useEffect } from 'react';
-import {
-  List,
-  ListItemCustom,
-  TabContainerDomRef,
-  Text,
-} from '@ui5/webcomponents-react';
+import { TileButton } from 'shared/components/TileButton/TileButton';
 
+import { ThemePreview } from './ThemePreview/ThemePreview';
 import './ThemeChooser.scss';
 
-const availableThemesArray: { theme: Theme; id: number }[] = [
-  {
-    theme: 'light_dark',
-    id: 0,
-  },
-  {
-    theme: 'sap_horizon',
-    id: 1,
-  },
-  {
-    theme: 'sap_horizon_dark',
-    id: 2,
-  },
-  {
-    theme: 'sap_horizon_hcw',
-    id: 3,
-  },
-  {
-    theme: 'sap_horizon_hcb',
-    id: 4,
-  },
+const AVAILABLE_THEMES: Theme[] = [
+  'light_dark',
+  'sap_horizon',
+  'sap_horizon_dark',
+  'sap_horizon_hcw',
+  'sap_horizon_hcb',
 ];
 
 interface ThemeChooserProps {
-  keyNavigationEnabled: boolean;
-  tabsListRef: RefObject<TabContainerDomRef>;
+  listRef: RefObject<ListDomRef>;
 }
 
-export default function ThemeChooser({
-  keyNavigationEnabled,
-  tabsListRef,
-}: ThemeChooserProps) {
+export default function ThemeChooser({ listRef }: ThemeChooserProps) {
   const { t } = useTranslation();
   const [theme, setUsedTheme] = useAtom(themeAtom);
 
-  useEffect(() => {
-    if (keyNavigationEnabled) {
-      //@ts-ignore
-      tabsListRef?.current?.children[0].children[0].focus();
-    }
-  }, [keyNavigationEnabled, tabsListRef]);
-
   return (
-    <List>
-      {availableThemesArray.map(({ theme: themeName }) => {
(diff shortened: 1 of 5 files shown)


mrCherry97 (changes requested):
It's not working correctly. I can't use the keyboard to change the theme; it always changes the left navigation in preferences.

mrCherry97 (dismissed):

src/components/Preferences/ThemeChooser.tsx
```diff
@@ -8,22 +8,78 @@ import { setTheme } from '@ui5/webcomponents-base/dist/config/Theme';
...
+          //@ts-ignore
+          tabsListRef?.current?.children[tabId + 1].children[0].focus();
+        } else if (key === 'ArrowUp' && tabId >= 1) {
+          setTabId(tabId - 1);
+          setUsedTheme(getThemeName(tabId - 1) || 'sap_horizon');
+          //@ts-ignore
+          tabsListRef?.current?.children[tabId - 1].children[0].focus();
+        }
+      }
+    },
```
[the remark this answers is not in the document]
awoloszyn22: I agree that the functions look similar, but they actually have a few differences that would complicate the process of making the logic reusable:
-presence of arrow left+right handling only in Preferences
-setting theme is only in ThemeChooser
-different calculation of if conditions for keys
-different indexes of list items that are being focused
Apart from that the reusable function would have to take at least 4 parameters.
Are we sure we want to proceed with separating the function?

src/components/Preferences/ThemeChooser.tsx
```diff
@@ -5,35 +5,69 @@ import {
...
+    id: 2,
+  },
+  {
+    theme: 'sap_horizon_hcw',
+    id: 3,
+  },
+  {
+    theme: 'sap_horizon_hcb',
+    id: 4,
+  },
```
chriskari: you're not using the ids, unnecessary change

src/components/Preferences/ThemeChooser.tsx
```diff
@@ -5,35 +5,69 @@ import {
...
+export default function ThemeChooser({
+  keyNavigationEnabled,
+  tabsListRef,
+}: ThemeChooserProps) {
   const { t } = useTranslation();
   const [theme, setUsedTheme] = useAtom(themeAtom);
 
+  useEffect(() => {
+    if (keyNavigationEnabled) {
+      //@ts-ignore
```
chriskari: if possible avoid //@ts-ignore--it's bad practice in the longrun as it can hide real issues
in this case use `(tabsListRef.current?.children[0].children[0] as HTMLElement)?.focus();` instead

src/components/Preferences/ThemeChooser.tsx
```diff
@@ -42,9 +76,32 @@ export default function ThemeChooser() {
...
+            onKeyDown={(e) => {
+              if (e.key === 'Enter' && !e.shiftKey) {
+                setUsedTheme(themeName);
+                if (theme === 'light_dark') {
+                  if (isSystemThemeDark()) setTheme('sap_horizon_dark');
+                  else setTheme('sap_horizon');
+                } else {
+                  setTheme(theme);
+                }
+              }
```
chriskari: it's the exact same logic as in the onClick so outsource it into a function:
`
const applyTheme = (themeName: Theme) => {
  setUsedTheme(themeName);
  if (themeName === 'light_dark') {
    setTheme(isSystemThemeDark() ? 'sap_horizon_dark' : 'sap_horizon');
  } else {
    setTheme(themeName);
  }
};
`
and then
`onClick={() => applyTheme(themeName)}`

src/components/Preferences/ThemeChooser.tsx
chriskari: you're not using the ids, unnecessary change

chriskari (commented):
Some general remarks:
In `ThemeChooser.tsx` you replaced `TileButton` by `ListItemCustom`, which is good since the native UI5 components usually are better set up for accessibility than our own custom plain components.
However, since the `TileButton` is used in other places as well (the VericalTabs in the Preference Dialog) it would be better if you adjusted the `TileButton` component itself and not just one of its occurences. I think this could also reduce the need for custom workarounds in the `Preferences`

chriskari (approved):

Issue #4650, opened 2026-02-24
Kubeconfig with generic exec credential plugin is incorrectly treated as OIDC


chriskari:
**Description**

When a kubeconfig uses a generic exec credential plugin — i.e., `users[].user.exec` is set but the command is not `kubectl oidc-login` (no `--oidc-issuer-url=` argument, e.g. a custom credential helper, aws eks get-token, or a shell script) — Busola incorrectly treats it as an OIDC cluster and attempts a browser-based OIDC login flow. Since no valid OIDC parameters can be extracted from the exec args, the login attempt fails silently and the user is redirected back to the cluster list without explanation.
Additionally, `findInitialValues` in `EditCluster` would throw a runtime error if the exec plugin had no `args` array defined.

**Expected result**

When a kubeconfig contains a generic exec plugin that cannot be executed in the browser, Busola should:

- Recognise it is not an OIDC flow and not attempt browser-based OIDC authentication.
- Show an authentication form displaying the exec command and its arguments (read-only), with a token input field and an explanation that the user must run the command locally and paste the resulting token.
- Correctly label the authentication type as "Exec plugin" (not "OIDC") in the Edit Cluster form.

**Actual result**
- `EditCluster` throws a runtime error if the exec plugin has no args array defined.
- If the runtime error is fixed, the wizard and Edit Cluster form display OIDC fields (Issuer URL, Client ID, etc.) for a kubeconfig that does not use OIDC.

**Steps to reproduce**
1. create a fresh k3d cluster: `k3d cluster create busola-test`
2. create a ServiceAccount, bind it to the cluster-admin role, and generate a long-lived token:
```
kubectl create serviceaccount busola-admin -n default`, 
```
```
kubectl create clusterrolebinding busola-admin-binding \
  --clusterrole=cluster-admin \
  --serviceaccount=default:busola-admin
```
```
# Generate a token valid for 24 hours and save it to a variable
SA_TOKEN=$(kubectl create token busola-admin -n default --duration=24h)
```
3. Create the mock exec script
```
cat <<EOF > mock-oidc.sh
#!/bin/sh
echo '{
  "apiVersion": "client.authentication.k8s.io/v1beta1",
  "kind": "ExecCredential",
  "status": {
    "token": "${SA_TOKEN}"
  }
}'
EOF
```
```
chmod +x mock-oidc.sh
```
4. Craft the test kubeconfig
```
k3d kubeconfig get busola-test > test-kubeconfig.yaml
```
[...]

PR #4651, opened 2026-02-24
fix: Don't treat generic exec plugin as OIDC

chriskari:
**Description**

- Introduces `isOIDCExec()` to distinguish OIDC exec plugins (kubectl oidc-login) from generic exec plugins, and adds a dedicated UI in the Add Cluster wizard and Edit Cluster form — showing the exec command read-only with a manual token input field and instructions.
- Adds a warning when memory storage is selected for OIDC exec clusters, as in-memory state is lost during the OIDC authentication redirect.

**Related issue(s)**
#4650

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/public/i18n/en.yaml
+++ b/public/i18n/en.yaml
@@ -141,7 +141,7 @@ clusters:
       inMemory: In memory
       localStorage: Local storage
       sessionStorage: Session storage
-    oidc-memory-warning: Memory storage is not compatible with browser-based OIDC login. After the authentication redirect, cluster data stored in memory will be lost. Use session or local storage instead.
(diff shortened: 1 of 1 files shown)


public/i18n/en.yaml
```diff
@@ -141,13 +141,16 @@ clusters:
       inMemory: In memory
       localStorage: Local storage
       sessionStorage: Session storage
+    oidc-memory-warning: Memory storage is not compatible with browser-based OIDC login. After the authentication redirect, cluster data stored in memory will be lost. Use session or local storage instead.
```
grego952:
```suggestion
    oidc-memory-warning: Memory storage is not compatible with browser-based OIDC login. After the authentication redirect, cluster data stored in memory is lost. Use session or local storage instead.
```

PR #4840, opened 2026-04-20
refactor: Adjust ai-chat payload on the frontend

awoloszyn22:
**Description**

- Adjusted kyma companion API calls to pass headers to backend in the body and omit parsing on the backend

**Related issue(s)**

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/backend/companion/companionRouter.js
+++ b/backend/companion/companionRouter.js
@@ -79,6 +79,20 @@ async function handlePublicKey(req, res) {
 }
 
 function extractAuthHeaders(req) {
+  let parsedBody;
+
+  if (typeof req.body === 'object' && !ArrayBuffer.isView(req.body)) {
+    parsedBody = req.body;
+  } else {
+    try {
+      parsedBody = JSON.parse(req.body.toString());
+    } catch (_) {
+      const error = new Error('Invalid JSON in request body');
+      error.status = 400;
+      throw error;
+    }
+  }
+
   const {
     clusterUrl,
     certificateAuthorityData,
@@ -86,10 +100,7 @@ function extractAuthHeaders(req) {
     clientCertificateData,
     clientKeyData,
     sessionId,
-  } =
-    typeof req.body === 'object' && !ArrayBuffer.isView(req.body)
-      ? req.body
-      : JSON.parse(req.body.toString());
+  } = parsedBody;
   return {
     clusterUrl,
     certificateAuthorityData,
--- a/backend/companion/companionRouter.js
+++ b/backend/companion/companionRouter.js
@@ -79,17 +79,24 @@ async function handlePublicKey(req, res) {
 }
 
 function extractAuthHeaders(req) {
+  const {
+    clusterUrl,
+    certificateAuthorityData,
+    clusterToken,
+    clientCertificateData,
+    clientKeyData,
+    sessionId,
+  } =
+    typeof req.body === 'object' && !ArrayBuffer.isView(req.body)
+      ? req.body
+      : JSON.parse(req.body.toString());
   return {
-    clusterUrl: req.headers['x-cluster-url'],
-    certificateAuthorityData:
-      req.headers['x-cluster-certificate-authority-data'],
-    clusterToken: req.headers['x-k8s-authorization']?.replace(
-      /^Bearer\s+/i,
-      '',
-    ),
-    clientCertificateData: req.headers['x-client-certificate-data'],
-    clientKeyData: req.headers['x-client-key-data'],
-    sessionId: req.headers['session-id'],
+    clusterUrl,
+    certificateAuthorityData,
+    clusterToken,
+    clientCertificateData,
+    clientKeyData,
+    sessionId,
   };
 }
 
--- a/src/components/KymaCompanion/api/getChatResponse.ts
+++ b/src/components/KymaCompanion/api/getChatResponse.ts
@@ -37,44 +37,18 @@ type GetChatResponseArgs = {
   t: TFunction;
 };
 
-const fillAuthHeaders = (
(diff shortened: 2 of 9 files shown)


chriskari (dismissed):

src/components/KymaCompanion/api/getChatResponse.ts
```diff
@@ -37,44 +37,18 @@ type GetChatResponseArgs = {
...
-  headers: Record<string, string>,
-  clusterAuth: ClusterAuth,
-) => {
-  if (clusterAuth?.token) {
-    headers['X-K8s-Authorization'] = clusterAuth?.token;
-  } else if (clusterAuth?.clientCertificateData && clusterAuth?.clientKeyData) {
-    headers['X-Client-Certificate-Data'] = clusterAuth?.clientCertificateData;
-    headers['X-Client-Key-Data'] = clusterAuth?.clientKeyData;
-  } else {
-    throw new Error('Missing authentication credentials');
```
mrCherry97, line 50: The old fillAuthHeaders in getChatResponse.ts explicitly threw:
`throw new Error('Missing authentication credentials');`
when neither token nor client cert were provided. The new code silently passes undefined values for clusterToken, clientCertificateData, and clientKeyData. The backend will now receive an unauthenticated payload and fail with a less actionable error from the companion API rather than a clear frontend message.

Please consider restoring the guard:
```
if (!clusterAuth?.token && !(clusterAuth?.clientCertificateData && clusterAuth?.clientKeyData)) {
  throw new Error('Missing authentication credentials');
}
```

backend/companion/companionRouter.js
```diff
@@ -79,17 +79,24 @@ async function handlePublicKey(req, res) {
...
+    clusterUrl,
+    certificateAuthorityData,
+    clusterToken,
+    clientCertificateData,
+    clientKeyData,
+    sessionId,
+  } =
+    typeof req.body === 'object' && !ArrayBuffer.isView(req.body)
+      ? req.body
+      : JSON.parse(req.body.toString());
```
mrCherry97: This dual-parse fallback will throw an unhandled exception if the body is malformed. Since all three companion endpoints rely on this function, a bad request could cause the handler to crash. The JSON.parse should be wrapped in a try/catch (or return a 400 early).

src/components/KymaCompanion/hooks/usePromptSuggestions.ts
```diff
@@ -84,6 +88,16 @@ export function usePromptSuggestions(
           resourceType: resourceType,
           groupVersion: groupVersion,
           resourceName: resourceName,
+          clusterUrl: cluster?.currentContext?.cluster?.cluster?.server,
```
mrCherry97: If the cluster atom isn't populated yet, this sends an undefined value for clusterUrl. The same pattern exists in Chat.tsx. A guard, before calling the API, would surface a clearer error.

tests/integration/tests/companion/test-companion-chat.spec.js
```diff
@@ -57,12 +57,15 @@ context('Test Companion Chat Behavior', () => {
...
-      });
+      const { body } = interception.request;
+      expect(body).to.have.property('resourceName', 'default');
+      expect(body).to.have.property('resourceType', 'Namespace');
+      expect(body).to.have.property('groupVersion', 'v1');
+      expect(body).to.have.property('namespace', '');
+      expect(body).to.have.property('clusterUrl');
+      expect(body).to.have.property('certificateAuthorityData');
+      expect(body).to.have.property('clientCertificateData');
+      expect(body).to.have.property('clientKeyData');
```
mrCherry97, line 68: The switch from deep.equal to have.property means extra unexpected fields in the request body won't fail tests. For stable fields like resourceName, resourceType, groupVersion, the old strictness was useful. Consider keeping deep.equal for the resource context fields and using have.property only for the new auth fields that vary by environment.

mrCherry97 (changes requested):

src/components/KymaCompanion/hooks/usePromptSuggestions.ts
```diff
@@ -82,26 +82,32 @@ export function usePromptSuggestions(
...
+            'Missing required authentication context: cluster or authData is undefined.',
+          );
+        } else {
+          const result = await getPromptSuggestions({
+            post,
+            namespace,
+            resourceType,
+            groupVersion,
+            resourceName,
+            clusterUrl: cluster.currentContext?.cluster?.cluster?.server,
```
mrCherry97, line 100: clusterUrl can still be undefined. Chat.tsx correctly checks cluster.currentContext?.cluster?.cluster?.server as part of its condition — usePromptSuggestions should do the same.

src/components/KymaCompanion/components/Chat/Chat.tsx
```diff
@@ -138,19 +138,27 @@ export const Chat = ({
...
-              cluster.currentContext.cluster.cluster[
-                'certificate-authority-data'
-              ],
-          });
+          if (
+            cluster &&
+            authData &&
+            cluster.currentContext?.cluster?.cluster?.server
+          ) {
+            getFollowUpQuestions({
```
mrCherry97, line 146: getFollowUpQuestions has no credential validation

getChatResponse now validates that either token or cert+key is present before proceeding, but getFollowUpQuestions has no equivalent guard. It will silently send a body with all auth fields as undefined if a user has no credentials.

mrCherry97 (changes requested):

mrCherry97 (approved):

chriskari (approved):

PR #4844, opened 2026-04-20
fix: CodeQL alerts

mrCherry97:
**Description**

- chore: Fix Server-side request forgery alert

**Related issue(s)**
[chore: Fix Server-side request forgery alert](https://github.com/kyma-project/busola/security/code-scanning/47)
[chore: Fix Missing rate limiting](https://github.com/kyma-project/busola/security/code-scanning/58)

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -1,6 +1,7 @@
 /* global  require, process, __dirname */
 import { handleK8sRequests } from './kubernetes/handler';
 import { proxyHandler, proxyRateLimiter } from './proxy.js';
+import rateLimit from 'express-rate-limit';
 import companionRouter from './companion/companionRouter';
 import communityRouter from './modules/communityRouter';
 import { pinoMiddleware, createSlowRequestLogger } from './logging';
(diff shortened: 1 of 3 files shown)


backend/index.js
```diff
@@ -60,7 +61,14 @@ app.use(createSlowRequestLogger(SLOW_REQUEST_THRESHOLD_MS));
 
 app.use('/proxy', proxyRateLimiter, proxyHandler);
 
-app.get('/backend/kubeconfig', (req, res) => {
+const kubeconfigRateLimiter = rateLimit({
+  windowMs: 1 * 60 * 1000,
+  max: 30,
+  standardHeaders: true,
```
chriskari, line 68: We can add 
`message: 'Too many requests, please try again later.',`
like we have it for the other rate limiters as well.

backend/index.js
chriskari, line 20: I think I just noticed that all our rate limiters might be broken. Because our backend sits behind some kind proxy, probably our rate limit applies to the proxy's IP address and not to the client's. So essentially our rate limits would be global rate limits for all users together.

Not 100% sure though, we need to confirm it.

The recommended fix for that would be something like 
`app.set('trust proxy', 1);`
But we need to know if there is only 1 proxy in between or potentially even more. So propably we need to add logging of IP requests to the backend, then deploy it dev/stage, and then inspect the logs if our requests log our IP address or the one of some proxy.

mrCherry97: good catch, I'll create a follow up issue to check it

chriskari (approved):

PR #4892, opened 2026-04-28
chore: React 19 migration

KonradPietocha:
**Description**

- Update packages,
- Refactor to breaking changes

**Related issue(s)**

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


src/state/settings/validateResourcesAtom.ts
```diff
@@ -45,15 +45,6 @@ export const useSyncedValidateResources = () => {
...
   }, [validationFeature]);
 
-  useEffect(() => {
-    if (typeof validateResources === 'boolean' && configPolicies.length > 0) {
-      setValidateResources({
-        isEnabled: validateResources,
-        policies: configPolicies,
-      });
-    }
-  }, [validateResources, configPolicies, setValidateResources]);
```
mrCherry97, line 55: This was a migration shim: for users with the old boolean format stored in localStorage, it would upgrade the atom to the {isEnabled, policies} shape once config policies loaded. Removing it means any user whose stored value is still a plain boolean will never get their policies merged in.
I don't see any adjustment after removal of this, so please consider bringing this back, or add another handling of the validation.

tests/integration/support/helpers.js
```diff
@@ -1,4 +1,4 @@
-export function chooseComboboxOption(selector, optionText, force = false) {
+export function chooseComboboxOption(selector, optionText, force = true) {
```
mrCherry97: This is a global default change — every test that calls chooseComboboxOption without passing an explicit force argument now silently bypasses Cypress's visibility and actionability checks. force: true skips checking whether the element is hidden, obscured, or disabled. This will make tests pass even when the combobox is not actually visible to a real user. The individual test sites that genuinely need force: true should pass it explicitly; the default should stay false.

src/resources/Pods/ContainersLogs.tsx
```diff
@@ -34,7 +28,7 @@ interface ContainersLogsProps {
   containerName: string;
 }
 
-const scrollToSelectedLog = (selectedLogIndex: MutableRefObject<number>) => {
+const scrollToSelectedLog = (selectedLogIndex: RefObject<number>) => {
```
mrCherry97: RefObject<T> in React 19 types current as readonly T | null. Writing to selectedLogIndex.current inside scrollToSelectedLog is technically writing to a readonly property. The correct type for a mutable ref passed around for write access is still MutableRefObject<number> — or the parameter type should reflect that it's writable. This may not be caught by the current TypeScript config (strict mode + noImplicitAny may not catch readonly violations through function boundaries in all cases), but it's a type contract violation worth fixing.

src/resources/createResourceRoutes.tsx
```diff
@@ -37,7 +37,7 @@ export const createPath = (
 };
 
 interface ColumnWrapperProps {
-  list: ReactElement;
+  list: ReactElement<any>;
```
mrCherry97: React 19 changed ReactElement (no generic) to default to ReactElement<unknown> rather than ReactElement<any>, making it stricter. The <any> workaround restores the old behavior but loses the new type safety. Ideally, these would use the actual prop types (ReactElement<ListComponentProps> etc.) or, at a minimum, a comment explaining why any is required here. Minor, but worth noting.

mrCherry97 (changes requested):

src/components/Extensibility/components-form/KeyValuePairRenderer.tsx
```diff
@@ -25,7 +26,7 @@ type DropdownCompProps = {
 const getEnumComponent = (
   enumValues: any,
   isKeyInput: boolean = true,
-  input: () => JSX.Element = Inputs.Text,
```
OliwiaGowor, line 28: Can the type be narrowed down to a more specific one?

src/components/Extensibility/components-form/KeyValuePairRenderer.tsx
```diff
@@ -61,9 +61,9 @@ const getValueComponent = (valueInfo: Record<string, any>) => {
 
   switch (type) {
     case 'number':
-      return getEnumComponent(valueEnum, false, Inputs.Number);
+      return getEnumComponent(valueEnum, false, Inputs.Number as any);
     case 'string':
-      return getEnumComponent(valueEnum, false, Inputs.Text);
```
OliwiaGowor, line 66: Same as above, could the type be narrowed down to a more specific one?

mrCherry97 (commented):
## Code Review: chore: React 19 migration

### Overview

This PR upgrades React 18 → 19, `@ui5/webcomponents-react` 2.19 → 2.22, Jotai 2.13 → 2.20, and related type packages. The bulk of the change is mechanical: `RefObject<T>` → `RefObject<T | null>`, explicit `JSX` imports, and moving `act` from `@testing-library/react` to `react`. The approach is correct and aligns with React 19 breaking changes.

---

### What's handled well ✅

- **`RefObject<T | null>` migration** — React 19 removed the `RefObject<T>` overload that didn't include `null`. All ~30 affected files have been updated correctly.
- **`act` import change** — Moving `act` from `@testing-library/react` to `react` directly in both `setupTests.js` and `useTokenValidation.test.ts` is the right React 19 approach.
- **`JSX` namespace** — The global `JSX` namespace is deprecated in React 19. Explicit `import { JSX } from 'react'` across affected files is correct.
- **`@types/prop-types` removed** — React 19 dropped PropTypes; removing this dependency is appropriate.
- **`MutableRefObject` replaced** — `ContainersLogs.tsx` correctly replaces `MutableRefObject<number>` with the structural `{ current: number }` since `MutableRefObject` was removed in React 19 types.
- **`ResourceValidationSettings` / `ValidationSwitch` toggle fix** — Using a functional updater `setValidateResources(prev => ...)` avoids a potential stale state bug that existed in the previous synchronous read-then-set pattern.

---

### Concerns 🔍

**1. `ref={modalRef as any}` in `TokenRequestModal.tsx`**
The ref is typed as `RefObject<HTMLElement>` but cast to `any` when passed. This silences the type error instead of fixing it — the root cause is likely that the `Dialog`/`Modal` DOM ref type from `@ui5/webcomponents-react` doesn't match `HTMLElement`. A proper fix would be to type `modalRef` with the correct UI5 DomRef type (e.g., `DialogDomRef`).

**2. `fullWidth` prop silently removed in `ComboboxArrayInput.tsx`**
The `fullWidth` prop was removed from the `<Inputs.ComboboxInput>` call without explanation or a CSS replacement. If `@ui5/webcomponents-react` 2.22 dropped this prop from the API, the layout impact should be verified visually — comboboxes inside arrays may now render at a narrower width.

**3. `KeyValuePairRenderer.tsx` bug fix bundled in**
The `valuKeyEnum` → `valueKeyEnum` typo fix is a correctness fix, not a React 19 concern. It's fine to include here but worth calling out so reviewers don't overlook it.

**4. `userModified: true` added in `ResourceValidationSettings.tsx`**
The `userModified` flag is new and appears in two places (add/remove policy), but not in `toggleVisibility`. Is this intentional? If `toggleVisibility` should also mark state as user-modified, it's missing there.

**5. No related issue linked**
The PR body has an empty "Related issue(s)" section. Is there a backlog ticket tracking this upgrade? Linking it would help trace regressions.

**6. `react-table` removed from `@ui5/webcomponents-react` 2.22**
The lockfile shows `react-table: 7.8.0` was a dependency of `@ui5/webcomponents-react` that is now gone. If any Busola code relied on it transitively, it will break silently at runtime. Worth a grep to confirm no internal usage leaks through.

---

### Testing

- The `act` migration and `RefObject` changes are low-risk and type-checked.
- The `fullWidth` removal and `userModified` flag gaps need manual/visual verification.
- Integration tests were updated for namespace cleanup and test stabilization — these look fine.
- No new unit tests were added, which is expected for a pure migration PR.

---

### Summary

The migration is mechanically sound and follows all React 19 upgrade guidance. The main items to address before merge are: (1) replace `as any` on the `TokenRequestModal` ref with a proper UI5 type, (2) confirm the `fullWidth` removal doesn't regress the combobox array UI, and (3) clarify whether `toggleVisibility` should also set `userModified: true`.

mrCherry97 (dismissed):
Image

https://github.com/user-attachments/assets/c83da705-f04c-4204-bca3-99dcc3554711

KonradPietocha:
> ### Summary
> The migration is mechanically sound and follows all React 19 upgrade guidance. The main items to address before merge are: (1) replace `as any` on the `TokenRequestModal` ref with a proper UI5 type, (2) confirm the `fullWidth` removal doesn't regress the combobox array UI, and (3) clarify whether `toggleVisibility` should also set `userModified: true`.

(1) The type for `TokenRequestModal` ref has been corrected.
(2) It's safe to remove `fullWidth`. This is a very old prop and hasn't worked for several previous versions.
(3) No, this isn't necessary. toggleVisibility only changes isEnabled — it doesn't affect policy selection. Setting userModified: true only makes sense in deleteSelectedPolicy and addSelectedPolicy, because there the user consciously modifies the policy selection.

> ### It flickers in some views

Yeah, I'm currently working on it.

OliwiaGowor:
LGTM

src/state/validationSchemasAtom.ts
```diff
@@ -3,7 +3,13 @@ import { authDataAtom } from 'state/authDataAtom';
 import { clusterAtom } from 'state/clusterAtom';
 import jsyaml from 'js-yaml';
 import { atom, useAtomValue, useSetAtom } from 'jotai';
+import { unwrap } from 'jotai/utils';
 import { permissionSetsAtom, PermissionSetState } from './permissionSetsAtom';
+
+const permissionSetsAtomSync = unwrap(
+  permissionSetsAtom,
+  (prev) => prev ?? null,
+);
```
mrCherry97: The const permissionSetsAtomSync = unwrap(...) declaration is inserted between import statements — specifically, import { useUrl } from 'hooks/useUrl' appears after the const declaration. While JS/TS hoists imports at runtime, this violates the import/order ESLint rule and is a stylistic regression. The unwrap(...) call should be placed after all imports, or the import for unwrap should be kept with the other jotai imports and the const moved after all imports

src/resources/Namespaces/NamespaceList.tsx
```diff
@@ -48,7 +48,7 @@ export function NamespaceList(props: any) {
   };
 
   useEffect(() => {
-    if (!hasPermissions) {
+    if (hasPermissions === false) {
```
mrCherry97, line 51: useHasPermissionsFor returns null while loading ([src/hooks/useHasPermissionsFor.js](vscode-webview://06sslph93p21sm41g6d7hceibfhfvh28a4di4qs0usu0dsc445lp/src/hooks/useHasPermissionsFor.js)): Before this PR the hook always returned a boolean. Now it returns null during the async load window. The two callers handle this:

NamespaceList.tsx:51— if (!hasPermissions) treats null as falsy and redirects away from the namespace. This might cause a brief unwanted redirect on first load before permissions resolve.
GenericRoleBindingCreate.tsx:93— skip: !hasPermissionsForClusterRoles skips the fetch while null. This is safe.
The NamespaceList case is worth auditing to confirm there's no observable flicker or unwanted redirect.

mrCherry97 (dismissed):

src/resources/Namespaces/NamespaceList.tsx
KonradPietocha: This change is fixing the problem with redirecting to '/no-permissions' before fetch

mrCherry97 (commented):
LGTM

akucharska (approved):
LGTM

PR #4906, opened 2026-05-05
fix: Replace backend's per-IP rate limiting 

### LINKED ISSUE #4848: fix: Investigate and fix rate limiter broken by proxy IP address
**Description**

Our backend sits behind a proxy, which likely causes all rate limiters to apply to the proxy's IP address rather than the actual client IP. This means rate limits are effectively global across all users simultaneously, rather than per-client.

This needs to be confirmed first by adding IP request logging to the backend, deploying to dev/stage, and inspecting the logs to determine whether requests are logged with the client IP or a proxy IP.

Once confirmed, the recommended fix is to configure Express to trust the proxy:

```
app.set('trust proxy', 1);
```

However, the exact value depends on how many proxies sit between the client and the backend, which is why the logging investigation is needed first.

**Reasons**

If rate limits are applied to the proxy IP, they act as a global throttle shared by all users. This means:
- A single busy user can trigger rate limits for everyone
- Per-user rate limiting provides no actual protection

**Acceptance Criteria**

- [x] Add IP address logging to the backend request handler
- [x] Deploy to dev/stage and inspect logs to confirm whether client or proxy IP is being logged
- [ ] Determine the number of proxy hops between client and backend
- [ ] Configure `trust proxy` in Express with the correct value based on findings
- [ ] Verify that rate limiting correctly targets individual client IPs after the fix

**Testing Strategy (DoD)**

- automated unit test for the new logic
- automated integration test (a new test suite or extending an existing test)

chriskari:
**Description**

- Removed per-IP rate limiting on all endpoints on the express app-level (confirmed to be broken, already covered by Cloud Armor)
- Replaced by per-K8s-credential rate limiting where applicable
- Added default per-IP rate limiting (50 RPS, 50 connections) on the shipped Ingress via ingress-nginx annotations
- Documented the rate-limiting model in the deployment guide

**Related issue(s)**
#4848

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/backend/utils/rate-limit-key.js
+++ b/backend/utils/rate-limit-key.js
@@ -1,0 +1,47 @@
+import crypto from 'crypto';
+
+export function hashCredential(credential, prefix) {
+  return (
+    prefix +
+    ':' +
+    crypto.createHash('sha256').update(credential).digest('hex').slice(0, 32)
+  );
+}
+
+function tryParseBody(req) {
+  try {
+    if (typeof req.body === 'object' && !ArrayBuffer.isView(req.body)) {
+      return req.body;
+    }
+    if (req.body) {
+      return JSON.parse(req.body.toString());
+    }
+  } catch (_) {
+    // ignore — extractors must not throw
+  }
+  return null;
+}
+
+export function getK8sCredentialFromHeaders(req) {
+  return (
+    req.headers['x-k8s-authorization'] ||
+    req.headers['x-client-certificate-data']
+  );
+}
+
+export function getK8sCredentialFromBody(req) {
+  const parsed = tryParseBody(req);
+  return parsed?.clusterToken || parsed?.clientCertificateData;
+}
+
+// Reject before the rate limiter when no credential is present, so a missing
+// credential cannot fall back to req.ip and become a global throttle key.
+export function requireCredential(extractCredential) {
+  return function (req, res, next) {
+    if (!extractCredential(req)) {
+      res.status(400).json({ error: 'Missing K8s credentials.' });
+      return;
+    }
+    next();
+  };
+}
--- a/backend/utils/rate-limit-key.js
+++ b/backend/utils/rate-limit-key.js
@@ -16,8 +16,12 @@ function tryParseBody(req) {
     if (req.body) {
       return JSON.parse(req.body.toString());
     }
-  } catch (_) {
-    // ignore — extractors must not throw
+  } catch (err) {
+    // Don't rethrow — runs inside the rate-limit keyGenerator which must not throw.
+    req.log?.warn(
+      err,
+      'Failed to parse request body for credential extraction',
+    );
   }
   return null;
 }
--- a/backend/utils/rate-limit-key.js
+++ b/backend/utils/rate-limit-key.js
@@ -34,8 +34,8 @@ export function getK8sCredentialFromBody(req) {
   return parsed?.clusterToken || parsed?.clientCertificateData;
 }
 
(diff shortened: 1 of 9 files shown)


backend/utils/rate-limit-key.js
```diff
@@ -0,0 +1,47 @@
...
+function tryParseBody(req) {
+  try {
+    if (typeof req.body === 'object' && !ArrayBuffer.isView(req.body)) {
+      return req.body;
+    }
+    if (req.body) {
+      return JSON.parse(req.body.toString());
+    }
+  } catch (_) {
+    // ignore — extractors must not throw
```
dbadura: At least log it, because we will never be informed about error in that section.

backend/utils/rate-limit-key.js
```diff
@@ -0,0 +1,51 @@
...
+function tryParseBody(req) {
+  try {
+    if (typeof req.body === 'object' && !ArrayBuffer.isView(req.body)) {
+      return req.body;
+    }
+    if (req.body) {
+      return JSON.parse(req.body.toString());
+    }
+  } catch (err) {
+    // Don't rethrow — runs inside the rate-limit keyGenerator which must not throw.
```
dbadura: Why?
I cound't find anything in docs: https://express-rate-limit.mintlify.app/reference/configuration#keygenerator

chriskari: You're right, the docs don't actually require that — the comment was off. Real reason for the swallow: tryParseBody is called by requireCredential before the rate limiter even runs, so an uncaught JSON.parse would 500 the request instead of returning the clean 400 from requireCredential. Removed the misleading comment.

dbadura (dismissed):

docs/user/01-40-deploy-access-kubernetes.md
```diff
@@ -7,6 +7,14 @@
 To expose Busola, you can use [APIRule](https://github.com/kyma-project/busola/tree/main/resources/istio), [Ingress](https://github.com/kyma-project/busola/tree/main/resources/ingress), or create your own exposing mechanism.
 For more details about environment configuration, see [Environment-Specific Settings](../user/technical-reference/configuration.md#environment-specific-settings).
 
+### Per-IP Rate Limiting
```
grego952, line 10: Can you tell me more about this section? Why is it here? Is it relevant for the users who want to deploy Busola on their clusters? If something must be configured on the user side, maybe we should provide instructions on how to do so? WDYT?

chriskari: You're right, maybe there is too little context for newcomers to understand the purpose of this section.

To sum it up for you:
This PR removes the per-IP rate limiting we previously had in the backend code (it was broken anyway, because behind our reverse proxy every request looked like it came from the same IP). For our managed Google Cloud Run dev/stage/prod deployments Cloud Armor at the edge already handles per-IP throttling, so that side is covered.
For self-host deployments (which this doc is relevant for) there's no Cloud Armor, so we shifted that responsibility to the ingress layer. The shipped resources/ingress/ingress.yaml now includes default per-IP rate-limit annotations for ingress-nginx (50 RPS, 50 connections). The section is there to inform operators that these defaults exist, where to tune them, and what to do if they use a different ingress controller (Traefik etc.) or expose Busola via APIRule/Istio instead.

grego952 (approved):

dbadura (approved):

PR #4974, opened 2026-05-20
feat: Protect against SSRF (SEC-375).

ManuelRucker:
**Description**

- validated and implemented SSRF Protection using Claude Code with SEC-375 as context
- all validation findings can be seen in `SEC-375-audit-findings.md`

**Related issue(s)**

- part of Agentic Compliance Frontrunner - No related issue

### DIFF
--- a/backend/request-filters.js
+++ b/backend/request-filters.js
@@ -1,5 +1,5 @@
 import net from 'net';
-import { isPrivateIp } from './utils/network-utils.js';
+import { isPrivateIp, isPrivateAddressCached } from './utils/network-utils.js';
 import config from './config.js';
 
 export const localIpFilter = (_req, headersData) => {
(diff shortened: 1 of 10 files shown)


backend/request-filters.js
```diff
@@ -93,6 +93,35 @@ export const portFilter = (_req, headersData) => {
...
+    return;
+  }
+
+  // Bare IPs are already handled by the synchronous localIpFilter.
+  // This filter resolves hostnames to catch domains that map to private addresses.
+  if (net.isIP(hostname) !== 0) {
+    return;
+  }
+
+  if (await isPrivateAddressCached(hostname)) {
```
KonradPietocha, line 122: it caches the result for 5 minutes, then the actual HTTP request is made with a separate DNS resolution inside Node's TLS stack. An attacker with DNS control can serve a public IP during the cache lookup and flip to a private IP before the connection is made.

PR #5025, opened 2026-06-02
feat: Adjust UI5Card and Selector to the review

mrCherry97:
**Description**

- feat: Adjust UI5Card and Selector to the review

**Related issue(s)**

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/shared/components/UI5Card/UI5Card.tsx
+++ b/src/shared/components/UI5Card/UI5Card.tsx
@@ -1,5 +1,5 @@
 import { forwardRef, ReactNode, useContext } from 'react';
-import { Card, Title } from '@ui5/webcomponents-react';
+import { Card, CardHeader, Title } from '@ui5/webcomponents-react';
 import { Toolbar } from '@ui5/webcomponents-react-compat/dist/components/Toolbar/index.js';
 import { ToolbarSpacer } from '@ui5/webcomponents-react-compat/dist/components/ToolbarSpacer/index.js';
 import { NestedContainerContext } from './NestedContainerContext';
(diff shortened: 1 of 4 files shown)


mrCherry97:
https://www.figma.com/design/J1jLOhcjIur9Y2Ui4jYFYv/Kyma-Dashboard-Designs?node-id=19443-30&p=f&t=bbglDOIf1XVyzHIf-0

src/shared/components/UI5Card/UI5Card.tsx
chriskari, line 41: we can add ` toolbarStyle="Clear"` here to remove the border

chriskari (commented):
Maybe in Panel.tsx we should also give the title the attributes `size="H6" level="H6"`?
I think you mentioned that we need to revert the migration of Panel to using Ui5Card, but I could imagine that probably the same title styling convention should apply there?

edit: Ah I saw you have created https://github.com/kyma-project/busola/pull/5027, so this comment applies to this PR :D

src/shared/components/UI5Card/UI5Card.tsx
mrCherry97: I opened an [issue for UI5](https://github.com/UI5/webcomponents-react/issues/8688) to provide proper support for this property, but I'm also wondering whether we need it. We are creating a new Widget, so we can consider restricting it to have "string only," and then we can use only `UI5CardHeader` and remove the Toolbar from here completely.

mrCherry97:
> Maybe in Panel.tsx we should also give the title the attributes `size="H6" level="H6"`? I think you mentioned that we need to revert the migration of Panel to using Ui5Card, but I could imagine that probably the same title styling convention should apply there?
> 
> edit: Ah I saw you have created #5027, so this comment applies to this PR :D

Yeah, I reverted a Panel migration. I don't think it's necessary; it was not a problem in the past, so we will see with the newest UXC requirements later.

chriskari (approved):

PR #5030, opened 2026-06-03
feat: Re-enable SSO in busola app

### LINKED ISSUE #3680: Re-enable SSO in busola
**Description**
User is supposed to be logged in with SAP account to use busola backend.

Acceptance criteria:
- [ ] each request to backend service should require authenticated user token
- [ ] introduce rate limiting for users to prevent abuse of busola backend

**Reasons**
Busola backend could be abused if it is accessible anonymously. Tracking users help us to prevent or at least stop malicious usage of backend service. 

**Attachments**
- [Old issue for enable tracking](https://github.com/kyma-project/busola/issues/879)
- [PR with removal of the SSO](https://github.com/kyma-project/busola/pull/2093)

KonradPietocha:
**Description**

- Re-enable SSO in busola

**Related issue(s)**
Closes #3680 

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


backend/environments/dev/config.yaml
```diff
@@ -4,6 +4,11 @@ config:
       isEnabled: false
     GZIP:
       isEnabled: true
+    JWT_CHECK_CONFIG:
+      isEnabled: false
```
mrCherry97: We agreed to enable that on dev to double-check that everything is working. Please enable that.

mrCherry97 (changes requested):

backend/environments/dev/config.yaml
KonradPietocha: That's because it seems like the tests are using it and won't pass. That was the fastest way. I'll see if I can do something else.

src/shared/hooks/BackendAPI/checkForTokenExpiration.ts
chriskari, line 1: unused files. should be deleted

backend/jwtCheck.js
```diff
@@ -0,0 +1,39 @@
...
+    algorithms: ['RS256'],
+  });
+
+const userRateLimiter = rateLimit({
+  windowMs: 1 * 60 * 1000,
+  max: 300,
+  message: 'Too many requests, please try again later.',
+  standardHeaders: true,
+  legacyHeaders: false,
+  keyGenerator: (req) => req.auth?.sub || req.ip,
```
chriskari: For rate limiting, we cannot key on IP addresses in NodeJS as found out and fixed in #4848
(req.ip would show the IP of a Google Cloud Run proxy for our own deployments there instead of client IP)

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+export function useIsSSOEnabled() {
+  const configuration = useAtomValue(configurationAtom);
+  return configuration?.features?.SSO_LOGIN?.isEnabled ?? false;
+}
+
+function createSSOAuth(ssoConfig: ConfigFeature) {
+  const { issuerUrl, clientId, clientSecret, scope } = ssoConfig.config;
+
+  return new UserManager({
+    redirect_uri: window.location.origin + '',
```
chriskari: Tiny thing: the + '' here doesn't do anything since origin is already a string. Can drop it.

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+}
+
+async function handleSSOLogin(
+  ssoConfig: ConfigFeature,
+  setSsoState: (user: SsoDataState) => void,
+) {
+  if (!ssoConfig || !ssoConfig.config) {
+    throw new Error('SSO configuration not found');
+  }
+  const userManager = createSSOAuth(ssoConfig);
```
chriskari: We end up building a brand new UserManager every time this runs, and each one starts its own token-refresh loop in the background. If the function runs more than once we'd have a few of those going at the same time. We should create it once and reuse it

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    userManager.events.addAccessTokenExpiring(async () => {
+      const refreshedUser = await userManager.signinSilent();
+      setSSOAuthData(refreshedUser);
+      setSsoState(refreshedUser);
+    });
+    userManager.events.addSilentRenewError((err) => {
+      console.warn('silent renew failed', err);
+      setSsoState(null);
+      setSSOAuthData(null);
```
chriskari: Same idea as above — these handlers get attached fresh on every run, so they could stack up and fire multiple times. Might be worth attaching them just once.

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+    });
+    document.addEventListener('visibilitychange', async () => {
+      if (document.visibilityState === 'visible') {
+        if (!!user?.expired || (user?.expires_in && user?.expires_in <= 2)) {
+          const refreshedUser = await userManager.signinSilent();
+          setSSOAuthData(refreshedUser);
+          setSsoState(refreshedUser);
+        }
+      }
+    });
```
chriskari: Couple of things here: it looks at user from when login first happened, so if the token expires while the tab is in the background I don't think we'd catch it when the user comes back. Also there's no cleanup for this listener, so it'd pile up alongside the others.

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    setSSOAuthData(user);
+    setSsoState(user);
+  } catch (e) {
+    if (e instanceof Error) {
+      await userManager.clearStaleState();
+      userManager.signinRedirect();
+    } else {
+      throw e;
+    }
```
chriskari: The else branch never runs since anything thrown here is already an Error. The bigger worry is that we always retry the redirect — if the IdP sends us back with an ?error=… instead of a code, I think we'd just bounce in a loop. Could we handle that case so it doesn't keep redirecting?

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+  if (!token) return;
+
+  const timeout = 30; // s
+  try {
+    const expirationTimestamp = (jwtDecode(token) as JwtPayload).exp!;
+    const secondsLeft = expirationTimestamp - Math.floor(Date.now() / 1000);
+
+    if (secondsLeft < timeout) {
+      setSSOAuthData(null);
+      window.location.reload();
```
chriskari: Doing a full page reload when the token's about to expire feels a bit harsh. Not blocking, just wondering if a quieter token refresh would be nicer here.

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+function createSSOAuth(ssoConfig: ConfigFeature) {
+  const { issuerUrl, clientId, clientSecret, scope } = ssoConfig.config;
+
+  return new UserManager({
+    redirect_uri: window.location.origin + '',
+    post_logout_redirect_uri: window.location.origin + '/logout.html',
+    loadUserInfo: false,
+    client_id: clientId,
+    authority: issuerUrl,
+    client_secret: clientSecret,
```
chriskari: Is this client registered in IAS as a public or confidential client? If it's public (PKCE, no secret), passing client_secret from the frontend doesn't do anything useful and we may want to drop it. If it's confidential, then sending the secret from the browser is a concern worth discussing.

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    const startLogin = async () => {
+      const bypass = await getEnv(Envs.SSO_LOGIN_BYPASS);
+      if (bypass === 'true') return;
+      handleSSOLogin(ssoConfig, setSsoState);
+    };
+    startLogin();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isSSOEnabled, ssoConfig]);
```
chriskari, line 169: A config reload can change its reference (even when nothing really changed) and re-run this effect. If that happens while login is still in progress, it'd start a second login on top of the first. I think this mostly goes away once the UserManager is created/reused once with its listeners attached a single time

chriskari (commented):
One more backend security thing: the JWT check validates the issuer but not the audience, so any token from that IAS tenant would be accepted — including ones issued for a different app on the same tenant. Could we also verify the aud claim matches Busola's client ID, so only tokens actually meant for us pass?

docs/user/technical-reference/feature-flags.md
```diff
@@ -39,4 +40,5 @@ This table lists the configurable frontend feature flags, their descriptions, an
...
+| **JWT_CHECK_CONFIG**  | Configures the data necessary for backend authentication, such as an issuer and JWKS (JSON Web Key Set) address. When the feature is disabled, no authentication occurs on the backend side. Backend feature. Cannot be modified at the cluster's ConfigMap level.                                                                                                                                                                                                                                                                                                                                                                                                                                                            | <ul><li>`isEnabled: false`</li><li>`config.issuer: https://apskyxzcl.accounts400.ondemand.com`</li><li>`config.jwksUri: https://apskyxzcl.accounts400.ondemand.com/oauth2/certs`</li></ul>                             |
```
grego952:
```suggestion
| **JWT_CHECK_CONFIG**  | Configures the data necessary for backend authentication, such as an issuer and JWKS (JSON Web Key Set) address. When the feature is disabled, no authentication occurs on the backend side.                                                                                                                                                                                                                                                                                                                                                                                                                                                          | <ul><li>`isEnabled: false`</li><li>`config.issuer: https://apskyxzcl.accounts400.ondemand.com`</li><li>`config.jwksUri: https://apskyxzcl.accounts400.ondemand.com/oauth2/certs`</li></ul>                             |
```
Not needed, since it's already mentioned in the Configurable Flags for Backend section and in the Note.

src/shared/hooks/BackendAPI/useFetch.ts
```diff
@@ -34,11 +41,13 @@ export const createFetchFn =
     init?: any;
     abortController?: AbortController;
   }) => {
+    checkForTokenExpiration(ssoData?.id_token);
```
OliwiaGowor, line 44: The `addAccessTokenExpiring` event in `oidc-client-ts` already fires when the access token is about to expire. The `checkForTokenExpiration` call at the start of every request in 'useFetch' is redundant. Decoding the JWT and firing an async silent refresh on every fetch is wasteful. It could be resolved either by: 
1. Removing `checkForTokenExpiration(ssoData?.id_token)` from `useFetch.ts` entirely - the event handlers in handleSSOLogin already cover expiration.
2. If kept as an additional check, adding the throttle guard and running it once every 30 seconds so it doesn't decode a JWT on every network request.

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const timeout = 30; // s
+  try {
+    const expirationTimestamp = (jwtDecode(token) as JwtPayload).exp!;
+    const secondsLeft = expirationTimestamp - Math.floor(Date.now() / 1000);
+
+    if (secondsLeft < timeout) {
+      trySilentRefresh().then((ok) => {
+        if (!ok) {
+          setSSOAuthData(null);
+          window.location.reload();
```
OliwiaGowor, line 189: A full page reload when silent refresh fails is a harsh UX - the user loses their current view state. The `trySilentRefresh()` path is good, but the reload fallback could instead redirect to a "session expired" page or show a dialog. No exactly a blocker but maybe something to consider as a follow up

src/state/ssoDataAtom.tsx
```diff
@@ -0,0 +1,191 @@
...
+function getOrCreateUserManager(ssoConfig: ConfigFeature): UserManager {
+  if (!userManager) {
+    const { issuerUrl, clientId, clientSecret, scope } = ssoConfig.config;
+    userManager = new UserManager({
+      redirect_uri: window.location.origin,
+      post_logout_redirect_uri: window.location.origin + '/logout.html',
+      loadUserInfo: false,
+      client_id: clientId,
+      authority: issuerUrl,
+      client_secret: clientSecret,
```
OliwiaGowor: `ssoDataAtom.tsx` reads `clientSecret` from `ssoConfig.config` and passes it directly to `UserManager`. Sending a client secret from the browser is a fundamental OAuth security violation - a browser cannot keep secrets, and anything in the config YAML is visible to every user. If the IAS client is meant to be public (PKCE), `clientSecret` should be removed from `UserManager` options and from the config schema entirely. If it is confidential, the token exchange must be moved to the backend.

backend/environments/dev/config.yaml
```diff
@@ -4,6 +4,11 @@ config:
       isEnabled: false
     GZIP:
       isEnabled: true
+    JWT_CHECK_CONFIG:
+      isEnabled: true
+      config:
+        issuer: https://apskyxzcl.accounts400.ondemand.com
+        jwksUri: https://apskyxzcl.accounts400.ondemand.com/oauth2/certs
```
chriskari, line 11: `clientId: d0316c58-b0fe-45cd-9960-0fea0708355a`

should be added here as well

backend/jwtCheck.js
```diff
@@ -0,0 +1,40 @@
...
+  keyGenerator: (req) => req.auth?.sub,
+});
+
+export function setupJWTCheck(app) {
+  const jwtConfig = config?.features?.JWT_CHECK_CONFIG;
+
+  if (!jwtConfig?.isEnabled || !jwtConfig?.config) {
+    return;
+  }
+
```
chriskari, line 35: maybe add some guard here to make sure clientId will be set?
  
```javascript
if (!jwtConfig.config.clientId) {
    throw new Error(
      'JWT_CHECK_CONFIG is enabled but clientId is empty — audience validation would be skipped. Set clientId to the SSO client ID.',
    );
  }
```

docs/user/technical-reference/feature-flags.md
```diff
@@ -39,4 +40,5 @@ This table lists the configurable frontend feature flags, their descriptions, an
...
+| **JWT_CHECK_CONFIG**  | Configures the data necessary for backend authentication, such as an issuer and JWKS (JSON Web Key Set) address. When the feature is disabled, no authentication occurs on the backend side.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  | <ul><li>`isEnabled: false`</li><li>`config.issuer: https://apskyxzcl.accounts400.ondemand.com`</li><li>`config.jwksUri: https://apskyxzcl.accounts400.ondemand.com/oauth2/certs`</li></ul>                             |
```
chriskari: and here the clientId flag should be listed as a required prop

chriskari (approved):

mrCherry97 (approved):

grego952 (approved):

mrCherry97:
Waiting for @OliwiaGowor review

OliwiaGowor (approved):

PR #5107, opened 2026-06-25
feat: Backend terminal proxy

### LINKED ISSUE #4920: Implement terminal connection in backend
**Description**
Implement backend endpoint to attach to the running pod.
The flow should be the same as k8s.
Do the request to attach to the desired pod and then upgrade connection to web socket.
The backend should be web socket Pipe (proxy) to send messages back and for between frontend and backend.

Example log of communication from `kubectl exec -v 6 -ti test-pod -- bash`, but for `attach` it should use `attach` endpoint.
```bash
I0403 13:11:16.393079   39806 round_trippers.go:632] "Response" verb="GET" url="https://0.0.0.0:46167/api/v1/namespaces/default/pods/test-pod" status="200 OK" milliseconds=5
I0403 13:11:16.394257   39806 podcmd.go:88] Defaulting container name to test-pod
I0403 13:11:16.397999   39806 round_trippers.go:632] "Response" verb="GET" url="https://0.0.0.0:46167/api/v1/namespaces/default/pods/test-pod/exec?command=bash&container=test-pod&stdin=true&stdout=true&tty=true" status="101 Switching Protocols" milliseconds=3
I0403 13:11:16.398018   39806 websocket.go:133] The subprotocol is v5.channel.k8s.io
I0403 13:11:16.398094   39806 websocket.go:251] Websocket initial read deadline: 1m1s
I0403 13:11:21.398948   39806 websocket.go:498] Websocket Ping succeeeded
I0403 13:11:21.399255   39806 websocket.go:456] Pong message received ()--resetting read deadline
```

**Acceptance Criteria**
- [ ] The connection proxy is established.

**Testing Strategy (DoD)**

- automated unit test for the new logic
- automated integration test (a new test suite or extending an existing test)

**Attachments**

**Related Issues**
Epic: https://github.com/kyma-project/busola/issues/2782

dbadura:
**Description**

-Implement backend websocket proxy

**Related issue(s)**
https://github.com/kyma-project/busola/issues/4920

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


backend/src/terminal/handler.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export default function registerWebSocket(server) {
+  const wss = new WebSocketServer({ server });
+
+  wss.on('connection', (frontWS, req) => {
+    const logger = pinoWebSocketLogger(req);
+    logger.info('Starting WebSocket connection proxy');
+    try {
+      const url = new URL(req.url, 'htt://' + req.socket.remoteAddress);
```
mrCherry97:
```suggestion
      const url = new URL(req.url, 'http://' + req.socket.remoteAddress);
```

htt:// instead of http://. new URL() will throw on an invalid base URL, which is caught by the outer catch and closes the connection with a 1011, but this means the proxy never works - every connection fails at this line.

backend/src/terminal/handler.js
```diff
@@ -0,0 +1,116 @@
...
+
+      let opts;
+      if (parsedHeaders.token) {
+        opts = {
+          ca: parsedHeaders.ca,
+          headers: {
+            Authorization: parsedHeaders.token,
+          },
+        };
+      } else if (parsedHeaders.clientKey && parsedHeaders.clientKey) {
```
mrCherry97:
```suggestion
      } else if (parsedHeaders.clientKey && parsedHeaders.clientCert) {
```

question if `parsedHeaders.ca` also should be checked

backend/src/terminal/handler.js
```diff
@@ -0,0 +1,116 @@
...
+          frontWS.send(data);
+        } else {
+          logger.info(
+            'Front WS is not open, cannot send message, status: ' +
+              frontWS.readyState,
+          );
+        }
+      });
+
+      k8sWS.addEventListener('onerror', (event) => {
```
mrCherry97: If I'm correct, the ws library uses DOM-style event names: 'error' and 'close', not 'onerror'/'onclose'. These listeners never fire - k8s-side errors and closes are silently ignored, leaving the frontend WebSocket dangling open.

backend/src/terminal/handler.js
```diff
@@ -0,0 +1,116 @@
...
+function encodeMsg(input, std = Stream.STDIN) {
+  const msgEncoded = Buffer.from(input + '\n', 'utf-8');
+  const encodedMsgForK8s = new Uint8Array(msgEncoded.length + 1);
+  encodedMsgForK8s[0] = std; // set STD[IN,OUT,ERR]
+  encodedMsgForK8s.set(msgEncoded, 1);
+  return encodedMsgForK8s;
+}
+
+export default function registerWebSocket(server) {
+  const wss = new WebSocketServer({ server });
```
mrCherry97: new WebSocketServer({ server }) intercepts every WebSocket upgrade on the HTTP server regardless of path. Any WS connection to the server (e.g. Vite HMR in dev, or future features) will go through the terminal proxy logic. Should filter by path prefix (e.g. req.url.startsWith('/backend/ws/')).

backend/index.js
```diff
@@ -121,6 +123,8 @@ if (isDocker) {
   app.use('/backend', requireK8sCredential, k8sRateLimiter, handleK8sRequests);
 }
 
+registerWebSocket(server);
```
mrCherry97: HTTP routes use requireK8sCredential and k8sRateLimiter middleware, but registerWebSocket(server) bypasses all Express middleware. Any unauthenticated client can open a WebSocket and the handler will forward connections to arbitrary cluster URLs from the protocol header.

Won't it be better to add `, requireK8sCredential, k8sRateLimiter`?

backend/src/terminal/protocolHeaderParser.js
```diff
@@ -0,0 +1,33 @@
...
+  const clientCert = extractedHeaders.get('x-client-certificate-data');
+  const ca = extractedHeaders.get('x-cluster-certificate-authority-data');
+  const clusterURL = extractedHeaders.get('x-cluster-url');
+  const token = extractedHeaders.get('x-k8s-authorization');
+
+  return {
+    protocol,
+    clientKey: Buffer.from(clientKey ?? '', 'base64').toString(),
+    clientCert: Buffer.from(clientCert ?? '', 'base64').toString(),
+    ca: Buffer.from(ca, 'base64').toString(),
```
mrCherry97, line 45:
```suggestion
    ca: Buffer.from(ca ?? '', 'base64').toString(),
```

clientKey and clientCert both use ?? '' but ca does not. If x-cluster-certificate-authority-data is missing from the headers, ca is undefined and Buffer.from(undefined) throws a TypeError.

backend/src/terminal/protocolHeaderParser.test.js
```diff
@@ -0,0 +1,39 @@
+import parseProtocolHeaders from './protocolHeaderParser';
+
+function encodeBase64URL(input) {
+  return Buffer.from(input).toString('base64url');
+}
+
+function encodeBase64(input) {
+  return Buffer.from(input).toString('base64url');
```
mrCherry97: Shouldn't this be:
```suggestion
  return Buffer.from(input).toString('base64');
```
?

backend/src/terminal/protocolHeaderParser.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+describe('Parse protocl headers with auth data', () => {
+  it('All data is presed and mTLS is used', () => {
+    //GIVEN
+    const protocolHeader = `v4.channel.k8s.io, base64url.header.x-client-certificate-data.value.${clientCertEncoded}, base64url.header.x-cluster-url.value.${clusterURLEncoded}, base64url.header.x-cluster-certificate-authority-data.value.${caEncoded}, base64url.header.x-client-key-data.value.${clientKeyEncoded}`;
+
+    //WHEN
+    const output = parseProtocolHeaders(protocolHeader);
+
+    console.log(output);
```
mrCherry97:
```suggestion
```

backend/src/terminal/protocolHeaderParser.js
```diff
@@ -0,0 +1,33 @@
+/* global Buffer */
+// TODO: implement stricter validation
+export default function parseProtocolHeaders(secHeader) {
+  const headers = secHeader.split(',');
+  if (headers.length < 1) {
```
mrCherry97: dead code.
''.split(',') returns [''] (length 1), never []. An empty secHeader silently produces an empty map with protocol = '' instead of throwing.

mrCherry97 (changes requested):

backend/src/terminal/handler.js
dbadura: If it would be like that, the connection will be always broken.
But in general I will fix this typo and add comment.

backend/src/terminal/handler.js
dbadura: https://github.com/websockets/ws/blob/HEAD/doc/ws.md#websocketaddeventlistenertype-listener-options

backend/index.js
dbadura: I will look at k8sRateLimiter, 
requireK8sCredential - is no go, as the credentials are passed in `protocol` header

backend/src/terminal/protocolHeaderParser.js
dbadura: Because the `ca` is required.
Maybe I will write it in cleaner way.

backend/src/terminal/protocolHeaderParser.test.js
dbadura: base64url is used because base64 has forbidden sings in web broweser websocket api. It's documented in frontend. I will check the backend

backend/src/terminal/handler.js
dbadura: In my opinion no, because we need two things:
- token 
- mTLS 
The CA is required in all of the above cases and it should be catch earlier or in parse method.

backend/src/terminal/handler.js
dbadura: O probably I removed :O

backend/src/terminal/handler.js
akucharska: Exactly, so should be:
`k8sWS.addEventListener('error', (event) => {`

backend/src/terminal/handler.js
```diff
@@ -0,0 +1,116 @@
...
+              frontWS.readyState,
+          );
+        }
+      });
+
+      k8sWS.addEventListener('onerror', (event) => {
+        logger.error({ err: event }, 'K8s WebSocket error: ');
+      });
+
+      k8sWS.addEventListener('onclose', () => {
```
akucharska:
```suggestion
      k8sWS.addEventListener('close', () => {
```

backend/src/terminal/protocolHeaderParser.test.js
```diff
@@ -0,0 +1,39 @@
...
+const clientCert = 'client-cert-data';
+const clientCertEncoded = encodeBase64URL(encodeBase64(clientCert));
+
+const clusterURL = 'https://0.0.0.0:46595';
+const clusterURLEncoded = encodeBase64URL(clusterURL);
+
+const clientKey = 'client-key';
+const clientKeyEncoded = encodeBase64URL(encodeBase64(clientKey));
+
+describe('Parse protocl headers with auth data', () => {
```
akucharska:
```suggestion
describe('Parse protocol headers with auth data', () => {
```

akucharska (dismissed):

backend/src/terminal/handler.js
```diff
@@ -0,0 +1,116 @@
...
+        logger.error({ err: event }, 'Front WebSocket error: ');
+      });
+
+      frontWS.addEventListener('message', (event) => {
+        if (k8sWS.readyState === WebSocket.OPEN) {
+          const data = event.data;
+          k8sWS.send(data);
+        } else {
+          logger.info(
+            'Front WS is not open, cannot send message, status: ' +
```
akucharska:
```suggestion
            'K8s WS is not open, cannot send message, status: ' +
```

backend/src/terminal/protocolHeaderParser.js
dbadura: It should be fixed

backend/src/terminal/handler.js
mrCherry97: yeah, exactly what you pasted @dbadura 
```
This method does nothing if type is not one of 'close', 'error', 'message', or 'open'.
```

backend/src/terminal/handler.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    let parsedHeaders;
+    try {
+      parsedHeaders = parseProtocolHeaders(
+        req.headers['sec-websocket-protocol'],
+      );
+    } catch (e) {
+      if (e instanceof InvalidInputError) {
+        socket.end('HTTP/1.1 400 Bad Request\r\n\r\n');
```
akucharska, line 49:
```suggestion
        socket.end('HTTP/1.1 400 Bad Request\r\n\r\n');
        return;
```

akucharska (commented):
LGTM

mrCherry97 (approved):

PR #5111, opened 2026-06-29
test: Add unit tests for configuration

### LINKED ISSUE #5067: Add unit tests for settings, configuration, and state utilities
**Description**
Several state management modules have 0% or very low coverage despite containing testable pure logic: settings atom reducers, configuration parsers, and shared state utilities. The affected directories are `src/state/settings/`, `src/state/configuration/`, `src/state/utils/`, and `src/state/discoverability/`.

**Reasons**
Settings and configuration state is read on every application boot and affects theming, feature flags, namespace defaults, and cluster connection parameters. Errors in this layer can silently corrupt the user's stored preferences or cause the app to fail loading entirely. These modules are Jotai atoms and plain reducers — they have no side effects and can be tested by directly invoking their logic with mock inputs. Establishing a test baseline here also creates a safety net for future settings additions, which are frequent as the project grows.

**Acceptance Criteria**
- [x] `src/state/settings/` statement coverage is higher than the current one
- [x] `src/state/configuration/` statement coverage is higher than the current one
- [x] `src/state/utils/` statement coverage is higher than the current one
- [x] `src/state/discoverability/` statement coverage is higher than the current one

**Testing Strategy (DoD)**
- Automated unit tests (Vitest) for atom reducers and configuration parsers
- Tests invoke state logic directly with mock inputs — no component mounting required

**Attachments**

**Related Issues**
#5071

KonradPietocha:
**Description**

- Tests for src/state/configuration/ have been added

**Related issue(s)**
Closes https://github.com/kyma-project/busola/issues/5067

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/state/configuration/test/configurationAtom.test.ts
+++ b/src/state/configuration/test/configurationAtom.test.ts
@@ -6,8 +6,14 @@ vi.mock('shared/utils/env', () => ({
 }));
 
 // Import after mocks are registered.
-const { configurationAtom } = await import('../configurationAtom');
-const { mergeWith, isArray } = await import('lodash');
+const { configurationAtom, getConfigs } = await import('../configurationAtom');
+
+const makeFetch = (responses: Record<string, string>) =>
+  vi.fn().mockImplementation((url: string) => {
+    const key = Object.keys(responses).find((k) => url.includes(k));
+    const body = key ? responses[key] : '';
+    return Promise.resolve({ text: () => Promise.resolve(body) });
+  });
 
 describe('configurationAtom', () => {
   let store: ReturnType<typeof createStore>;
@@ -28,7 +34,7 @@ describe('configurationAtom', () => {
     const config = {
       features: {
         EXTENSIBILITY: { isEnabled: true },
-        TERMINAL: { isEnabled: false },
+        SNOW: { isEnabled: false },
       },
     } as any;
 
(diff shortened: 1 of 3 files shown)


src/state/configuration/test/configurationAtom.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      }),
+    );
+
+    const result = await getConfigs(undefined);
+
+    expect(result?.storageType).toBe('sessionStorage');
+  });
+
+  it('continues without configmap when fetchFn throws', async () => {
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
```
mrCherry97, line 172: warnSpy.mockRestore() is the last line of each test. If any expect() before it throws, the spy is never restored — console.warn stays silenced for all subsequent tests. The correct pattern (used in useTerminalSession.test.ts) is afterEach(() => vi.restoreAllMocks()). Also: getFeatures.test.ts line ~119 sets up warnSpy but never calls mockRestore() at all.

src/state/configuration/test/configurationAtom.test.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    storeA.set(configurationAtom, { storageType: 'localStorage' });
+
+    expect(storeB.get(configurationAtom)).toEqual({});
+  });
+});
+
+describe('getConfigs', () => {
+  beforeEach(() => {
+    vi.stubGlobal('fetch', undefined);
```
mrCherry97, line 79: vi.stubGlobal('fetch', undefined) in beforeEach is never cleaned up with vi.unstubAllGlobals() in an afterEach. The stub stays registered in Vitest's registry after the describe block ends, leaving fetch corrupted for any subsequent test file running in the same worker. The beforeEach sets it to undefined but doesn't restore the original — those are different operations.

src/state/configuration/test/getFeatures.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    });
+  });
+
+  it('disables a feature whose check throws, leaving others intact', async () => {
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const failingCheck = vi.fn().mockRejectedValue(new Error('fail'));
+    const features = {
+      SENTRY: { isEnabled: true, checks: [failingCheck] },
+      FEEDBACK: { isEnabled: true },
+    } as any;
```
mrCherry97, line 130: In 'disables a feature whose check throws', warnSpy is created and mockRestore() is called, but expect(warnSpy).toHaveBeenCalled() is never asserted. The warning behavior is silently untested — a swallowed error would still pass.

src/state/configuration/test/configurationAtom.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  it('overwrites the previous value on each set', () => {
+    store.set(configurationAtom, { storageType: 'localStorage' });
+    store.set(configurationAtom, { storageType: 'sessionStorage' });
+    expect(store.get(configurationAtom)).toEqual({
+      storageType: 'sessionStorage',
+    });
+  });
+
+  it('notifies subscribers on change', () => {
+    const subscriber = vi.fn();
```
mrCherry97, line 58: store.sub() returns an unsubscribe function that is never called. The subscriber vi.fn() stays registered on the module-level atom and can fire in later tests, causing spurious mock-call counts.

src/state/configuration/test/configurationAtom.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  });
+
+  it('merges configmap values on top of defaultConfig and config.yaml', async () => {
+    vi.stubGlobal(
+      'fetch',
+      makeFetch({
+        'defaultConfig.yaml': 'config:\n  storageType: sessionStorage\n',
+        'config.yaml': '',
+      }),
+    );
```
mrCherry97, line 123: config.yaml is set to '' (empty), so the test only proves configmap > defaultConfig. The stated priority order defaultConfig < config.yaml < configmap is never exercised. A regression where configmap stops overriding a non-empty config.yaml would pass silently.

src/state/configuration/test/configurationAtom.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    warnSpy.mockRestore();
+  });
+
+  it('arrays from a later layer replace the base array entirely', async () => {
+    vi.stubGlobal(
+      'fetch',
+      makeFetch({
+        'defaultConfig.yaml':
+          'config:\n  features:\n    HIDDEN_NAMESPACES:\n      isEnabled: true\n      namespaces: [a, b, c]\n',
+        'config.yaml':
```
mrCherry97, line 194: The test only asserts namespaces: ['x']. If the array customizer accidentally dropped sibling fields (isEnabled: true), the test would still pass. Should assert the full merged shape.

src/state/configuration/test/configurationAtom.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  beforeEach(() => {
+    store = createStore();
+  });
+
+  it('has an empty object as the default value', () => {
+    expect(store.get(configurationAtom)).toEqual({});
+  });
+
+  it('is not null by default', () => {
+    expect(store.get(configurationAtom)).not.toBeNull();
```
mrCherry97: 'is not null by default' is fully subsumed by 'has an empty object as the default value' — anything satisfying toEqual({}) is already not null.
'can be set to null' tests Jotai's generic set/get round-trip, not any logic in configurationAtom.ts. Neither test can catch a real regression in this codebase.

src/state/configuration/test/configurationAtom.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    } as any;
+
+    store.set(configurationAtom, config);
+
+    expect(store.get(configurationAtom)).toEqual(config);
+  });
+
+  it('can be set to null', () => {
+    store.set(configurationAtom, null);
+    expect(store.get(configurationAtom)).toBeNull();
```
mrCherry97: 'is not null by default' is fully subsumed by 'has an empty object as the default value' — anything satisfying toEqual({}) is already not null.
'can be set to null' tests Jotai's generic set/get round-trip, not any logic in configurationAtom.ts. Neither test can catch a real regression in this codebase.

mrCherry97 (changes requested):

mrCherry97 (changes requested):
I was checking and everything was from my PoV to vote LGTM, but I double checked with Claude, and we discover interesting findings, please take a look:

Finding 1 — CONFIRMED, production bug gap
config.yaml network error returns null instead of falling back to defaultConfig
[configurationAtom.ts:50](vscode-webview://06sslph93p21sm41g6d7hceibfhfvh28a4di4qs0usu0dsc445lp/src/state/configuration/configurationAtom.ts#L50)

The fetch('/config/config.yaml') call sits inside the outer try but outside the inner try/catch that guards YAML parsing. A network error or DNS failure on config.yaml causes the entire function to return null, silently discarding the defaultConfig values that were already loaded. The existing "falls back when config.yaml contains invalid YAML" test only exercises the YAML parse error path (inner catch), not this path.

This is a production code issue, not just a test gap. The fix is to move the config.yaml fetch inside the inner try/catch at line 69, or wrap it in its own try/catch with the same configParams = {} fallback.

Finding 2 — CONFIRMED, missing test
No test for a check that resolves with { isEnabled: false }
[getFeatures.test.ts](vscode-webview://06sslph93p21sm41g6d7hceibfhfvh28a4di4qs0usu0dsc445lp/src/state/configuration/test/getFeatures.test.ts)

discoverFeature supports a meaningful contract: a check can disable a feature by returning { isEnabled: false }. The spread currentConfig = { ...currentConfig, ...newPartialConfig } would apply it. But no test exercises this path — the existing tests cover early-return (config starts with isEnabled: false), and check-throws. A regression removing isEnabled: false from the spread would go undetected.

mrCherry97 (approved):

PR #5113, opened 2026-06-30
feat: Disable Joule for non-Kyma IAS and EU Access Only clusters

### LINKED ISSUE #5063: feat: cluster region validation before enabling Joule
We need to disable Joule for clusters which are EU Access Only.

## Task(s):
- [x] If the --oidc-issuer-url of the SKR is not the Kyma default IAS, then disable the Joule feature.
- [x] If the joule is enabled, consume the [kyma-companion API](https://github.com/kyma-project/kyma-companion/issues/1222) to check if the cluster is EU Assess Only or not.
- [x] If the cluster is EU Assess Only, then disable the Joule feature, even if the kyma_companion and use_joule flags are set to true in the busola config
  ```
  KYMA_COMPANION:
            isEnabled: true
            config:
              feedbackLink: https://www.xxx.com/watch
              documentationLink: https://help.sap.com/docs/btp/sap-business-technology-platform-internal/kyma-companion?locale=en-US&version=Internal&state=DRAFT
              model: 'gpt-xx'
              queryMaxTokens: xxx
            useJoule: true
            jouleConfig:
              url: 'https://xxxxxx/resources/public/webclient/bootstrap.js'
              botname: 'kyma'
  ```

chriskari:
**Description**

- Disable Joule for non-Kyma IAS and EU Access Only clusters
- ...
- ...

**Related issue(s)**
#5063

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


src/hooks/useJouleEligibility.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function useJouleEligibility(): boolean {
+  const { isEnabled, useJoule, config } = useFeature<KymaCompanionFeature>(
+    configFeaturesNames.KYMA_COMPANION,
+  );
+  const cluster = useAtomValue(clusterAtom);
+  const serverUrl = cluster?.currentContext?.cluster?.cluster?.server;
+  const userExec = (cluster?.currentContext?.user?.user as KubeconfigOIDCAuth)
+    ?.exec;
+  const allowedIssuerUrl = config?.issuerUrl;
```
akucharska: `UseFeature` takes data also from `kube-public/busola-config`, so if user overrides joule, it will still be considered an allowedIssuerUrl

akucharska (changes requested):

jacekon:
@chriskari @akucharska, please also enable Joule, either as part of this PR or in a separate one, so that it will make it to prod by Tuesday. 
For context: On Monday, the (Green) Joule team will release Joule Kyma Capability, and our Companion will also be propagated from Stage to Prod. 
That leaves us with the last step of merging the above issue and enabling Joule in Busola on prod.

src/hooks/useJouleEligibility.ts
mrCherry97: Exactly, so the flow should be to check which active env is set, and the allowedIssuerUrl should be taken from the env config.

src/header/Header.tsx
```diff
@@ -154,7 +167,7 @@ export function Header() {
       >
         <SnowFeature />
         <FeedbackPopover />
-        {isKymaCompanionEnabled && isSAPUser && !isOnClustersPage && (
+        {showAssistant && (
```
akucharska, line 174: This will disable both joule and companion. Is it intended?

mrCherry97 (dismissed):

src/components/KymaCompanion/hooks/useJouleEligibility.ts
```diff
@@ -0,0 +1,102 @@
...
+      .catch((err) => {
+        if (err?.name === 'AbortError') return;
+        console.warn(
+          `[Joule] Could not determine eligibility (${err.message}); Joule will remain disabled.`,
+        );
+      });
+    return () => controller.abort();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
+    jouleConfigured,
```
akucharska: useJouleEligibility does not re-run when certificateAuthorityData changes for the same clusterUrl

chriskari: certificateAuthorityData isn't used by the eligibility check. The endpoint only uses clusterUrl for checking the region, the OIDC issuer, and the client cert/key for the cert-auth flag, and never reads the CA. So a CA change for the same clusterUrl would produce the identical answer, which is why it's intentionally out of the deps.

akucharska (approved):

PR #5122, opened 2026-07-03
fix: Prevent OIDC auth loop when SSO and cluster OIDC use different client IDs

mrCherry97:
**Description**

- Fix an infinite redirect/alert loop that occurred in non-incognito browser tabs when SSO (IAS) was enabled alongside cluster OIDC authentication. The root cause was both `UserManager` instances calling `signinRedirectCallback` on the same redirect URL, causing `oidc-client-ts` to throw `authority mismatch` or `client_id mismatch` because the stored state belonged to a different provider.
- Add `isOwnOidcCallback` helper in both `authDataAtom.ts` and `ssoDataAtom.tsx` that checks `sessionStorage['oidc.<stateId>'].client_id` against the current `UserManager`'s `clientId` before attempting to process the callback. This reliably identifies callback ownership without depending on the `iss` query param (which is environment-specific and not always present).
- Add a safety-net catch for `authority mismatch` errors in `authDataAtom.ts` alongside the existing `No state in response` handler — both now trigger `clearStaleState` + `signinRedirect` instead of showing a blocking `alert`.

**Related issue(s)**

N/A

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/kyma/environments/dev/config.yaml
+++ b/kyma/environments/dev/config.yaml
@@ -87,7 +87,7 @@ config:
     COMMUNITY_MODULES:
       isEnabled: true
     SSO_LOGIN:
-      isEnabled: false
+      isEnabled: true
       config:
         issuerUrl: https://apskyxzcl.accounts400.ondemand.com
         scope: openid
--- a/src/state/authDataAtom.ts
+++ b/src/state/authDataAtom.ts
@@ -55,6 +55,22 @@ function getToken(user: User | null, useAccessToken: boolean): string {
   throw new Error('id_token is empty');
 }
 
+// oidc-client-ts stores pending signin state as sessionStorage['oidc.<stateId>']
+// containing a JSON string with client_id and authority. Checking it tells us
+// whether the current callback URL belongs to this UserManager without relying
+// on the 'iss' query param, which varies across environments.
+function isOwnOidcCallback(clientId: string): boolean {
+  const stateId = new URLSearchParams(window.location.search).get('state');
+  if (!stateId) return false;
+  try {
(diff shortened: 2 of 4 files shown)


kyma/environments/dev/config.yaml
```diff
@@ -87,7 +87,7 @@ config:
     COMMUNITY_MODULES:
       isEnabled: true
     SSO_LOGIN:
-      isEnabled: false
+      isEnabled: true
```
KonradPietocha, line 90: I take it this is intentional, since we’ll want to test it on dev?

src/state/authDataAtom.ts
```diff
@@ -55,6 +55,22 @@ function getToken(user: User | null, useAccessToken: boolean): string {
   throw new Error('id_token is empty');
 }
 
+// oidc-client-ts stores pending signin state as sessionStorage['oidc.<stateId>']
```
KonradPietocha, line 58: `sessionStorage` remains in the comments in a few places, while `localStorage` is actually being used.

kyma/environments/dev/config.yaml
mrCherry97: yes, I want to test it on deployed version on dev.

src/state/authDataAtom.ts
mrCherry97: doesn't matter, we will make a follow up issue to make more corrections, but I don't want to wait for another round of the tests.

KonradPietocha (approved):

PR #5125, opened 2026-07-06
refactor: Extract ResourceRadialChart to a separate component

### LINKED ISSUE #3735: Refactor of Resources charts
**Description**
The logic responsible for displaying Radial chart with resources is getting more and more complicated because of unit formatting.
I propose to create component `ResourceRadialChart` which will deal with data format.
Additioanly it will remove duplication and improve readibility of our components.

**Reasons**
Currenlty the data format logic is implemented on normal view which uses directly UI5RadialChart.

**Attachments**
Example code responsible for displaying and formatting Chart:
```
        <UI5RadialChart
          color="var(--sapChart_OrderedColor_5)"
          value={
            cpusToHumanReadable(resources?.requests?.cpu, {
              unit: 'm',
            }).value
          }
          max={
            cpusToHumanReadable(cpu.capacity, {
              unit: 'm',
            }).value
          }
          additionalInfo={`${
            cpusToHumanReadable(resources?.requests?.cpu, {
              unit: 'm',
            }).string
          } / ${
            cpusToHumanReadable(cpu.capacity, {
              unit: 'm',
            }).string
          }`}
        />
```

awoloszyn22:
**Description**

- Extracted UI5RadialChart with complex display logic to a separate component (ResourceRadialChart) and adjusted its usages

**Related issue(s)**
Resolves #3735

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/shared/components/ResourceRadialChart/test/ResourceRadialChart.cy.jsx
+++ b/src/shared/components/ResourceRadialChart/test/ResourceRadialChart.cy.jsx
@@ -35,4 +35,55 @@ describe('ResourceRadialChart', () => {
     cy.contains('0%').should('exist');
     cy.contains('0m / 1000000m').should('exist');
   });
+
+  it('renders "50%" and "512Mi / 1024Mi" for bytes with unit Mi', () => {
+    cy.mount(
+      <ResourceRadialChart
+        value={512 * 2 ** 20}
+        max={1024 * 2 ** 20}
+        valueType="bytes"
+        unit="Mi"
+        titleText="Memory Usage"
+        accessibleName="Memory usage chart"
+      />,
+    );
+
+    cy.get('.radial-chart').should('exist');
+    cy.contains('50%').should('exist');
+    cy.contains('512Mi / 1024Mi').should('exist');
+  });
+
+  it('renders "0%" and "0Gi / 4Gi" for bytes with unit Gi when value is 0', () => {
+    cy.mount(
+      <ResourceRadialChart
+        value={0}
+        max={4 * 2 ** 30}
+        valueType="bytes"
+        unit="Gi"
+        titleText="Memory Usage"
+        accessibleName="Memory usage chart"
+      />,
+    );
+
+    cy.get('.radial-chart').should('exist');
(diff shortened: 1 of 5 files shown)


mrCherry97 (changes requested):
On cluster overview:

Image

And also I agree with a Claude:
- Wrong default unit for bytes. The unit prop defaults to 'm' (a CPU millicore unit), but ClusterStats.tsx uses valueType="bytes" without passing unit. This means bytesToHumanReadable(memory.usage, { unit: 'm' }) is called, which is semantically incorrect. NodeResources and ResourcesUsage both correctly pass unit="Mi" for memory usage. ClusterStats should too, or the default should be split per valueType.

- Repeated formatting calls. cpusToHumanReadable / bytesToHumanReadable is called 4 times per render (twice for value, twice for max) to build both chart values and the additionalInfo string. Could be called twice and results reused, but this is a minor efficiency concern.

- Implicit bytes fallback. When valueType is undefined, the component silently falls through to bytesToHumanReadable. This isn't documented in the type or prop description, making the API less obvious.

- No tests. The issue doesn't require tests explicitly, but the new formatting logic in ResourceRadialChart is a good candidate for a unit test (especially edge cases like value=0 or max=0).

awoloszyn22:
Follow-up issue for fixing the console warning: https://github.com/kyma-project/busola/issues/5139

src/shared/components/ResourceRadialChart/test/ResourceRadialChart.cy.jsx
```diff
@@ -0,0 +1,38 @@
+/* global cy, describe, it */
+import { ResourceRadialChart } from '../ResourceRadialChart';
+
+describe('ResourceRadialChart', () => {
```
mrCherry97, line 4: Testing the CPU valueType doesn't make sense from my pov. Please add more cases with bytes and different units, maybe.

mrCherry97 (changes requested):

mrCherry97 (approved):

PR #5134, opened 2026-07-08
fix: Silent OIDC renewal and preserve location on session drop

### LINKED ISSUE #5101: Session expires after a few minutes of inactivity, requiring manual re-login and losing navigation context
**Description**

Users are frequently interrupted by a session timeout error — "Your session could not be renewed. Refresh token is invalid or expired." — after only a few minutes of inactivity (reported as ~3–5 minutes). After dismissing the dialog and re-selecting a cluster, the session resumes without requiring credentials, but the user's previous location in the UI is lost. 

Additionally, if the user tries to navigate immediately after dismissing the error, the screen flashes and they are thrown back to the cluster selection screen repeatedly. They must wait 1–2 minutes before the session stabilizes.

Reported across Windows 10, Windows 11, and macOS. Confirmed by multiple users including students, suggesting it is not environment-specific.

In addition, there is a comment related to this in the UX deep dive dashboard: 

> It is probably one of the worst experiences I have every had with a platform. I get logged out every 10 minutes and logging back is is not possible unless I restart my browser.

**Expected result**

Whatever the session timeout policy requires, renewal should happen silently in the background without interrupting the user. Only if silent renewal genuinely fails should an error be shown — and even then, the user should be returned to their previous location after re-authenticating.

**Actual result**

- Error dialog appears: "Your session could not be renewed. Refresh token is invalid or expired."
- User must dismiss the dialog and manually re-select their cluster.
- Previous navigation context is lost.
- Immediately retrying after dismissal causes a flash and redirect loop; the user must wait ~1–2 minutes before the session is usable again.

**Steps to reproduce**

1. Log in to Kyma Dashboard and navigate to any page
2. Leave the dashboard open but inactive for approximately 3–5 minutes
3. Attempt to interact with the UI

**Troubleshooting**

Reported by a user who experiences this daily across multiple machines and operating systems. Also observed by students using macOS. Issue has been present for several years according to the user.

**Screenshot:**

Image

chriskari:
**Description**

- Fix the silent token renewal so it no longer races itself and fails with "Refresh token is invalid or expired" after a few minutes of inactivity.
- On a real session expiry, save the user's current URL and redirect them through the IdP so they land back on the same page — no more forced trip to the cluster list.
- Remove the 2-second page reload in `ClusterRoutes` and gate the openapi error recovery on an in-flight renewal, which together eliminated the flash + redirect loop.

**Related issue(s)**
#5101

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


chriskari:
Small info: the lighthouse check is red, but it's not due to my PR actually lowering the accessibility, but the accessibility score on main is already below the threshold, and until now the check was only passing by luck. My PR makes the page load more cleanly, which ironically means that some UI5 components from the sidebar with internal accessibility violations, now load faster and lighthouse now catches it every time. So this cannot really be fixed in my PR. We'll either need to raise this issue with UI5 or just need to adjust the threshold slightly.

awoloszyn22 (commented):
A few issues with tests:
1) usePreviousPathTracker.test.ts
The test name says "does not overwrite storage when path is /clusters", but savePreviousPath doesn't have a /clusters guard — the guard is in useAfterInitHook (the restore logic). savePreviousPath would in fact write /clusters to storage. Checking: the tracker calls savePreviousPath(location.pathname, location.search), which calls localStorage.setItem(PREVIOUS_PATHNAME_KEY, '/clusters'). So the test is likely wrong — it expects the old value to be preserved, but the actual code will overwrite it with /clusters.
This should be verified and either the test fixed or the guard added to savePreviousPath. 

Also: the second test reaches into localStorage directly (localStorage.getItem(PREVIOUS_PATHNAME_KEY)) while the first uses getPreviousPath(). Pick one approach for consistency.

2) useAfterInitHookRestoreScope.test.ts
 Top-level `await import (…)` after vi.mock - the dynamic imports after vi.mock calls are an anti-pattern that exists only to work around Vitest's module hoisting. This is fragile and confusing to future editors. Standard import at the top of the file, with the mocks declared before them (using vi.hoisted if needed), is clearer.

chriskari:
Fixed both style points in the tests (consistent getPreviousPath(), static imports).

On the `/clusters `guard though: it's actually in `savePreviousPath` itself, not in the restore logic.
See `useAfterInitHook.ts:47-49` (was already there before this PR). So `/clusters` never gets written and the test is right as-is.

awoloszyn22:
LGTM

mrCherry97 (dismissed):

KonradPietocha (approved):

PR #5173, opened 2026-07-23
fix: show terminal button label on narrow viewports

Thanvik931:
Updated the header's terminal `ToggleButton` to render the localized `terminal.name label when `isLargeScreen` is false. Ensures the text label remains visible alongside the icon on small viewports so the button's action is clear.


dbadura:
Closing in favor of: https://github.com/kyma-project/busola/pull/5160

PR #5183, opened 2026-07-24
fix: Command palette navigation opens wrong column layout

### LINKED ISSUE #5114: Navigating to Namespace list with Command Palette is not working correctly
**Description**
As a user, when I'm on resource details, I want to switch the namespace, but I'm redirected to the column view, which isn't what I expect. 

**Expected result**
After going to another namespace in the command palette, I should land on the namespace details or switch context instead of the namespace details, but with a list view in column layout. 

**Actual result**

Image

**Steps to reproduce**
1. Go to some namespace
2. Go to some resource 
3. Go to the details
4. Click `cmd +k` 
5. Type `ns -> Tab -> NAMESPACE_NAME`
6. Click Enter

**Troubleshooting**

chriskari:
**Description**

- Namespace navigation opens single-column details, not list + details.
- ModuleTemplates (and other CR types) open two-column, no empty third column.
- Layout now derived from the target route instead of a hard-coded ?layout= param.

**Related issue(s)**
#5114

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


mrCherry97 (changes requested):
If you are adding tests, then please think about more cases than only those listed in the issue. In my opinion it will be good to check not only custom resources, but also namespaces, list, details, and some extensibility

mrCherry97 (approved):

PR #5192, opened 2026-07-29
feat: Sync Joule theme with Busola user theme selection

tanweersalah:
**Description**

- Sync the Joule Web Client theme with the user's active Busola theme on every theme change, covering all five options: Light, Dark, High-Contrast White, High-Contrast Black, and system-adaptive Light/Dark.
- Apply the correct theme on initial Joule script load by storing the resolved theme in a ref and calling `setTheme` inside `script.onload`, so the first render uses the user's selected theme rather than Joule's default.
- Listen for OS-level `prefers-color-scheme` changes when the "Automatically matches your system settings" option is active, so Joule updates without requiring a page reload.

**Related issue(s)**

N/A

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/components/KymaCompanion/JouleChat.jsx
+++ b/src/components/KymaCompanion/JouleChat.jsx
@@ -34,7 +34,7 @@ export default function JouleChat() {
   const resourceRef = useRef(currentResource);
   const clusterRef = useRef(cluster);
   const authDataRef = useRef(authData);
-  const jouleThemeRef = useRef(null);
+  const resolveJouleThemeRef = useRef(null);
 
   useEffect(() => {
     resourceRef.current = currentResource;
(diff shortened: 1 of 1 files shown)


src/components/KymaCompanion/JouleChat.jsx
```diff
@@ -32,6 +34,7 @@ export default function JouleChat() {
   const resourceRef = useRef(currentResource);
   const clusterRef = useRef(cluster);
   const authDataRef = useRef(authData);
+  const jouleThemeRef = useRef(null);
```
OliwiaGowor: I would suggest renaming it to something like `resolveJouleThemeRef ` because a function is stored in this ref, not the theme string.

tanweersalah: updated

OliwiaGowor (approved):

PR #5242, opened 2026-08-12
fix(overview): Keep dismissed banners hidden after a reload

### LINKED ISSUE #5102: Cluster overview banner reappears after being dismissed — check local storage persistence
**Description**

The banners on the cluster overview page (e.g. "Meet Joule") can reappear after being dismissed. Dismissing a banner is supposed to store that action in local storage so it does not show again. The persistence logic may not be working correctly.

**Expected result**

Once a user dismisses a banner, it should not reappear on subsequent visits. The dismissed state should be correctly written to and read from local storage.

**Actual result**

According to a user report, the banner reappears after being dismissed. We have not been able to reproduce this independently, but the user confirmed it happens consistently.

**Steps to reproduce**

1. Open Kyma Dashboard and navigate to the cluster overview page
2. Dismiss one of the onboarding banners using the × button
3. Reload the page or return to the cluster overview later

**Troubleshooting**

Reported by a user. Please verify that the logic is working correctly.

Image

MFA-G:
**Description**

- Initialise `FeatureCardBanner`'s `hideBanner` state lazily from local storage, so a dismissed banner stays hidden after a reload.
- Add unit tests for the dismiss/persist behaviour of `FeatureCardBanner`.

`FeatureCardBanner` only read `hideBanner<id>` from local storage inside the `id !== prevId` branch:

```ts
const [hideBanner, setHideBanner] = useState(false);
const [prevId, setPrevId] = useState(id);

if (id !== prevId) {
  setPrevId(id);
  const storedHideValue = localStorage.getItem(hideBannerKey);
  if (storedHideValue !== null) setHideBanner(storedHideValue === 'true');
}
```

That branch never runs on the first render — `prevId` is initialised to `id`. So on every fresh mount the component starts with `hideBanner = false` and renders the banner, even though `'true'` was written under `hideBanner<id>` when the user closed it. Because the banners on the cluster overview (`AIBanner`, `KymaCLIBanner`) mount with a fixed `id`, the id-change path never fires for them at all and the stored value was effectively never read — the banner comes back on every page load.

The fix initialises the state from local storage directly:

```ts
const isBannerHidden = (hideBannerKey: string): boolean =>
  localStorage.getItem(hideBannerKey) === 'true';

const [hideBanner, setHideBanner] = useState(() => isBannerHidden(hideBannerKey));
```

The id-change branch keeps working and now shares the same helper. Writing the dismissal (`localStorage.setItem(hideBannerKey, 'true')`) is unchanged.

**Tests**

Added `src/shared/components/FeatureCard/FeatureCard.test.tsx` with four cases: renders when never dismissed, persists the dismissal to local storage on close, stays hidden on the next mount, and keeps the dismissed state scoped per banner id.

```
npx vitest run src/shared/components/FeatureCard/FeatureCard.test.tsx
Test Files  1 passed (1)
     Tests  4 passed (4)
```

Against the previous implementation, `stays hidden on the next mount after being dismissed` fails, so it pins the regression.

`npx eslint src/shared/components/FeatureCard/`, `npx prettier --check` on the touched files, and `npx tsc --noEmit` (no new errors in the touched files) are clean.

**Related issue(s)**

Closes #5102

### DIFF
--- a/src/shared/components/FeatureCard/FeatureCard.test.tsx
+++ b/src/shared/components/FeatureCard/FeatureCard.test.tsx
@@ -1,0 +1,64 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent } from '@testing-library/react';
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key: string) => key,
+    i18n: { changeLanguage: () => new Promise(() => {}) },
+  }),
+}));
+
+vi.mock('jotai', () => ({
+  useAtomValue: () => 'sap_horizon',
+}));
+
+import { FeatureCardBanner } from './FeatureCard';
+
+const renderBanner = (id = 'test-banner') =>
+  render(
+    <FeatureCardBanner
+      id={id}
+      title="Banner title"
+      description="Banner description"
+      design="information-1"
+    />,
+  );
+
+describe('FeatureCardBanner', () => {
+  beforeEach(() => {
+    localStorage.clear();
+  });
+
+  it('renders the banner when it was never dismissed', () => {
+    renderBanner();
+
+    expect(screen.getByText('Banner title')).toBeInTheDocument();
+  });
+
+  it('stores the dismissal in local storage when closing the banner', () => {
+    const { container } = renderBanner();
+
+    const closeButton = container.querySelector('.decline-button');
+    expect(closeButton).not.toBeNull();
+    fireEvent.click(closeButton!);
+
+    expect(screen.queryByText('Banner title')).not.toBeInTheDocument();
+    expect(localStorage.getItem('hideBannertest-banner')).toEqual('true');
+  });
+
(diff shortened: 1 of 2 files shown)


chriskari (approved):

chriskari:
Thanks for the contribution. The changes look good and the PR is approved from our side.
Before we can merge, you'll need to sign the Contributor License Agreement. You can do so by clicking the link in the CLA bot's comment above.

mrCherry97:
@MFA-G Thanks for the contribution. The changes look good and the PR is approved from our side.
Before we can merge, you'll need to sign the Contributor License Agreement. You can do so by clicking the link in the CLA bot's comment above.

KonradPietocha:
@MFA-G Thanks for the contribution. In order to merge the changes, you must sign the Contributor License Agreement. You can do so by clicking the link in the cla-assistant bot's comment above.

mrCherry97:
In case of not signing the CLA licensing, we need to close a PR. Thanks for the contribution!

PR #5250, opened 2026-08-18
fix: All OIDC providers defined in the OIDC list are available

KonradPietocha:
**Description**

- Making non-interactive user OIDC providers unavailable.

**Related issue(s)**
Closes https://github.tools.sap/kyma/backlog/issues/7693 

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


OliwiaGowor (commented):
- [x] SSRF — unauthenticated outbound proxy
The `/backend/oidc-discovery` endpoint is registered before `requireK8sCredential`, so it requires no authentication. Anyone who can reach the backend can supply an arbitrary `issuer` URL and use it to probe internal network services (e.g. cluster-internal hostnames, cloud metadata endpoints reachable over HTTPS, other pods in the same namespace).

The `https:` protocol check blocks plain HTTP but does not prevent requests to internal hosts. Suggested mitigations (pick one or combine):

- Resolve the hostname and reject requests to private/loopback IP ranges (RFC 1918 + `127.x`, `::1`, `169.254.x.x`, etc.) before making the fetch.
- Restrict the endpoint to authenticated users only — since it is called during the cluster-add wizard, it could require a lightweight check (e.g. a signed nonce or session cookie) rather than full K8s credentials.

- [x] No rate limiting on the discovery endpoint
Because the endpoint is unauthenticated and triggers an outbound HTTP request, it can be abused in a tight loop to exhaust backend connections or to amplify traffic toward a third-party host. The existing `k8sRateLimiter` does not cover this route.

Suggested fix: apply a per-IP rate limiter (e.g. `express-rate-limit`) specifically on this route

backend/modules/oidcDiscoveryRouter.js
```diff
@@ -0,0 +1,40 @@
...
+  } catch {
+    return res.status(400).json({ error: 'Invalid issuer URL' });
+  }
+
+  if (url.protocol !== 'https:') {
+    return res.status(400).json({ error: 'Issuer URL must use HTTPS' });
+  }
+
+  try {
+    const base = issuer.replace(/\/$/, '');
```
OliwiaGowor: We can omit the regex I think :D 

```suggestion
    const base = issuer.endsWith('/') ? issuer.slice(0, -1) : issuer;
```

src/components/Clusters/components/test/oidc-interactive-check.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    const { result } = renderHook(() =>
+      useNonInteractiveOidcContexts(contexts, users as any),
+    );
+    await waitFor(() => expect(result.current.has('gha-ctx')).toBe(true));
+  });
+
+  it('resolves the user via context.context.user, not by context name', async () => {
+    // context name is 'my-cluster', but the user ref is 'oidc-user'
+    // user 'my-cluster' has a token (no OIDC exec) — old buggy code would find this and skip
+    // user 'oidc-user' has a non-interactive OIDC issuer — new code correctly finds this
```
OliwiaGowor: Are those necessary for clarity?

OliwiaGowor (approved):

grego952 (approved):

PR #5252, opened 2026-08-19
test: Add unit tests for shared resource hooks

### LINKED ISSUE #5072: Improve unit test coverage for BackendAPI and shared hooks
**Description**
`src/shared/hooks/BackendAPI/` sits at only 39% coverage — this module is the backbone of all Kubernetes API calls in Busola. Error handling paths, polling logic, and cache invalidation are largely untested. The broader `src/shared/hooks/` directory is at 7% with many individual hooks at 0%.

**Reasons**
Every single resource list, detail view, and form in Busola ultimately fetches data through `BackendAPI`. Untested error handling means that network failures, permission errors (403), or missing resources (404) may not surface correctly in the UI, leading to silent failures or broken states that are hard to diagnose. Polling and cache invalidation logic is particularly risky — regressions here can cause stale data to display or excessive requests to the Kubernetes API server. These hooks are also the most frequently touched layer when adding new resource support, so a strong test baseline prevents regressions from accumulating over time.

**Acceptance Criteria**

- [x] `src/shared/hooks/BackendAPI/` statement coverage is higher than the current one
- [x] Key hooks in `src/shared/hooks/` with >20 statements currently at 0% have baseline coverage
- [x] Error handling branches (network errors, 403, 404 responses) are explicitly tested

**Testing Strategy (DoD)**

- Automated unit tests (Vitest) using `renderHook` and mocked `fetch`/backend responses
- Error path tests must cover at minimum: network failure, 403 Forbidden, 404 Not Found

**Attachments**

**Related Issues**
#5071

chriskari:
Test useProtectedResources: rule matching (exact + regex), isProtected vs isProtectedResource (respects user disable setting), disabled feature. Test useSidecar: initial state, label write on toggle, no-write guard. Test useGetHiddenNamespaces: enabled/disabled/invalid config. Test useIsInCurrentNamespace: in/out of namespace, all-namespaces, non-namespaced.

**Description**

- Added unit tests for shared resource hooks
- ...
- ...

**Related issue(s)**
#5072

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


OliwiaGowor (changes requested):
There are many errors in `src/shared/hooks/tests/useGetHiddenNamespaces.test.ts` and `src/shared/hooks/tests/useIsInCurrentNamespace.test.ts`

src/shared/hooks/tests/useGetHiddenNamespaces.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    expect(result.current).toEqual([]);
+  });
+
+  it('returns an empty array when namespaces config is not an array', () => {
+    featureValue = { isEnabled: true, config: { namespaces: 'not-an-array' } };
+
+    const { result } = renderHook(() => useGetHiddenNamespaces());
+
+    expect(result.current).toEqual([]);
+  });
```
OliwiaGowor, line 43: Missing edge case: isEnabled: true, config: undefined
`featureValue = { isEnabled: true };  // no config key at all`

The source evaluates `Array.isArray(hiddenNamespacesConfig?.config?.namespaces) `— this is false when config is undefined, and the function returns []. The test "returns an empty array when namespaces config is not an array" `covers config: { namespaces: 'not-an-array' }`

src/shared/hooks/tests/useSidecarInjection.test.js
```diff
@@ -0,0 +1,107 @@
...
+    expect(result.current.isSidecarEnabled).toBe(true);
+
+    // label removed directly in the YAML -> the reactive effect flips the toggle off
+    act(() => {
+      rerender({ ...props, res: { metadata: { labels: {} } } });
+    });
+
+    expect(result.current.isSidecarEnabled).toBe(false);
+  });
+});
```
OliwiaGowor, line 118: Missing edge case: null/undefined initialRes

The source guards with `jp.value(initialRes || {}, ...)`. There's no test for initialRes: null.

src/shared/hooks/tests/useSidecarInjection.test.js
```diff
@@ -0,0 +1,107 @@
...
+      }),
+    );
+  });
+
+  it('writes the disabled label when toggled off', () => {
+    const res = { metadata: { labels: { [label]: 'enabled' } } };
+    const { result, setRes } = setup({
+      res,
+      initialRes: { metadata: { labels: { [label]: 'enabled' } } },
+    });
```
OliwiaGowor, line 69: Redundant mutation assertion 

```
expect(res.metadata.labels[label]).toBe('enabled');
expect(setRes).toHaveBeenCalledWith(expect.objectContaining({ ... }));

```
The first assertion tests a jsonpath implementation detail: that jp.value() mutates the object in-place before setRes is called. The setRes mock call assertion already covers the hook's contract. The first assertion could silently break if the underlying jsonpath behavior changes, without indicating a real regression.

src/shared/hooks/tests/useSidecarInjection.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+    expect(res.metadata.labels[label]).toBe('enabled');
+    expect(setRes).toHaveBeenCalledWith(
+      expect.objectContaining({
+        metadata: expect.objectContaining({
+          labels: expect.objectContaining({ [label]: 'enabled' }),
+        }),
+      }),
+    );
+  });
```
OliwiaGowor, line 62: Silent ordering constraint in toggle tests

```
act(() => {
  result.current.setIsChanged(true);
  result.current.setSidecarEnabled(true);
});
```

isChanged is intentionally excluded from the first useEffect's dependency array. This means calling setSidecarEnabled(true) before setIsChanged(true) in a separate act() would not write the label — the effect would have fired while isChanged was still false. The tests happen to batch both updates together, which works, but this constraint is undocumented. A brief comment or a separate negative test would make the intended API clearer.

OliwiaGowor (approved):

PR #5259, opened 2026-08-20
chore: Add keep-alive agent for outbound proxy requests

mrCherry97:
**Description**

- Added a shared `proxyAgent` (`https.Agent` with `keepAlive: true`) in `https-agent.js` for the `/proxy` route
- Wired `proxyAgent` into `proxy.js` so outbound HTTPS requests reuse TCP connections instead of opening a new one per request

Previously, every request through `/proxy` paid a full TCP connection + TLS handshake (~100–200ms overhead) because no keep-alive agent was configured. The K8s handler already had a shared agent (`tokenAuthAgent`); this change applies the same pattern to the proxy route.

**Related issue(s)**

N/A

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/backend/utils/https-agent.js
+++ b/backend/utils/https-agent.js
@@ -14,8 +14,8 @@ import https from 'https';
 export const tokenAuthAgent = new https.Agent({
   keepAlive: true,
   keepAliveMsecs: 1000, // Detect dead connections quickly
-  maxSockets: 100,
-  maxFreeSockets: 20,
+  maxSockets: 200,
+  maxFreeSockets: 50,
   timeout: 60000, // Match typical load balancer timeouts
   scheduling: 'lifo',
 });
--- a/backend/utils/https-agent.js
+++ b/backend/utils/https-agent.js
@@ -24,9 +24,8 @@ export const tokenAuthAgent = new https.Agent({
 // Without this, every proxied request pays a full TCP+TLS handshake (~100-200ms).
 export const proxyAgent = new https.Agent({
   keepAlive: true,
-  keepAliveMsecs: 1000,
   maxSockets: 50,
   maxFreeSockets: 10,
   timeout: 30000,
-  scheduling: 'lifo',
+  scheduling: 'fifo',
 });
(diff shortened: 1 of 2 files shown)


awoloszyn22 (dismissed):

backend/utils/https-agent.js
```diff
@@ -14,8 +14,19 @@ import https from 'https';
...
+
+// Shared keep-alive agent for outbound proxy requests (/proxy route).
+// Without this, every proxied request pays a full TCP+TLS handshake (~100-200ms).
+export const proxyAgent = new https.Agent({
+  keepAlive: true,
+  keepAliveMsecs: 1000,
+  maxSockets: 50,
+  maxFreeSockets: 10,
+  timeout: 30000,
+  scheduling: 'lifo',
```
dbadura: According to docs:
```
selects the most recently used socket, while 'fifo' selects the least recently used socket. In case of a low rate of request per second, the 'lifo' scheduling will lower the risk of picking a socket that might have been closed by the server due to inactivity. In case of a high rate of request per second, the 'fifo' scheduling will maximize the number of open sockets, while the 'lifo' scheduling will keep it as low as possible. Default: 'lifo'.
```
I would choose `fifo`.

backend/utils/https-agent.js
```diff
@@ -14,8 +14,19 @@ import https from 'https';
...
+  maxFreeSockets: 50,
   timeout: 60000, // Match typical load balancer timeouts
   scheduling: 'lifo',
 });
+
+// Shared keep-alive agent for outbound proxy requests (/proxy route).
+// Without this, every proxied request pays a full TCP+TLS handshake (~100-200ms).
+export const proxyAgent = new https.Agent({
+  keepAlive: true,
+  keepAliveMsecs: 1000,
```
dbadura: It can be removed, default value is `1000`

dbadura (approved):

PR #5270, opened 2026-08-26
fix: Use custom dns lookup function

dbadura:
**Description**

- Use custom DNS lookup function to connect to not trusted URL provided to block all local calls

**Related issue(s)**

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/backend/utils/network-utils.js
+++ b/backend/utils/network-utils.js
@@ -68,15 +68,12 @@ async function isPrivateAddressCached(hostname) {
   try {
     const addresses = await dns.lookup(hostname, { all: true });
     for (const addr of addresses) {
+      ipAddress = addr.address;
+      familyAddress = addr.family;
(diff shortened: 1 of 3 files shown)


backend/utils/network-utils.js
```diff
@@ -62,6 +73,11 @@ export async function isPrivateAddressCached(hostname) {
         break;
       }
     }
+    // Connect to the first resolved address we validated above.
+    if (addresses.length > 0) {
```
KonradPietocha: Should we also check here whether is private?
```suggestion
    if (!isPrivate && addresses.length > 0) {
```

awoloszyn22 (approved):

KonradPietocha (approved):

PR #5272, opened 2026-08-28
fix: The terminal remains open when the user switches clusters

### LINKED ISSUE #5137: The terminal is still opened when user leave the dashboard
**Description**
When user want change the cluster having terminal opened, the old terminal is visible and opened.

**Expected result**
The terminal should be closed and connection terminated.

**Actual result**

Image

**Steps to reproduce**
- Connect to 2 clusters
- On first cluster, open terminal and wait for loading
- Go to cluster overview (the terminal will be accessible)
- Go to other cluster (the terminal will be accessible)
**Troubleshooting**

KonradPietocha:
**Description**

- Closing the terminal when the cluster is being changed.
- A test for the BusolaTerminal during cluster switching has been added.

**Related issue(s)**
Closes #5137 

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


OliwiaGowor (commented):
The current test _"does not close when the cluster stays the same"_ is designed to set the atom to a cluster object with the same name, rather than actually switching away and then back again. Here's an unverified scenario to consider:

1. Open a terminal on cluster-a.
2. Switch to cluster-b → the terminal closes.
3. The terminal reopens on cluster-b (new mount, openedOnClusterRef = "cluster-b").
4. Switch back to cluster-a → the terminal should close.

src/components/App/BusolaTerminal/BusolaTerminal.tsx
```diff
@@ -29,6 +30,14 @@ export function BusolaTerminal({
...
   const theme = useAtomValue(themeAtom);
   const { connect, disconnect } = useTerminalSession();
+  const cluster = useAtomValue(clusterAtom);
+  const openedOnClusterRef = useRef(cluster?.name);
+
+  useEffect(() => {
+    if (cluster?.name !== openedOnClusterRef.current) {
+      setShowTerminal((prev) => ({ ...prev, isOpen: false }));
+    }
+  }, [cluster?.name, setShowTerminal]);
```
OliwiaGowor: `disconnect()` should also be called here, like in the `handleClose,` to not depend entirely on the component unmount. The effect should mirror `handleClose`:

```suggestion
  useEffect(() => {
  if (cluster?.name !== openedOnClusterRef.current) {
    openedOnClusterRef.current = cluster?.name;
    disconnect(podNameRef.current);                         // explicit, not relying on unmount
    setShowTerminal((prev) => ({ ...prev, isOpen: false }));
  }
}, [cluster?.name, setShowTerminal, disconnect]);

```

OliwiaGowor (approved):

OliwiaGowor (changes requested):
The terminal stays open when I go to the clusters list view - shouldn't it also be closed here?

OliwiaGowor (approved):

PR #5274, opened 2026-09-01
fix: Logs fetching and logs panel scrolling

### LINKED ISSUE #5104: Improve log viewing experience — scroll behavior, log completeness, and settings persistence
**Description**

Log viewing is a primary use case for at least some of our users, but the current experience has several pain points identified during a user interview. This issue captures the full context and suggests improvements.

**Current problems**

1. **Constant full-page refresh disrupts scroll position**
The log view refreshes frequently to fetch new logs. When a user is scrolling through older logs, the refresh jumps them back to the top. The workaround is to enable "Reverse logs", but that inverts the chronological order, which adds cognitive overhead.

2. **Not all logs are shown for the selected timeframe**
When a user selects a timeframe (e.g. 6 hours, 3 hours), the logs shown do not always match. The same logs are visible via `kubectl logs` from the command line, so the data exists and this is where the user goes to see it completely.

3. **Log settings are not persisted across page refreshes**
The timeframe and "Show Timestamps" settings reset every time the log view refreshes. Since the view refreshes frequently to fetch new logs, users have to re-apply their preferences repeatedly within the same session.

**Some possible workarounds**

- On refresh, preserve scroll position or auto-scroll to the bottom if the user was already at the bottom (standard "tail" behavior). Only reload the log pane (if possible separate it from the controls), not the full page.
- Investigate why the selected timeframe is not always returning the expected log volume and align with what `kubectl logs` returns.
- Persist log view settings (timeframe, show timestamps, reverse logs) so they survive page refreshes.

**Future consideration**

The user expressed openness to viewing logs in a terminal if that would provide a better experience. Once the web-based terminal feature is available, it may be worth exploring whether logs can be opened there as an alternative. For example, through a button next to the "view logs" one.

**User context**

Reported by a daily Kyma Dashboard user. Logs are the primary reason they use the dashboard. The issues have been present for a long time and are experienced across operating systems and machines.

awoloszyn22:
**Description**

- Fixed wrong behaviour (jumping to top) of logs panel during view refreshing
- Fixed lack of alignment between logs shown and accessed by `kubectl logs` by keeping `tailLines` param only for 'All' timeframe 
- The issue with log settings persistence was not reproducible

**Related issue(s)**
Resolves #5104

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


tsconfig.json
mrCherry97, line 17: Is this change necessary?

src/resources/Pods/ContainersLogs.tsx
```diff
@@ -157,23 +237,26 @@ const ContainersLogs = ({
...
               >
                 {t('pods.labels.save-to-file')}
               </Button>
               <SearchInput
-                disabled={!logsToSave?.length}
+                disabled={!displayData.length}
                 entriesKind={'Logs'}
+                filteredEntries={[]}
+                suggestionProperties={[]}
+                allowSlashShortcut={false}
```
mrCherry97, line 238: Is this change necessary?

mrCherry97 (changes requested):
I think the logic is really complicated, and the root cause of the complexity is that all scroll control is imperative (el.scrollTop = ...) and coordinated through refs rather than being derived from a single source of truth. A cleaner model would be:
- One scrollBehavior state: 'tail' | 'free' | 'snap-to-top'
- One place that reads it and applies scroll (a single effect on displayData)
- The scroll listener, timeframe change, and reverse toggle all just set that state
That would replace the 4 refs + 5 effects with 1 state + 2 effects, and make the state machine explicit and testable.

src/resources/Pods/ContainersLogs.tsx
awoloszyn22: It fixes TS `missing props` error in `SearchInput`

tsconfig.json
awoloszyn22: It fixes `react-router` import errors in `DynamicPageComponent.tsx`

mrCherry97 (dismissed):

mrCherry97 (approved):

PR #5276, opened 2026-09-02
fix: Wait for Terminating pod before provisioning a new terminal session

### LINKED ISSUE #5229: Busola cannot connect to backend terminal pod
**Description**
Sometimes when user turn on/off terminal too fast, the terminal pod is in `Termination` state and the connection is not possible. The error is displayed to the user instead of working terminal.

**Expected result**
The terminal should be resistant to lifecycle of terminal pod

**Actual result**
Sometimes the terminal turns off with pre-terminated connection:

Image

**Steps to reproduce**
Turn on and off the terminal in random part of time. 
You can watch the terminal status:
```
watch -n 1 kubectl get pod -n busola-terminal
```

**Troubleshooting**

mrCherry97:
**Description**

- Added `waitIfTerminating()` helper in `provisionPod.ts` that checks whether an existing terminal pod has a `deletionTimestamp` (i.e. is Terminating) before attempting to create a new one
- If a Terminating pod is detected, the helper polls until the pod is fully deleted (404) before proceeding with pod creation, preventing a race condition where the attach WebSocket connects to a dying pod
- Shared the 120 s deadline across the new wait step and `pollPodReady` so the overall timeout budget is respected
- Added three new unit tests covering: waiting for a Terminating pod to be deleted, fast-path when no Terminating pod is present, and fast-path when no pod exists yet

**Related issue(s)**

Resolves #5229

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/components/App/BusolaTerminal/provisionPod.ts
+++ b/src/components/App/BusolaTerminal/provisionPod.ts
@@ -72,12 +72,49 @@ async function createIfMissing(
   }
 }
 
+// Waits until a pod that is currently Terminating (has deletionTimestamp) is
+// fully deleted. If the pod does not exist or is not terminating, returns
+// immediately. The caller should re-create the pod afterwards.
+async function waitIfTerminating(
+  fetchFn: FetchFn,
+  podName: string,
+  abortController: AbortController,
+  deadline: number,
+): Promise<void> {
+  const podUrl = `/api/v1/namespaces/${TERMINAL_NAMESPACE}/pods/${podName}`;
+  let isTerminating: boolean;
+  try {
+    const res = await fetchFn({ relativeUrl: podUrl, abortController });
+    const pod = await res.json();
+    isTerminating = Boolean(pod?.metadata?.deletionTimestamp);
+  } catch (err) {
+    if (err instanceof HttpError && err.code === 404) return;
+    throw err;
+  }
+  if (!isTerminating) return;
+
+  while (Date.now() < deadline) {
+    if (abortController.signal.aborted)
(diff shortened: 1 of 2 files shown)


src/components/App/BusolaTerminal/provisionPod.ts
```diff
@@ -72,12 +72,49 @@ async function createIfMissing(
...
+  } catch (err) {
+    if (err instanceof HttpError && err.code === 404) return;
+    throw err;
+  }
+  if (!isTerminating) return;
+
+  while (Date.now() < deadline) {
+    if (abortController.signal.aborted)
+      throw new DOMException('Aborted', 'AbortError');
+    await new Promise((resolve) => setTimeout(resolve, POD_POLL_INTERVAL_MS));
```
OliwiaGowor: The loop unconditionally sleeps before making its first follow-up `GET`, unlike `pollPodReady` which polls immediately then sleeps. There is always an unnecessary 2s delay before checking whether the pod already finished terminating. The `setTimeout` could be moved to the end of the loop

OliwiaGowor (commented):
- [ ] There is no test that verifies the _'Timed out waiting for terminal pod to finish terminating.'_ path is reached and the correct error is thrown. 
- [ ] The existing _'stops polling once the controller is aborted'_ test only covers the abort path inside `pollPodReady`. A similar pod-terminating-plus-abort scenario for `waitIfTerminating` is missing. Aborting mid-wait is a realistic scenario, this would be nice to have

OliwiaGowor (approved):

PR #5281, opened 2026-09-03
fix: Center protected resources icon horizontally

mrCherry97:
**Description**

- Replace `<Icon>` rendered as Button children with the `icon` prop on `Button` for the icon-only variant of `ProtectedResourceWarning`. Using the `icon` prop triggers UI5's icon-only button mode, which centers the icon horizontally — placing an `Icon` component as children renders it in the text slot and left-aligns it.
- Remove the previous `marginTop` workaround that was patching vertical misalignment.
- Override `--sapButton_Lite_TextColor` and `--sapButton_Lite_Hover_TextColor` CSS custom properties inline to keep the icon orange (`--sapCriticalColor`), since the `icon` prop does not accept a `design` attribute.

**Related issue(s)**

N/A

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/shared/components/ProtectedResourcesButton.tsx
+++ b/src/shared/components/ProtectedResourcesButton.tsx
@@ -7,6 +7,7 @@ import {
 } from '@ui5/webcomponents-react';
 import { useFeature } from 'hooks/useFeature';
 import { useId, useState } from 'react';
+import type React from 'react';
 import { createPortal } from 'react-dom';
 import { configFeaturesNames } from 'state/types';
 import jp from 'jsonpath';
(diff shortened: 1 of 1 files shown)


src/shared/components/ProtectedResourcesButton.tsx
```diff
@@ -7,6 +7,7 @@ import {
 } from '@ui5/webcomponents-react';
 import { useFeature } from 'hooks/useFeature';
 import { useId, useState } from 'react';
+import type React from 'react';
```
chriskari: this import is unnecessary

chriskari (approved):

PR #5283, opened 2026-09-03
fix: Validate pending kubeconfig ID read from localStorage

mrCherry97:
**Description**

- Added format validation for the kubeconfig ID read from `localStorage` in `consumePendingKubeconfigId()` — only values that survive an `encodeURIComponent` round-trip unchanged (i.e. contain no characters needing URL-encoding) are returned; anything else is discarded as `null`
- The `localStorage` entry is now always removed on read regardless of validation result, preventing stale or malicious values from persisting across SSO redirects
- Addresses a SAST-flagged taint path (Client DOM Stored XSS) from `localStorage.getItem` through `window.location.href` in `ssoDataAtom.tsx`

**Related issue(s)**

N/A

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit

### DIFF
--- a/src/state/intendedPathAtom.ts
+++ b/src/state/intendedPathAtom.ts
@@ -5,10 +5,13 @@ export function savePendingKubeconfigId(kubeconfigId: string): void {
   localStorage.setItem(PENDING_KUBECONFIG_ID_KEY, kubeconfigId);
 }
 
+const SAFE_ID_RE = /^[\w.-]+$/;
+
 export function consumePendingKubeconfigId(): string | null {
   const value = localStorage.getItem(PENDING_KUBECONFIG_ID_KEY);
-  if (value) localStorage.removeItem(PENDING_KUBECONFIG_ID_KEY);
-  return value;
+  if (!value) return null;
+  localStorage.removeItem(PENDING_KUBECONFIG_ID_KEY);
+  return SAFE_ID_RE.test(value) ? value : null;
 }
 
 export interface IntendedPath {


src/state/intendedPathAtom.ts
```diff
@@ -5,10 +5,13 @@ export function savePendingKubeconfigId(kubeconfigId: string): void {
...
 
+const SAFE_ID_RE = /^[\w.-]+$/;
+
 export function consumePendingKubeconfigId(): string | null {
   const value = localStorage.getItem(PENDING_KUBECONFIG_ID_KEY);
-  if (value) localStorage.removeItem(PENDING_KUBECONFIG_ID_KEY);
-  return value;
+  if (!value) return null;
+  localStorage.removeItem(PENDING_KUBECONFIG_ID_KEY);
+  return SAFE_ID_RE.test(value) ? value : null;
```
OliwiaGowor: It could probably be done without the regex by adding
`return encodeURIComponent(value) === value ? value : null;`

```suggestion
export function consumePendingKubeconfigId(): string | null {
const value = localStorage.getItem(PENDING_KUBECONFIG_ID_KEY);
  if (!value) return null;
  localStorage.removeItem(PENDING_KUBECONFIG_ID_KEY);
  return encodeURIComponent(value) === value ? value : null;
```

If this doesn't work, please add a comment explaining the regex pattern

OliwiaGowor (approved):

PR #5285, opened 2026-09-04
fix: Block DNS-resolved private IPs in Kubernetes backend proxy

### LINKED ISSUE #5115: Check and fix security vulnerabilities detected by "Frontier AI Security" initiative
In the context of "Frontier AI Security" initiative, an internal study was conducted to evaluate the capabilities of frontier AI models for static vulnerability detection.

Each of the findings that is valid should be fixed.

[COMBINED_DETAILED_SECURITY_REPORT.pdf](https://github.com/user-attachments/files/29506685/COMBINED_DETAILED_SECURITY_REPORT.pdf)

| Status | # | Finding | Action / Details |
|--------|---|---------|-----------------|
| ✅ | 1 | Kubernetes backend proxy accepts attacker-controlled private or DNS-resolved cluster hosts | Add `lookup: resolveOrBlockPrivateIpAddress` to `https.request` options in `handleK8sRequests` (`backend/kubernetes/handler.js:99-106`). `localIpFilter` only rejects `.cluster.local` and literal private IPs; DNS hostnames are not resolved before the connection is made. PR: #5285 |
| ⬜ | 2 | Unauthenticated /proxy can serve attacker-controlled active content under the Busola origin | Add authentication middleware (JWT or `requireK8sCredential`) before `proxyHandler` (`backend/index.js:71`). Also strip upstream response headers and force `X-Content-Type-Options: nosniff` so relayed HTML/JS cannot execute in the Busola origin. |
| 🚫 | 3 | ConfigMap custom extensions execute Busola-origin script with victim Kubernetes fetch capability | **Not relevant** — `EXTENSIBILITY_CUSTOM_COMPONENTS` is a stage-only feature, disabled in production. The `kymaFetchFn` exposure and `customHtml` rendering are intentional capabilities for extension developers on that tier; they are not available to end users in production deployments. |
| ⬜ | 4 | Custom web component prop_* attributes eval ConfigMap-provided HTML attribute values | Remove `eval(this.attributes[i].value)` in `src/web-components/createWebComponent.jsx:115-117`. Replace with `JSON.parse` for serializable props or a strict typed prop format. Any cluster admin who can write a ConfigMap extension can execute arbitrary JS in the Busola origin. |
| ✅ | 5 | Unauthenticated /proxy DNS precheck-to-connect gap can reach private HTTPS services | **Fixed** — `backend/proxy.js:37` now passes `lookup: resolveOrBlockPrivateIpAddress` into `https.request`, intercepting DNS resolution at the OS `connect()` level rather than in a separate pre-check step. |
| 🚫 | 6 | AI companion routes can spend backend service credentials after missing or presence-only caller gates | **Not relevant** — The `/public-key` route performs its own gate via `checkJouleEligibility` (issuer URL + cluster region checks). Joule key exchange must work before K8s credentials are fully established; the absence of `requireCompanionCredential` here is by design. |
| ⬜ | 7 | Unauthenticated /proxy forwards inbound sensitive headers to attacker-selected HTTPS origins | Replace `headers: { ...req.headers, host: parsedUrl.host }` in `backend/proxy.js:34` with an explicit allowlist (e.g. `Accept`, `Content-Type`, `Content-Length` only). All inbound headers including `Authorization`, `Cookie`, and SSO identity headers are currently forwarded verbatim. |
| ⬜ | 8 | Kubernetes proxy path allowlist is bypassed by x-cluster-url base path | Reject non-root pathnames in `x-cluster-url` during header extraction in `backend/kubernetes/handler.js:189-196`. `pathWhitelistFilter` checks `req.originalUrl` but `buildK8sRequestPath` prepends `targetApiServer.pathname`, so a cluster URL like `https://host/blocked/` bypasses the whitelist. |
| ⬜ | 9 | Community module URL allowlist can be bypassed by redirects followed by backend fetch | Pass `{ redirect: 'error' }` to `fetch()` in `backend/modules/communityRouter.js:31`. The initial URL is validated against the GitHub domain allowlist, but default redirect-following means a `301` from a trusted GitHub URL can redirect to an arbitrary destination. |
| 🚫 | 10 | Community module and build extension sources use mutable GitHub-hosted YAML without integrity binding | **Not relevant** — This is a broad supply-chain concern applicable to any project fetching from GitHub. SHA-pinning community YAML is an ops/policy decision, not a code defect. No standalone attack path exists without a prior compromise of the GitHub account serving the YAML. |
| ⬜ | 11 | Busola UI link renderers accept unsafe URL schemes from extension and module-provided metadata | Validate the final resolved URL starts with `https://` or `http://` in `ExternalLink` components. `makeHref` conditionally prepends `https://` but values like `javascript:alert(1)` or `data:text/html,...` pass through unmodified to `href` and `window.open`. |
| 🚫 | 12 | Cluster kubeconfig authentication material persists in browser Web Storage | **Not relevant** — `localStorage` persistence is intentional for session continuity across page reloads. The risk is only exploitable via XSS, which is addressed separately. There is no standalone attack path that does not already require code execution in the Busola origin. |
| ⬜ | 13 | Community module importer performs unbounded backend fetch and YAML parse from attacker-controlled GitHub URLs | Check `Content-Length` and reject oversized responses; cap streaming read size; add `signal: AbortSignal.timeout(ms)` to the `fetch` in `backend/modules/communityRouter.js:37-38`. Unbounded `response.text()` followed by `jsyaml.loadAll` on large or deeply nested YAML can exhaust backend memory/CPU. |
| ⬜ | 14 | Kubeconfig files exposed through /backend/kubeconfig and static /kubeconfig | Add auth middleware (`requireK8sCredential`) to the `/backend/kubeconfig` route (`backend/index.js:73`), which is currently registered with no authentication and returns filenames of all mounted kubeconfig YAML files. |
| ✅ | 15 | SSO Authorization header can leak upstream during client-certificate Kubernetes proxying | **Fixed** — `backend/kubernetes/handler.js:78-89` now builds `filteredHeaders` from an explicit `K8S_FORWARDED_HEADERS` allowlist (`accept`, `content-type`, `content-length`, `transfer-encoding`) before adding the `authorization` value. Ambient `Authorization` from the browser is no longer included. |
| 🚫 | 16 | HTTP x-cluster-url can leak Kubernetes credentials over plaintext | **Not relevant** — `http:` cluster targets are intentional for local/dev environments. No production cluster should use plaintext HTTP. This is a deployment configuration concern, not a code defect. |
| ✅ | 17 | IPv4-mapped IPv6 private-address bypass | In the IPv6 branch of `isPrivateIp` (`backend/utils/network-utils.js:35-41`), add: `if (lowerIp.startsWith('::ffff:')) return isPrivateIp(lowerIp.slice(7));`. Currently `::ffff:10.0.0.1`, `::ffff:127.0.0.1`, etc. bypass the private-address check. PR: #5285 |
| 🚫 | 18 | Joule loads configured external script and exposes bridge callbacks | **Not relevant** — `busola-config` lives in `kube-public` and requires elevated cluster permissions to write. Loading an operator-configured Joule script URL is by design; operators who can write `kube-public` are already fully trusted. |
| ⬜ | 19 | BUSOLA_SSL_ENABLED string/number comparison can disable backend TLS unexpectedly | Change `process.env.BUSOLA_SSL_ENABLED === 1` to `=== '1'` in `backend/index.js:93`. Environment variables are always strings; the strict equality check against the number `1` is always `false`, making the HTTPS server branch dead code. |
| 🚫 | 20 | Cluster-controlled config can set Sentry DSN | **Not relevant** — `kube-public/busola-config` requires operator-level cluster permissions to write. Allowing operators to configure a Sentry DSN is a legitimate cluster administration capability, not an attack surface. |

mrCherry97:
**Description**

- Add `lookup: resolveOrBlockPrivateIpAddress` to `https.request` options in `handleK8sRequests` so DNS resolution for the cluster hostname is intercepted at connect time and private IPs are rejected — previously the `localIpFilter` only blocked literal private IPs and `.cluster.local` suffixes, leaving DNS-resolved private addresses open (Finding 1 from the Frontier AI Security report, issue #5115)
- Fix IPv4-mapped IPv6 bypass in `isPrivateIp`: addresses like `::ffff:10.0.0.1` now delegate to the IPv4 check instead of passing through unblocked (Finding 17 from the same report)

**Related issue(s)**

Resolves #5115

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


backend/kubernetes/handler.js
```diff
@@ -10,10 +10,13 @@ import {
...
 } from '../utils/rate-limit-key.js';
 import { buildK8sRequestPath } from './path-utils.js';
+import { resolveOrBlockPrivateIpAddress } from '../utils/network-utils.js';
+import config from '../src/config/config.js';
 
 const https = require('https');
 const http = require('http');
 const fs = require('fs');
+const path = require('path');
 const escape = require('lodash.escape');
```
chriskari: this file mixes `require(...)` syntax with `import/export` at the top and `import.meta.dirname` down in `loadCerts`. Could we make these `require` statements be import statements too so the module is consistently ESM?

backend/utils/network-utils.js
chriskari: Now that this guard runs on the main K8s path too, I think the caching here needs a tweak. When the DNS lookup fails we block the request, which is correct — but we also store that "blocked" result in the cache. So if DNS is briefly unavailable, a perfectly valid cluster stays blocked for the full 5 minutes of the cache's TTL even after DNS is back. Let's only write to the cache when the lookup actually succeeds, and leave the failure case uncached so the next request just retries. The blocking behavior for the failed request itself stays exactly the same.

chriskari (approved):

docs/user/technical-reference/feature-flags.md
```diff
@@ -37,6 +37,9 @@ This table lists the configurable frontend feature flags, their descriptions, an
 > [!NOTE]
 > Backend features cannot be modified at the cluster's ConfigMap level.
 
+> [!NOTE]
+> **ALLOW_PRIVATE_IPS** can additionally be enabled through the `ALLOW_PRIVATE_IPS=true` backend environment variable (deployer-controlled, mirroring `JWT_CHECK_BYPASS`). This is intended for self-hosted deployments that run Busola in-cluster and must reach the in-cluster API server (`kubernetes.default.svc`, a private ClusterIP). Because it is a backend environment variable rather than the tenant-facing cluster ConfigMap, it does not weaken the secure-by-default posture of hosted deployments, which never set it.
```
grego952, line 41: deployer-controlled or deployer-controller?

mrCherry97: "controlled", person which is setting up busola on their local k3d can decide to use this flag or not

grego952 (approved):

mrCherry97:
Will be merged after https://github.com/kyma-project/busola/pull/5293

PR #5318, opened 2026-09-15
fix: Add heartbeat to prevent idle reconnection loop in terminal

chriskari:
**Description**

- Add heartbeat to prevent idle reconnection loop in terminal
- Cover changes in unit tests
- ...

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/10553

**Definition of done**

  - feat: A new feature
  - fix: A bug fix
  - test: Adding tests
  - revert: Revert commit


backend/src/terminal/connection.test.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+    this.ping = vi.fn();
+    this.send = vi.fn();
+    this.close = vi.fn();
+  });
+  WebSocket.OPEN = 1;
+  return { WebSocket };
+});
+
+const HEARTBEAT_INTERVAL_MS = 30_000;
```
KonradPietocha: The same one as in the `connection.js` file. It might be worth importing it; that way, if the value ever changes, the test will still work by using the same constant.

KonradPietocha (approved):

